PR #3, opened 2020-02-27
add missing word in readme

hrharder:
just a missing "to" (I think this is the correct fix. feel free to change of course)

library looks great!


gbotrel:
Good catch, thanks

PR #5, opened 2020-02-28
Updated go.mod to Go1.14 

gbotrel:
--> goff is now a go.mod dependency
--> fix txt/template updates in go 1.14 in internal code generation


ecc/bls377/fr/element.go
```diff
@@ -16,7 +16,7 @@
 //
 // 8444461749428370424248824938781546531375899335154063827935233455917409239041
 // Code generated by goff DO NOT EDIT
-// goff version: 1.0.0-alpha - build: 55b68d34592753ef2276c3678ea274ef57e7ba77
+// goff version:  - build:
```
ggutoski: These goff header comments should be cleaned up.

PR #9, opened 2020-03-05
Fix example in readme

nkeywal:


README.md
```diff
@@ -32,16 +32,17 @@ You'll need to [install Go](https://golang.org/doc/install).
 git clone https://github.com/consensys/gnark.git
 cd gnark
 make
+go build
```
ggutoski: My understanding is that `go build` is not needed here.  @gbotrel ?

nkeywal: The make alone does not create the gnark binary

gbotrel: indeed --> `gnark` uses build tags to determine at compile time which curve the binary is going to be compiled for. 
For example, `go build -tags bls381` (or bn256, bls377).

The makefile takes care of that, build a special binary that can switch between these "curve-specific" binaries depending on the input of the CLI. 

Note that the Makefile is not Windows friendly in its current form.

gbotrel: --> @nkeywall what platform? Can you copy paste here the output of the make?

nkeywal: linux

But actually the problem is that is writes in ~/.../go/bin/ not the current dir. If it's what you want you can remove this 'go build' from the patch.

gbotrel: Yes, it mimicks the `go install` command, which doesn't produce a binary in the current dir but in the GOPATH/bin dir.
We may do a `make install` and `make build` target to mimick both behaviors.

nkeywal: (nit inside) There is the directory examples/cubic_equation to report as well in the current master (it's examples/cubic)

'cd examples/cubic_equation' is not included in this patch but should be fixed as well.

PR #20, opened 2020-03-13
Feature/faster pippenger

ThomasPiellard:
- A somewhat optimized version of Pippenger is implemented and plugged in the template generator, I kept the old one and the benchmark tests both the former and the new one.
It is faster (on my mbp...) because the data is stored in a more efficient manner (at max I have a list of list instead of list of list of list), and a new addition formula is provided (parameterized Jacobian coordinates).

- The parameters (the bucket size) are chosen empirically after doing extensive testing.

- On my mbp the new version is faster (but not by much) below 1M points, after it is much faster, so it scales way better, it's in the order of twice as fast. The bucket size is blocked at 16 bits after 1M point (having a bucket size that divides 256 allows for a better workload distribution accross goroutines)

General remark: it's harder to parallelize than expected,  because for each task (add points in each buckets) have to be done sequentially, and in a correct order... So the real advantage here compared to the previous version is that it scales much better and the code is proper.


ecc/bls381/g1.go
```diff
@@ -466,39 +568,60 @@ func (p *G1Jac) MultiExp(curve *Curve, points []G1Affine, scalars []fr.Element)
...
 		chRes <- *p
 	}()
 	return chRes
 }
 
-// MultiExp set p = scalars[0]*points[0] + ... + scalars[n]*points[n]
-// see: https://eprint.iacr.org/2012/549.pdf
-// if maxGoRoutine is not provided, uses all available CPUs
-func (p *G1Jac) MultiExpNew(curve *Curve, points []G1Affine, scalars []fr.Element) chan G1Jac {
+type lockedG1Jac struct {
```
gbotrel: is that used anywhere? @ThomasPiellard

ecc/bls381/g1.go
```diff
@@ -26,6 +26,101 @@ type G1Affine struct {
 	X, Y fp.Element
 }
 
+// G1ZZZ parameterized jacobian coordinates (x=X/ZZ, y=Y/ZZZ, ZZ**3=ZZZ**2)
+type G1ZZZ struct {
```
gbotrel: should we export this new struct? if we want to, then we need a better name :-) If it's internal to the multiExp, start with lower case later to not export it

ecc/internal/gpoint/base.go
```diff
@@ -50,6 +50,101 @@ type {{.Name}}Affine struct {
 	X, Y {{.CType}}
 }
 
+// {{.Name}}ZZZ parameterized jacobian coordinates (x=X/ZZ, y=Y/ZZZ, ZZ**3=ZZZ**2)
+type {{.Name}}ZZZ struct {
```
gbotrel: cf comment on generated file: do we want to export that (if yes find better name) if not, lowercase letter

ecc/internal/gpoint/scalarmul.go
```diff
@@ -110,41 +117,63 @@ func (p *{{.Name}}Jac) MultiExp(curve *Curve, points []{{ .Name}}Affine, scalars
...
 		chRes <- *p
 	}()
 	return chRes
 }
-// MultiExp set p = scalars[0]*points[0] + ... + scalars[n]*points[n]
-// see: https://eprint.iacr.org/2012/549.pdf
-// if maxGoRoutine is not provided, uses all available CPUs
-func (p *{{.Name}}Jac) MultiExpNew(curve *Curve, points []{{ .Name}}Affine, scalars []fr.Element) chan {{.Name}}Jac  {
+
+type locked{{.Name}}Jac struct {
```
gbotrel: delete ?

ecc/internal/gpoint/scalarmul.go
```diff
@@ -154,96 +183,154 @@ func (p *{{.Name}}Jac) MultiExpNew(curve *Curve, points []{{ .Name}}Affine, scal
...
-			var index uint64
-			for i := start; i < end; i++ {
-				index = (scalars[i][limb] >> offset)
-				index &= mask
-				if index != 0 {
-					sharedBuckets[id][chunk][index-1].AddMixed(&points[i])
-				}
+
+	var bitsForTask [][]int
+	if 256%chunkSize == 0 {
```
gbotrel: too many magics number here --> 255, 256 etc should be constant referencing fr.Element defined constants

ecc/internal/gpoint/scalarmul.go
```diff
@@ -154,96 +183,154 @@ func (p *{{.Name}}Jac) MultiExpNew(curve *Curve, points []{{ .Name}}Affine, scal
...
+	}
+
+	mask = (1 << chunkSize) - 1
+	nbPointsPerSlots := len(scalars) / int(mask)
+	indices := make([][]int, int(mask)*nbChunks) // [][] is more efficient than [][][] for storage, elmts are accessed via i*nbChunks+k
+	for i := 0; i < int(mask)*nbChunks; i++ {
+		indices[i] = make([]int, 0, nbPointsPerSlots)
+	}
+
+	accumulate := func(cpuid, nbTasks, n int) {
```
gbotrel: cpuid --> cpuID

ecc/internal/gpoint/scalarmul.go
```diff
@@ -154,96 +183,154 @@ func (p *{{.Name}}Jac) MultiExpNew(curve *Curve, points []{{ .Name}}Affine, scal
...
+		for i := 0; i < nbTasks; i++ {
+			//start := time.Now()
```
gbotrel: remove commented line before merging

ecc/internal/gpoint/scalarmul.go
gbotrel: add comment to inner function accumulate and aggregate

cs/groth16/prove.go
```diff
@@ -78,112 +80,275 @@ func Prove(r1cs *cs.R1CS, pk *ProvingKey, solution map[string]cs.Assignment) (*P
...
-	chTokenA := make(chan struct{}, 1)
-	chTokenB := make(chan struct{}, 1)
+	// krs need to add Ar1 and Bs1
+	chKrsRandom := make(chan ecc.G1Jac, 2)
+
+	chMAr1, chMBs1, chMBs2 := multiExp(pk.G1.A, pk.G1.B, pk.G2.B, wireValues)
 
 	// Ar1 (1 multi exp G1 - size = len(wires))
-	chAr1 := computeAr1(pk, _r, wireValues, chTokenA)
+	go func() {
```
gbotrel: what is is doing in this PR? :)

ecc/bls381/g1.go
ThomasPiellard: I am not using it, I didn't check if it is used elsewhere

ecc/bls381/g1.go
ThomasPiellard: Not sure if we should export it, it's faster when one the input is affine, otherwise the standard Jacobian coordinates are fine.

Issue #19, opened 2020-03-11
Support programmatic pairing selection instead of build constraints


Nik-U:
Currently, choosing between BN256, BLS377, and BLS381 is a compile-time decision controlled by build constraints. If none of `bn256`, `bls377`, or `bls381` are provided as build tags, the library is currently unusable.

The current situation is undesirable for several reasons:
- Go libraries should be usable with a simple `go get` and without the need for a `Makefile` or adding custom build tags. Relying on the Go toolchain to handle all of this allows the project to be painlessly used as a (possibly transitive) dependency.
- Users of the library may need to work with a specific pairing. For example, proving statements about a custom twisted Edwards curve requires that the curve's field is the order of G1, and so users that provide such curves and make assumptions about them must also have a guarantee that gnark is using the correct pairing.
- Build constraints in Go are primarily meant for including architecture-specific optimizations (see for example the use in `big` and other uses in the standard library). The infrastructure is built around this intended use. Because of this, gnark is currently difficult to use as a transitive dependency. For example, if a developer writing an application depends on a networking library, that networking library depends on a crypto library, and that crypto library depends on gnark, then the application developer is now responsible for choosing the pairing used by gnark even though they are likely in no position to make an informed choice.
- If Library 1 depends on gnark and requires BLS381, and Library 2 depends on gnark and requires BLS377, then no application can depend on both Library 1 and Library 2.

While most of these issues could be solved by simply making one pairing the default (e.g., BLS381) when the other build tags are not present, that would pose a problem in the future if the default option is shown to have a security flaw, because then the default could not be changed without breaking old users.

The behavior that I would expect as a library user after this enhancement would be:
- Expose all of the different pairing choices in the API and let the user choose between them based on which packages they import.
- In cases where exported functions need to access the underlying pairing, allow the user to pass in the pairing that is in use (either with an interface implemented by other parts of the library or, less desirably, with a constant) or extract that information from the unexported contents of other parameters.
- If a default is provided, then ensure that there is some way to detect that an old default is in use and behave accordingly if the default is changed in the future.
- Export information about the different pairing choices (e.g., the order of the groups). It would be convenient for users requiring a specific pairing to ask the library for the modulus rather than assigning a `big.Int` themselves.

gbotrel:
Fair points, here is the though process we had and the (current) options:

[...]

Issue #17, opened 2020-03-06
Ecosystem of gadgets


gbotrel:
**WIP**
"gadgets" (ie circuit functions) are normal Go code. They can be used and imported like any other Go project. 

`gnark` should facilitate discovery of community-developed gadgets, and should examplify / enforce a consistant way to write gadgets to ensure homogeneity and code-quality in circuits.

PR #22, opened 2020-03-23
Architecture refactor

### LINKED ISSUE #16: architecture: move ecc/ package into its own repository
Once #15 and #13 are done would be a good time to move `ecc` into its own repository.

`gnark` would only contain Constraint Systems and Proving protocols. The crypto primitives they rely on would be reference in go.mod . 

Current `cs` pacakge needs refactoring (move out subpackages, etc).

gbotrel:
That's a big one:

* clear separation between frontend and backend: 
backend is now independent of frontend module
`R1CS` is not sharing data structures with `ConstraintSystem`

* code generation for test circuits
test circuits, used in `groth16` tests and `integration_test.go` are generated by `internal/generators/testcircuits`

* code generation for backends (groth16)
`backend` is generated by `internal/generators/backend` 

* no more Makefile, `go install` works as expected. 
`cmd/` imports `static/bls377`, `static/bn256`, ... etc. CLI do not need build tags 

Fixes #16  and #19 .


ThomasPiellard:
So there is a dispatcher in cmd/ to know which curve is used at runtime, and all different versions groth16 are available thanks to static/ thats stores all versions --> ok with that, but what is the purpose of having a groth16 version that still works with build tags (the one that is in backend/groth16/ and uses curve/)? I suppose it is for the frontend? The frontend depends then on the build tags?

Also since this branch is for refactoring I think we can go further:

there are 2 packages to handle the different curves: backend_bls377 and groth16_bls377 eg for bls377. We need to fix this: the backend_bls377 is only here to handle assignments (with associated visibility, etc) and R1CS structure
 
- the assignment management should be part of groth16_<curve>. It is entirely defined by the groth16 and shouldn't be separated (for MIRAGE for instance the visibility is different, it really depends on the protocol), so assignment.go should go in groth16_<curve>.

- the R1CS should be defined in a common root of groth16_<curve>/ : currently it depends on the curve because we store linear combinations, and the coeffs of those LC are of type Fr. We should instead make the coeffs strings: it's is up to the chosen version of groth16 to convert the coeffs in Fr elements. Then the Solve, Inspect methods will become functions and we call them with groth16_bls377.Solve(...)/Inspect(). By doing so it will allow in the frontend to import the different packages from static/ and Solve, Inspect the circuit are at user's will by calling the function from the correct package, this would be really useful for recursive snark, and it would allow to only have the groth16 in static/ and get rid of the one in backend/ 

- for the gadgets management: the gadget should be in either gorth16_<curve>, and the edwards curve related parameters should go in gurvy/, so the fronted would only depend on the backend, and the backend only on gurvy/. Currently we have a mix where the fronted depends somehow on gurvy, and on Fr, but the packages are added more or less by end to avoid the frontend to depend directly on gurvy... The frontend should be completely "Fr agnostic", each time something depends on the Fr it should be stored in the backend, or in gurvy

gbotrel:
1. There is no dispatcher in `cmd/`, did you checkout the last commit on this branch ? 
2. So, workflow overview: you write your `circuit.go` and `circuit_test.go` files. Using the `frontend` APIs, and hence build tags. Now, to run `Setup`, `Prove` or `Verify` in your application, you either import `gnark/backend/groth16` and use build tags, or, as your curve choice is fixed, directly import `gnark/backend/static/blsXXX/groth16` . 
3. I'm not sure we should merge `backend` and `groth16`, maybe move `r1cs.go` and `assignment.go` in another package called `middleend`? Other constructs we may decide to implement would rely on the R1CS data structure too, and `assignment` is what `r1cs.Solve()` consumes, so it's very tied to this. If there are some specifics on the `visibility` of assignments, we can abstract that when/if time comes?
4. "By doing so it will allow in the frontend to import the different packages from static/ and Solve" are you refering to the test part of the frontend? Why would we want the frontend to import things from the backend? It's currently completely independent from the `backend`, on purpose: someone using `gnark` to write circuits should be able to easily switch to another `backend` with "minimal" or no modifications to his circuit. Ie: `frontend` describes the circuit, doesn't care about `backend` stuff (except the data structure, i.e `r1cs.go`)
5.  ok for the edward curve parameters in `gurvy`
6. there may be something to dig to use `string` instead of `fr.Element` in the `R1CS`. That would enable not code generating the `R1CS` and probably get rid of build tags for `frontend`. However, it seems hard to do, for example, all the `toR1CS` methods like `func (i *implyExpression) toR1CS` . We need coefficients here for linear expressions, that depends on the modulus: `minusOne.Neg(&one)` . How do you encode that in a string? I have a feeling manipulating typed `elements` instead of strings will enable us to add some "minimization" logic in the CS overtime, but I see the interest of using "symbols" only in the `frontend`. 
7. Gadget management: frontend doesn't depend on `gurvy`. It depends on `gnark/curve` (and so build tags). You say the `frontend` should be `fr` agnostic, and I say it should be `backend` agnostic :p . I think there is something to do here (see #17 ) , and the `std/` may not stay in its current form / in the `frontend` package, but I don't think it makes sense to tie gadgets (which stays for me a fancy name to say "function" that are used in your circuit definition) to the `backend`

ThomasPiellard:
1-  What I call dispatcher is the big switch statement in prove.go for instance (it is in there the last version)

2-  The frontend depends on the backend (there is import github.com/consensys/gnark/backend everywhere, all the Fr elements depend on the backend... -> we can get rid of that using symbols instead of Fr, string seems the be the easiest solution, for -1 we just have to tweak the FroString method in Fr). What do you mean by it not depending on the backend?

3- it's just that with the different versions of groth16 as a user I wouldn't want to have 2+ packages that are curve dependent, the simpler the better, just having an API exposed by let's say groth16_bls377 is enough (again the structures depending on backend_<curves> are the r1cs, which is Fr independant, and the assignment, that we will need to change anyway if we switch for a universal circuit (tiny RAM, or MIRAGE, which have another type of inputs, they call them random inputs)

6- I think it will be relatively easy to do, just in the r1c.Solve() method, each time we solve a constraint we need to invoke fr.FromString(..) for each coefficient instead of calling it directly. For the code in frontend/ it will just consist in replacing the coeff with strings (there is no math operations in the frontend we don't care what we manipulate)

7- Yes the front end does not depend on gurvy but currently it is allowed to use it with build tags and it completely depends on the chosen backend, but a circuit should be independent from the backend -> if I write a circuit, with gadgets, etc, it is supposed to be executed on any "snark environment", if we don't do it will get really ugly to write recursive snarks

PR #23, opened 2020-04-20
Refactored backend and frontend to avoid use of build tags (curve dependency)

ThomasPiellard:
### Frontend

The CS and R1CS structures are now Fr agnostic (it uses big.Int now), so the frontend doesn't depend on the chosen Fr, and therefore gnark non longer uses tags. Tests to handle the Montgomery/Regular form of variables are OK (especially in the from/to_binary, select constraints for instance).

Tests are OK, but for creating a circuit, since now big.int are used, some variables can exceed the modulus: it's up to the user to query the modulus (not necessarily the r modulus if custom operations are created) to manually do the modular reduction (cf in point.go in gadgets/...twistededwards/ for instance).

### Backend

Assignments now receives big.Int.
Addition of a utils.go file to handle conversion from interface{} to big.Int.

### Circuits testing

Testing the correctness of the circuit solver is now done only in the backend/. In the frontend, only the consistency of the compiled circuit (R1CS struct) is tested because it's only what matters from the frontend's perspective. Nevertheless the tests in the backend cover all the subfunctions from cs.go.

### Cryptolib

The code previously in /reference (reference implementations for the gadgets) is now at the root of gnark under the name cryptolib/. We might want to move it to gurvy eventually.
The mimc functions implement the hash function interface provided by https://golang.org/pkg/hash/.
For the Merkle tree we should use the package https://godoc.org/github.com/NebulousLabs/merkletree with our mimc functions.
For the cryptolib some code generation is used for the different Fr (in crypotlib/internal).

### Gadgets

The gadgets are now at the root of the repo under the folder gadgets/. 
Gadgets that depend on the curve (like mimc) are now created by providing the ID of the chosen curve, the constructors are now in a hashmap ID->constructor(). 
The Merkle trees are wip.

### Gob (encoding)

The encoding package has been moved at the root of gnark, previously it was in internal/ preventing its use when using gnark as a module.

go.sum has been updated.

CircleCi whines because of the version, I should have messed with the version :/


backend/assignment_test.backup
```diff
@@ -0,0 +1,30 @@
+package backend
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
```
gbotrel, line 8: while is that file commited? need to restore these tests maybe?

ThomasPiellard:
Ok for the remarks (especially this
`_r1cs := circuit.ToR1CS()
	r1cs := backend_bn256.New(_r1cs)
	return &r1cs`
) I was planning to change this.

For the first remark:

I think eventually the mimc and eddsa functions (of the cryptolib) should be moved in gurvy or in their own repo, but the gnark is not the right place. I thought about moving those into gurvy, so I generated in advance the template, that should be put in the gurvy's generators... And also it solves as always the lack of generics.

PR #25, opened 2020-05-20
Feature/toy rollup

ThomasPiellard:
- Fixed MUSTBE_LESS_OR_EQ constraints and added associated tests

- Changed PublicKeyGadget, SignatureGadget in the eddsa gadget to more easy to use

- Added a toy version of a rollup (there is no db, the operator just stores the raw byte array corresponding to users accounts, I only implemented the spend circuit, not the deposit), the goal was to test a "real" circuit, evaluate the number of constraints, etc.

- Modified the Tag method for debugging constraints: now we pass a boolean to tell if we want to display the inputs or not.


crypto/accumulator/merkletree/readers.go
```diff
@@ -0,0 +1,72 @@
+// from https://gitlab.com/NebulousLabs/merkletree
```
gbotrel, line 1: you need to include the licence header from NebulousLabs here. If you can use their repo as a go mod dependency that would be cleaner (ie if code is not modified)
`
// Original Copyright 2015 Nebulous
// MIT licence header
// Copyright 2020 ConsenSys AG
// Apache licence

ThomasPiellard: I needed to modify their code: originally the leaf sum and node sum in the Merkle tree were done differently for domain separation purpose. The thing is the way it is done (according to  the specs) makes it snark unfriendly, it adds a lot of constraints. So after discussion with the applied research team we decided to find an easier way to differentiate the leaf node from the node sum (currently it's the same, but a future solution would be to have 2 mimc with different seeds for the leaf and the node).

ThomasPiellard:
For 2.: without this it automatically displays all the inputs and it overloads the logs, it super annoying when you want to debug a specific part of the circuit, debugging was way easier for me with this feature.

PR #33, opened 2020-09-24
only halve as many roots need to be precomputed for fft

mottla:
during the pre-computation of the roots of unity for the FFT domain,
the algorithm prepares two times as many elements, as actually needed.
This fix approximately halves the required storage space and runtime of the function.

### DIFF
--- a/internal/backend/bls377/fft/domain.go
+++ b/internal/backend/bls377/fft/domain.go
@@ -114,7 +114,7 @@ func (d *Domain) preComputeTwiddles() {
 	// for each fft stage, we pre compute the twiddle factors
 	twiddles := func(t [][]fr.Element, omega fr.Element) {
 		for i := uint(0); i < nbStages; i++ {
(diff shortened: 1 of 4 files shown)


gbotrel:
thanks for spotting that, merged into develop (with a slight modification: this part is generated from templates in `internal/generator/backend/fft` which is the file that needs modification)

mottla:
oh.. so no contributor entry for me I guess :(

gbotrel:
still merged your branch into develop so you have a commit in your name if that's any consolation ;) 
```
$ git log --author=mottla
commit 246234ea14c12164e65f4be2f2f78c06906bbde9
Author: mottla <wolf.mathias@outlook.com>
Date:   Thu Sep 24 17:07:28 2020 +0200

    only halve as many roots need to be precomputed for fft
```

mottla:
yess :D thanks!

PR #41, opened 2020-12-07
features(profiling): Adds a way to print the number of constraints in the circuit

AlexandreBelling:
Implements a two line methods `(*ConstraintSystem).NbConstraints()` that would enable constraints profiling and let the user know how many constraints have been created between two points of the circuit definition.

### DIFF
--- a/frontend/cs.go
+++ b/frontend/cs.go
@@ -140,6 +140,15 @@ func (cs *ConstraintSystem) Term(v Variable, coeff *big.Int) r1c.Term {
 	return term
 }
 
+// NbConstraints enables circuit profiling and helps debugging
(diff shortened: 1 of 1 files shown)


frontend/cs.go
```diff
@@ -140,6 +140,15 @@ func (cs *ConstraintSystem) Term(v Variable, coeff *big.Int) r1c.Term {
 	return term
 }
 
+// NbConstraints enables circuit profiling and helps debugging
+// It returns the number of constraints created at the current stage of the circuit construction.
+//
+// The number returns included both the assertions and the non-insertion constraints
```
gbotrel, line 146: non-insertion --> non-assertion ?

AlexandreBelling: Updated

PR #42, opened 2020-12-08
removed LinearExpressions: cs.Add and cs.Sub transparently create a constraint only when needed

gbotrel:
Previously, when defining a circuit, calling `cs.Add(x, y)` would create a constraint in many situation where this "additive constraint" could be combined with others into a single linear expression, forcing the use of `LinearExpression`.

This behavior is now fixed: no need for `LinearExpression`, `Add` and `Sub` behaves like expected and create a constraints only when a linear expression is referenced into a multiplicative constraint (or assertion).


frontend/cs.go
```diff
@@ -158,6 +175,93 @@ func (cs *ConstraintSystem) LinearExpression(terms ...r1c.Term) r1c.LinearExpres
 	return res
 }
 
+// reduces redundancy in a linear expression
```
gbotrel, line 178: @ThomasPiellard can you add more info in the doc of this function?

frontend/cs.go
```diff
@@ -158,6 +175,93 @@ func (cs *ConstraintSystem) LinearExpression(terms ...r1c.Term) r1c.LinearExpres
...
+	copy(res[accSize:], reduceSecret)
+	accSize += len(reduceSecret)
+	copy(res[accSize:], reduceInternal)
+	accSize += len(reduceInternal)
+	copy(res[accSize:], reduceUnset)
+	return res
+}
+
+func (cs *ConstraintSystem) bigIntValue(term r1c.Term) big.Int {
+	// const maxInt = int(^uint(0) >> 1)
```
gbotrel, line 253: clean up commented lines

frontend/variable.go
```diff
@@ -24,16 +21,27 @@ import (
...
 }
 
+// Variable of a circuit
+// represents a Variable to a circuit, plus the  linear combination leading to it.
+// the linExp is always non empty, the PartialVariabl can be unset. It is set and allocated in the
+// circuit when there is no other choice (to avoid wasting wires doing only linear expressions)
+type Variable struct {
+	PartialVariable
+	linExp    r1c.LinearExpression
+	isBoolean bool // TODO it doesn't work, we need a pointer for that
```
gbotrel, line 42: @ThomasPiellard  fix the TODO ?

PR #64, opened 2021-03-19
Feat/plonk prover

ThomasPiellard:
This PR contains a first version of the plonk prover, instantiated with a dummy polynomial commitment scheme. There are no custom gates.

The exposed API for using PLONK is the same as Groth16: `plonk.Setup(...)`, `plonk.Prove(...)`, `plonk.Verify(...)`.

**Note 1:** currently the verifier is not succint. Indeed the public polynomials corresponding to a circuit are not committed, and the verifier has to evaluate them at a random point, and this operation is not succint.

**Note 2:** Fiat Shamir is not implemented, so the numerous challenges are hard coded at the moment. It's the next step to add Fiat Shamir.

# API

The workflow for writing a circuit is the same as Groth16. Things starts to differ at the circuit compilation stage.

## Compilation

* `sparseR1cs, err := frontend.Compile(curve.ID, backend.PLONK, circuit.Circuit)` -> `sparseR1cs ` is now a list of constraints of the form `ql*l+qr*r+qm*l*r+qo*o+qk=0`, where` ql,qr,qm,qo,qk` are coefficients corresponding to the circuit, and` l,r,o` are the wires.

## Setup

Once a circuit is compiled, the list of coefficients `(ql,qr,qm,qo,qk)` needs to be converted to polynomials, and the permutation leaving `l || r || o` invariant (corresponding to the gates wiring) needs to be computed. Domains for FFTs and associated precomputed values are computed (there are 2 domains). Finally a polynomial commitment scheme is chosen.

Those information are public (since they correspond to a circuit that a prover and a verifier share) and are stored in a struct called `PublicRaw` (the raw means that the polynomials are not committed).

Generation of public data:
`publicData, err := Setup(sparseR1cs, polynomialCommitmentScheme, witness)`

Currently only a dummy polynomial commitment scheme is implemented in `crypto/` (it is not necessary for validating the plonk prover and plonk verifier).

The witness is necessary in the setup to add special constraints encoding the public input, namely the constraints` l-witness[i]=0` are prepend to the constraint system. The permutation takes in account those constraints (which are **not** encoded in the circuit after compilation), so they cannot be tampered. This operation is described in the comments of the `Setup` function which refer to "placeholders".

## Prove

`proof, err := Prove(sparseR1cs, publicData, witness)` --> it returns a `proof` object, containing commitments to the solution `l,r,o`, as well as the permutation accumulating polynomial Z, and the quotient polynomial H.

The order of operations is such that the number of FFT calls is minimized.

## Verify

`err = Verify(proof, publicData, witness)` --> verifies the openings of `l,r,o,h,z` at a challenge (on which the prover has no control), and verifies the final polynomial equality evaluated at the challenge.

**Note**: the opening of `l` from the prover is incomplete, and the verifier needs to complete it by adding the public witness, just as in Groth16. This mecanism ensures that the verifies is convinced that the proof is associated to the right public inputs, exactly like in Groth16.

# Status

* All tests circuits used for Groth16 pass for PLONK
* The verifier is not optimized, some operations need some speed ups (for instance using batch inversion at several places)
* No real polynomial commitment scheme is implemented ATM
* Fiat Shamir is **not** implemented, so all the challenges are hardcoded
* There are no breaking API change


crypto/internal/mock_commitment/digest.go.tmpl
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// MockDigest contains the polynomial itself
+type MockDigest struct {
+	Digest {{ toLower .Curve }}.Poly
+}
+
+// WriteTo mock impementation
+func (md *MockDigest) WriteTo(w io.Writer) (n int64, err error) {
+	return 0, nil
```
gbotrel, line 14: return errNotImplemented or panic

crypto/internal/mock_commitment/digest.go.tmpl
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// ReadFrom mock impementation
+func (md *MockDigest) ReadFrom(r io.Reader) (n int64, err error) {
+	return 0, nil
```
gbotrel, line 19: return errNotImplemented or panic

crypto/internal/mock_commitment/scheme.go.tmpl
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// Scheme mock commitment, useful for testing polynomial based IOP
+// like PLONK, where the scheme should not depend on which polynomial commitment scheme
+// is used.
+type Scheme struct{}
+
+// WriteTo panics
+func (s *Scheme) WriteTo(w io.Writer) (n int64, err error) {
+	return 0, nil
```
gbotrel, line 15: return errNotImplemented or panic

crypto/internal/polynomial.go.tmpl
```diff
@@ -0,0 +1,25 @@
+import (
+	"github.com/consensys/gurvy/{{ toLower .Curve }}/fr"
+)
+
+// Poly polynomial represented by coefficients {{ toLower .Curve }} fr field.
+type Poly []fr.Element
+
+// Degree returns the degree of the polynomial, which is the length of Data.
+func (p Poly) Degree() uint64 {
+	res := uint64(len(p) - 1)
```
gbotrel, line 10: if we have a unitialized Polynomial, (ie len(p) == 0) this return a large degree. Maybe return 0 if len(p) == 0? or panic, if we never hit that.

crypto/internal/polynomial.go.tmpl
```diff
@@ -0,0 +1,25 @@
...
+	return res
+}
+
+// Eval evaluates p at v
+func (p Poly) Eval(v interface{}) interface{} {
+	var res, _v fr.Element
+	_v.Set(v.(*fr.Element))
```
gbotrel, line 17: since you don't modify v, you may try to keep it as an address, to avoid unneeded copies

crypto/internal/polynomial.go.tmpl
```diff
@@ -0,0 +1,25 @@
...
+	s := len(p)
+	res.Set(&p[s-1])
```
gbotrel, line 19: same as above, this panics if len(p) == 0.

internal/backend/bls377/cs/r1cs_sparse.go
```diff
@@ -63,30 +63,25 @@ func (cs *SparseR1CS) CurveID() gurvy.ID {
...
+		return 1
 	}
-	if lro == -1 { // only O remains
-		lro = 2
+	if c.M[1].CoeffID() != 0 && !wireInstantiated[c.M[1].VariableID()] {
+		// M[1] corresponds to R by default
+		return 1
 	}
-	return lro
+	// TODO panic if wire is already instantiated
```
gbotrel, line 82: check if O was already instantiated

internal/backend/bls377/fft/fft.go
```diff
@@ -94,14 +106,21 @@ func (domain *Domain) FFTInverse(a []fr.Element, decimation Decimation, coset ui
 		panic("not implemented")
 	}
 
-	// scale by CardinalityInv
+	// scale by CardinalityInv (+ cosetTableInv is coset!=0)
 	if coset != 0 {
+		if decimation == DIF {
```
gbotrel, line 111: as discussed, this significantly slows down `groth16.Prove`

gbotrel (changes requested):
💯  awesome. 
Added few minor remarks -- 
will probably be refactored a bit once we have KZG commitment and various PLONK instantiations.

Issue #88, opened 2021-04-16
Split S in EdDSA signature to prevent overflow


ThomasPiellard:
In std/signature/eddsa/, the scalar S in the signature can exceed r of the underlying snark field. S must be split in 2 to avoid this overflow.

PR #89, opened 2021-04-16
docs: added comments for the splitting of S in eddsa

ThomasPiellard:
In EdDSA, the `S` scalar in the signature is now split in 2, to prevent overflowing --> S should NOT be reduced modulo r, the size of the snark field. So if n is the bit length of r, S is split in S1, S2 such that S = 2^(n/2)S1+S2, so S1 and S2 are the digits of S in basis 2^(n/2).

## Breaking changes

The signature struct is now
```
type Signature struct {
	R      twistededwards.Point
	S1, S2 frontend.Variable // --> instead of: S frontend.Variable
}
```
 Examples of how to properly split S given a signature is in std/signature/eddsa/eddsa_test.go. If r is n-bits long, splitting S amounts to take the first n/2 bits of the part of the signature corresponding to S for S1, and the last n/2 bits of the part of the signature corresponding to S for S2 (the signature is a byte slice point||S, the first n-bits chunk is the compressed point, the last bits are S).

### DIFF
--- a/std/signature/eddsa/eddsa.go
+++ b/std/signature/eddsa/eddsa.go
@@ -31,9 +31,13 @@ type PublicKey struct {
 }
 
 // Signature stores a signature  (to be used in gnark circuit)
+// An EdDSA signature is a tuple (R,S) where R is a point on the twisted Edwards curve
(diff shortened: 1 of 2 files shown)


gbotrel:
See also #88

PR #105, opened 2021-06-09
Feature/Point: `Neg`

SherLzp:
- Add `Neg` function for std/twistededwards/point
- Add tests for `Neg` function

### DIFF
--- a/std/algebra/twistededwards/point.go
+++ b/std/algebra/twistededwards/point.go
@@ -17,6 +17,10 @@ limitations under the License.
 package twistededwards
 
 import (
+	bls12377fr "github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
+	bls12381fr "github.com/consensys/gnark-crypto/ecc/bls12-381/fr"
+	bn254fr "github.com/consensys/gnark-crypto/ecc/bn254/fr"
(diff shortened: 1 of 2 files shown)


std/algebra/twistededwards/point.go
```diff
@@ -171,3 +175,33 @@ func (p *Point) ScalarMulFixedBase(cs *frontend.ConstraintSystem, x, y interface
...
+	case ecc.BLS12_381:
+		r = cs.Constant(bls12381fr.Modulus())
+		break
+	case ecc.BLS12_377:
+		r = cs.Constant(bls12377fr.Modulus())
+		break
+	case ecc.BW6_761:
+		r = cs.Constant(bw6761fr.Modulus())
+		break
+	default:
```
gbotrel, line 197: need to panic on default case when curve is not supported / implemented yet
For example we just added BLS24-315 curve --

SherLzp: that's right. Already add panic.

gbotrel:
@SherLzp thanks for the submission. Sorry for the delay to merge that, we have few other tasks right now and need to discuss if the `Neg` operation shouldn't be in the `cs` API like the `Inv` one.

SherLzp:
Noted

Issue #80, opened 2021-04-09
`gnark/std` needs some love


gbotrel:
`gnark/std` started as ad-hoc proof of concepts to validate `gnark` design choices and use cases like rollups. 

Needs a re-visit to ensure the packages under `gnark/std`:
* are rock solid
* have coherent APIs names and patterns 
* are well documented

SherLzp:
Hi, please see #103

gbotrel:
Comment; `gnark/std` is tested using `Groth16` on some curves only.

Need to use `backend.Implemented()` and  `ecc.Implemented()` to ensure good test coverage across backends en curves.

PR #159, opened 2021-10-28
Add Bandersnatch curve

yelhousni:
A proposition to add bandersnatch curve (in twisted Edwards form) to gnark. This PR implements Bandersnatch as a separate package from `twistededwards` mainly because each companion curve is identified by the base curve ID (e.g. BLS12-381 for both Bandersnatch and Jubjub). This has the advantage to keep backward compatibility but duplicates the exact same code. The other option is to create in gnark-crypto an ID for companion curves but gnark users should update their code.

### DIFF
--- a/std/algebra/twistededwards/bandersnatch/curve.go
+++ b/std/algebra/twistededwards/bandersnatch/curve.go
@@ -1,0 +1,68 @@
+/*
+Copyright © 2020 ConsenSys
+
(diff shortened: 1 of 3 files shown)


gbotrel:
Will leave the duplicated code for this release to avoid API changes, but revisit along side #80 . 
Track here: #165.

Issue #174, opened 2021-11-12
Automate and check for regressions `std/` number of constraints 


gbotrel:
Find a clean (and automated) way to document the number of constraint for each circuit component (maybe even API?) in `gnark/std` (per curve, per backend). 

This would give valuable information to compare `gnark/std` to other libraries, and to help make optimal design decision when picking a circuit component. 

See also #171 and #80

gbotrel:
note: number of constraints is a good metric for Groth16, but can be deceiving with PlonK (i.e custom gates may lower the number of constraints but augment the CPU time on the prover side)

gbotrel:
Idea suggested by @yelhousni : have an object, similar to the go std timer, that measures constraints added by a piece of circuit. 

```
Define(...) {

start := "time".Now() 
api.Mul(x, y)
// ... 
nbConstraints  := "time".Since(start)

}
```

* harder to do accurately with PlonK post-compile linear expression factorizations.

PR #176, opened 2021-11-16
feat: adds a constraint counter api

gbotrel:
Usage, inside a circuit:

```
	start := api.Tag("start_mulfp2")

	l1 := api.Add(e1.A0, e1.A1)
	l2 := api.Add(e2.A0, e2.A1)
	// ... etc

	end := api.Tag("end_mulfp2")
	api.AddCounter(start, end)

```

Then access through `CompiledConstraintSystem` with 
```
ccs, _ := frontend.Compile(...)
counters := ccs.GetCounters()

// in a test file
assert.SolvingSucceeded(&circuit, &witness, test.WithCurves(ecc.BW6_761))
counters := assert.GetCounters(&circuit)

for _, c := range counters {
t.Log(c)
}
``` 

`Counter` struct: 
```
// Counter contains measurements of useful statistics between two Tag
type Counter struct {
	From, To      string
	NbVariables   int
	NbConstraints int
	CurveID       ecc.ID
	BackendID     backend.ID
}
```

displays as:
```
groth16[bn254] start[g1.go:210] - end[g1.go:220]: 2277 variables, 2508 constraints
```

Notes:
* measurement for PlonK is contextual, since linear expression reduction may use previously reduced expression (ie measuring two repeating piece of circuit may yield a smaller number of constraint the second time)
* api naming is not definitive 

Also, first steps for #174

### DIFF
--- a/backend/witness/witness.go
+++ b/backend/witness/witness.go
@@ -48,11 +48,6 @@ import (
 	"reflect"
 
 	"github.com/consensys/gnark-crypto/ecc"
-	fr_bls12377 "github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
-	fr_bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381/fr"
-	fr_bls24315 "github.com/consensys/gnark-crypto/ecc/bls24-315/fr"
-	fr_bn254 "github.com/consensys/gnark-crypto/ecc/bn254/fr"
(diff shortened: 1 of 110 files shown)


gbotrel:
+ need a better way to identify `Counter` objects. `From`  / `To` will (currently) include line number, bad idea if we want to compare counters across branches for non-regression for example. 
+ figure out feature-wise, if maybe we want to have a default counter (englobing the whole circuit). What other stats we want to put in there. If we want to make that available only with a build tag.

PR #181, opened 2021-11-18
std/sw: use faster double-and-add for scalar multiplication and add constant scalar multiplication fast-path

ivokub:
Implemented the method for computing (P+Q)+P as described in https://github.com/zcash/zcash/issues/3924. Then, separated two execution paths for scalar multiplication depending on the scalar type:
* if scalar is a variable, then executes the generic scalar multiplication on variables (i.e. the scalar can be private,public,constant variable).
* otherwise, if scalar is a circuit parameter, then compiles the circuit based on the scalar bits.

For BW6-761 using Groth backend, the number of constraints went down from ~3.7k to 2662.

There are still some further improvement:

- [x] every double-and-add adds 7 constraints instead of claimed 5.
- [x] double-and-add is incomplete for case P=Q. Due to this, we need to handle two high bits separately in the main scalar multiplication loop.
- [x] scalar multiplication depends on the group order of BLS12-377. The dependence should be on the supplied curve (maybe add a method for getting companion curve params to ecc.ID?)
- [x] there is additional optimization as described on page in https://github.com/daira/halographs/blob/master/halographs.pdf using group endomorphism. This allows to reduce the numbers in the main loop by two.
- [x] the number of constraints explodes when using PLONK backend.
- [x] rebase on develop after #186 merged.


gbotrel:
Remark; you can use `api.CurveID().Info().Fr.Modulus()` or `api.CurveID().Info().Fr.Bits` to make the code field independent.

gbotrel:
@ivokub can this make its way in v0.6.0 ?

ivokub:
> @ivokub can this make its way in v0.6.0 ? 

Porting the changes on top of develop should be straightforward and I can do it tomorrow. If v0.6 closes after tomorrow then can manage it.

I was also working on#190 which is a bit related but will postpone it for later release.

ivokub:
I moved the changes over to #222 as my fork doesn't run CI. This PR can be closed without merging.

PR #187, opened 2021-11-30
Mark outputs of boolean operations as booleans

ivokub:
The boolean operations (AND, OR, XOR) already check that the inputs are
boolean constrained. The outputs of these operations are also always
boolean. We mark them in the constraint system as such so that further
boolean operations on them wouldn't introduce additional constraints.

### DIFF
--- a/circuitstats_test.go
+++ b/circuitstats_test.go
@@ -39,6 +39,7 @@ func TestCircuitStatistics(t *testing.T) {
 			check(backend.GROTH16)
 			check(backend.PLONK)
 		}
+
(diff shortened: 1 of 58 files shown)


ivokub:
Converted this PR to draft - will rebase properly after fix/plonk_constraints is merged.

gbotrel:
@ivokub is this still needed ?

ivokub:
> @ivokub is this still needed ?
> 

It is not anymore. The variables have now themselves an option if they are boolean constrained.

PR #191, opened 2021-12-07
Add subtests to test.Assert

ivokub:
Added method `Run(fn func(assert *test.Assert), descs ...string)` to Assert type. This is similar to the `Run` method in Go's standard testing package to allow running subtests. Refactored tests in places to use subtests.

Removed explicit `Fuzz` call in integration tests as Fuzz is already called in `Assert.ProverSucceeded`.

An potential added benefit of using subtests is better coverage with race detector and enabling running subtests in parallel. I didn't implement it now though as Assert uses unsynchronized cache for compiled circuits.

### DIFF
--- a/circuitstats_test.go
+++ b/circuitstats_test.go
@@ -3,13 +3,14 @@ package gnark
 import (
 	"encoding/gob"
 	"os"
+	"sync"
 	"testing"
 
 	"github.com/consensys/gnark-crypto/ecc"
 	"github.com/consensys/gnark/backend"
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/internal/backend/circuits"
-	"github.com/stretchr/testify/require"
+	"github.com/consensys/gnark/test"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


circuitstats_test.go
```diff
@@ -3,41 +3,40 @@ package gnark
...
-				nbConstraints := ccs.GetNbConstraints()
-				internal, secret, public := ccs.GetNbVariables()
-				checkStats(t, name, nbConstraints, internal, secret, public, curve, backendID)
+func TestCircuitStatistics(t *testing.T) {
+	assert := test.NewAssert(t)
+	for name := range circuits.Circuits {
+		for _, curve := range ecc.Implemented() {
+			for _, b := range backend.Implemented() {
+				// copy the circuit now in case assert calls t.Parallel()
+				tData := circuits.Circuits[name]
```
gbotrel, line 29: --> Circuits store `frontend.Circuit` which is an interface, this is not copying the actual underlying value. 

Here I would probably just go for a lock in the internal/backend/TestCircuit data structure.

gbotrel (changes requested):

ivokub:
I went over again over the different range loops and initialized the iteration-local variables everywhere they were used locally.

circuitstats_test.go
gbotrel: ha, I see. What I saw on some user defined circuits is developers re-using circuits variable along the way, or doing some slice allocations in the Define. 
For example:
```
type Circuit struct {
A frontend.Variable
B []frontend.Variable
}

Define(...) {
circuit.A = api.Mul(circuit.A, circuit.A)
circuit.B = make([]frontend.Variable, ....)
}

```
which is definitely not thread safe --> 2 concurrent call to compile will not work.

ivokub: Do you think if trying to do it [here](https://github.com/ConsenSys/gnark/blob/master/test/assert.go#L371) would work? I.e. deep copy the circuit before first compilation and compare the circuit to the copied circuit after compilation?

gbotrel: I found deep-copying circuits from their interface (frontend.Circuit) to be more complex than expected :) In internal/utils I believe, there is a ShallowCopy method.

ivokub: I see. I created a new issue for it https://github.com/ConsenSys/gnark/issues/195.

gbotrel (approved):

PR #192, opened 2021-12-09
Add multiple outputs for a hint function

ivokub:
This is an API breaking PR - the `frontend.API` interface changes to allow multiple outputs and now also returns errors (in case the inputs do not correspond to the hint signature).

This is still a draft PR - some tests fail with PLONK backend, but waiting for the refactoring and then will try again.

It is based on top of #191. As it is not yet merged, then this PR also contains commits from that PR but will be rebased later.

### DIFF
--- a/backend/hint/builtin.go
+++ b/backend/hint/builtin.go
@@ -1,7 +1,6 @@
 package hint
 
 import (
-	"errors"
 	"math/big"
 	"sync"
 
@@ -12,9 +11,9 @@ var initBuiltinOnce sync.Once
 
 func init() {
 	initBuiltinOnce.Do(func() {
-		IsZero = builtinIsZero
+		IsZero = NewStaticHint(builtinIsZero, 1, 1)
 		Register(IsZero)
-		IthBit = builtinIthBit
+		IthBit = NewStaticHint(builtinIthBit, 2, 1)
 		Register(IthBit)
 	})
 }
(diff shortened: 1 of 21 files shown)


ivokub:
I think the PR is done from my side. It may be more complex as necessary and it may make sense to simplify.

So - now the `hint.Function` is an interface which contains some annotations about the hint function. Namely, it returns the number of allowed inputs and outputs. For computing the number of allowed inputs, we consider only the `ecc.ID` but for number of allowed outputs we also consider the number of inputs. This should cover the use-cases where the number of outputs depends on the curve (e.g. for decomposing a scalar into bits) or on the inputs (e.g. a function which operates on arbitrary vectors).

I am not sure if it makes sense to have `TotalInputs()` method on `hint.Function`, but it allows to perform compile-time checks on the circuits (instead of erroring during solving).

See the new test in `internal/backend/circuits/hint.go` for how the hint functions should be defined in that case.

I also added a default constructor `hint.NewStaticHint` for "simple" hint functions which take constant number of inputs and returns constant number of outputs. This constructor outputs an instance implementing the `hint.Function` interface. It should be sufficient for the most of the cases.

gbotrel:
LGTM, remove TotalInputs & renamed TotalOutputs to NbOutputs which I think prettify/simplify hint definition. 
Also, in the solution part, there was a small big.Int pool leak, fixed. 

Merging 🎉  👍

PR #217, opened 2021-12-22
Re-exposed frontend.CompiledConstraintSystem

gbotrel:
Introduced by #200  --> 
`frontend.CompiledConstraintSystem` became `compiled.ConstraintSystem`, but `compiled` is `internal/` . 

Note that `GetCounters() []compiled.Counter` in the interface returns something from `internal` which may be impractical  / need revisit.


ivokub (approved):
Looks good. I agree that it is not great that `GetCounters()` returns something from `internal`, but we can revisit it later.

PR #228, opened 2022-01-08
refactor: Witness API with MarshalBinary and MarshalJSON

gbotrel:
Work in progress. Main changes:

Previously:
`backend/witness` package offered api like `witness.WriteFullTo(writer)` `witness.ReadPublicFrom(reader)`... 

Now has a struct that implements `encoding/BinaryMarshaler` and `encoding/json/Marshaler`. Essentially the witness can be in 3 different "state": 
1. `frontend.Circuit` --> this is referred to an "assignment"
2. `[]fr.Element`  --> this is represented under `witness.Witness` struct introduced by this PR
3. serialized (binary or json)

witness package enables to go from each of these states.

Other changes to validate approach:

`backend/groth16` and `backend/plonk` had `ReadAndProve` and `ReadAndVerify` methods. Now `Prove` and `Verify` take a `witness.Witness` as parameter. Puts more work on caller (can't pass a `frontend.Circuiit` directly) but enables all workflow more cleanly. 

This PR might be significantly simplified if we add generics. 

Added round trip marshaling test (binary & json) in the `test.Assert` call. 

Another PR should

- [x] Change APIs of `IsSolved` (in backend/groth16 & backend/plonk) to take a `witness.Witness` and to be defined, if possible directly on the `frontend.CompiledConstraintSystem` interface, since it's not directly dependent on the backend.


ivokub (changes requested):
From the examples it makes sense to require that witness is instantiated separately from the circuit. I haven't tried it out myself, but from the examples the proposed interface looks good.

There are a few comments, but mostly looks good.

PR #230, opened 2022-01-11
feat: added GetSchema in CompiledConstraintSystem

gbotrel:


ivokub:
Hmm, I see the changes and they are good, but I do not get the motivation for including Schema as part of CCS. Is it building up for upcoming PR?

gbotrel:
Well, in my mind, I have cases like a Prover / Verifier service, that'll have to potentially unmarshal json witnesses. 
This is the case for example in the gnark playground. 

We could probably store the schema away from the CCS , and leave it to the developer to handle that. But since schema is likely very cheap to store, I think it simplifies API to put in the CCS.

gbotrel:
(but I did put it in a separate PR because it's up for discussion since there may be better option to do that)

ivokub:
I see, makes sense. But wouldn't `SetSchema()` allow to create a conflicting state with CCS? It doesn't break CCS if the user sets invalid schema as it isn't used right now (? yet?).
I'll have to think it through a bit to understand more if there are any implications.

gbotrel:
`SetSchema` is not exposed (i.e not defined on the interface CompiledConstraintSystem)

ivokub (approved):
I think it was getting late for me. I followed through where `SetSchema` was defined and it seems safe.

PR #235, opened 2022-01-14
feat: added witness.Public to return Public part of the witness vector

gbotrel:


ivokub (approved):
I was just thinking about it when rebasing some code against develop. PR makes total sense and looks good.

PR #244, opened 2022-01-28
`ccs.GetConstraints` readable output

gbotrel:
This PR removes the `ccs.ToHTML` deadcode. It's not `gnark` job to produce HTML, the constraint system returns a list of formatted constraints in a `[][]string` and a user (like play.gnark.io ) can then render it. 

`ccs.GetConstraints` returns for PlonK the constraints in the form `qL⋅xa + qR⋅xb + qO⋅xc + qM⋅(xaxb) + qC == 0` (as in the paper) and for Groth16 `L⋅R == O` . 

Additionally, this PR adds a constraint ID number in the message for a unsatisfied constraint at solving time.

### DIFF
--- a/internal/backend/circuits/cmp.go
+++ b/internal/backend/circuits/cmp.go
@@ -55,5 +55,5 @@ func init() {
 		},
 	}
 
-	addNewEntry("cmp", &cmpCircuit{}, good, bad, []ecc.ID{ecc.BN254})
+	addNewEntry("cmp", &cmpCircuit{}, good, bad, ecc.Implemented())
 }
--- a/internal/backend/circuits/cmp.go
+++ b/internal/backend/circuits/cmp.go
@@ -55,5 +55,5 @@ func init() {
 		},
 	}
 
-	addNewEntry("cmp", &cmpCircuit{}, good, bad, []ecc.ID{ecc.BW6_761})
+	addNewEntry("cmp", &cmpCircuit{}, good, bad, []ecc.ID{ecc.BN254})
 }
--- a/internal/generator/backend/template/representations/r1cs.go.tmpl
+++ b/internal/generator/backend/template/representations/r1cs.go.tmpl
@@ -273,7 +273,7 @@ func (cs *R1CS) solveConstraint(r compiled.R1C, solution *solution) error {
 // GetConstraints return a list of constraint formatted as L⋅R == O
 // such that [0] -> L, [1] -> R, [2] -> O
 func (cs *R1CS) GetConstraints() [][]string {
-	var r [][]string 
+	r  := make([][]string , len(cs.Constraints))
 	for _, c := range cs.Constraints {
 		// for each constraint, we build a string representation of it's L, R and O part 
 		// if we are worried about perf for large cs, we could do a string builder + csv format.
--- a/internal/generator/backend/template/representations/r1cs.go.tmpl
+++ b/internal/generator/backend/template/representations/r1cs.go.tmpl
@@ -100,11 +100,11 @@ func (cs *R1CS) Solve(witness, a, b, c []fr.Element, opt backend.ProverConfig) (
 		// ensure a[i] * b[i] == c[i]
 		check.Mul(&a[i], &b[i])
 		if !check.Equal(&c[i]) {
+			errMsg := fmt.Sprintf("%s ⋅ %s != %s", a[i].String(), b[i].String(), c[i].String())
 			if dID, ok := cs.MDebug[i]; ok {
-				debugInfoStr := solution.logValue(cs.DebugInfo[dID])
-				return solution.values, fmt.Errorf("%w: %s", errUnsatisfiedConstraint(i), debugInfoStr)
+				errMsg = solution.logValue(cs.DebugInfo[dID])
 			}
-			return solution.values, errUnsatisfiedConstraint(i)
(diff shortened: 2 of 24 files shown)


internal/generator/backend/template/representations/solution.go.tmpl
```diff
@@ -14,8 +14,10 @@ import (
 	{{ template "import_curve" . }}
 )
 
-// ErrUnsatisfiedConstraint can be generated when solving a R1CS
-var ErrUnsatisfiedConstraint = errors.New("constraint is not satisfied")
+// errUnsatisfiedConstraint can be generated when solving a R1CS
```
ivokub, line 17: Previously, as ErrUnsatisfiedConstraint was a instantiated type, it could be checked using errors.Is. Now, it is essentially a utility function to call fmt.Errorf. When looking at the places it is called, it makes difficult to understand where the error actually happened (grepping leads here, not to the actual place the error happened).
Maybe remove this function and construct new errors where necessary? If it is necessary to assert this error, then a conventional way would be to define this as a type (satisfying the error interface), and use errors.As for assertion.

internal/generator/backend/template/representations/r1cs.go.tmpl
```diff
@@ -103,9 +102,9 @@ func (cs *R1CS) Solve(witness, a, b, c []fr.Element, opt backend.ProverConfig) (
 		if !check.Equal(&c[i]) {
 			if dID, ok := cs.MDebug[i]; ok {
 				debugInfoStr := solution.logValue(cs.DebugInfo[dID])
-				return solution.values, fmt.Errorf("%w: %s", ErrUnsatisfiedConstraint, debugInfoStr)
+				return solution.values, fmt.Errorf("%w: %s", errUnsatisfiedConstraint(i), debugInfoStr)
```
ivokub, line 105: As the error is not assertable anymore, then it doesn't make sense to wrap the error with "%w" anymore. Imo internally it returns a struct which contains the wrapped error and the message, but this overhead doesn't matter as the current function returns early.

internal/generator/backend/template/representations/r1cs.sparse.go.tmpl
```diff
@@ -99,9 +99,15 @@ func (cs *SparseR1CS) Solve(witness []fr.Element, opt backend.ProverConfig) ([]f
 		if err := cs.checkConstraint(cs.Constraints[i], &solution); err != nil {
 			if dID, ok := cs.MDebug[i]; ok {
 				debugInfoStr := solution.logValue(cs.DebugInfo[dID])
-				return solution.values,  fmt.Errorf("%w: %s", ErrUnsatisfiedConstraint, debugInfoStr)
+				if debug.Debug {
```
ivokub, line 102: See the same comments as for r1cs.go.tmpl

internal/generator/backend/template/representations/r1cs.sparse.go.tmpl
```diff
@@ -245,38 +251,80 @@ func (cs *SparseR1CS) IsSolved(witness *witness.Witness, opts ...backend.ProverO
...
+// https://eprint.iacr.org/2019/953.pdf section 6 such that
+// qL⋅xa + qR⋅xb + qO⋅xc + qM⋅(xaxb) + qC == 0
+// each constraint is thus decomposed in [5]string with
+// 		[0] = qL⋅xa
+//		[1] = qR⋅xb
+//		[2] = qO⋅xc
+//		[3] = qM⋅(xaxb)
+//		[4] = qC
 func (cs *SparseR1CS) GetConstraints() [][]string {
 	var r [][]string
```
ivokub, line 264: We know the number of returned strings ahead of time. We could allocate the whole slice to reduce the number of allocations in the loop.

internal/generator/backend/template/representations/r1cs.sparse.go.tmpl
```diff
@@ -245,38 +251,80 @@ func (cs *SparseR1CS) IsSolved(witness *witness.Witness, opts ...backend.ProverO
...
-	if tID == compiled.CoeffIdOne {
-		// do nothing, just print the variable
-	} else if tID == compiled.CoeffIdMinusOne {
-		// print neg sign
-		sbb.WriteByte('-')
-	} else if tID == compiled.CoeffIdZero {
-		sbb.WriteByte('0')
-		return sbb.String()
+	
+	sbb.Reset()
```
ivokub, line 278: Builder is already initialized.

internal/backend/circuits/cmp.go
```diff
@@ -55,5 +55,5 @@ func init() {
 		},
 	}
 
-	addNewEntry("cmp", &cmpCircuit{}, good, bad, []ecc.ID{ecc.BW6_761})
+	addNewEntry("cmp", &cmpCircuit{}, good, bad, []ecc.ID{ecc.BN254})
```
ivokub, line 58: Is there a reason for changing the curve for this test? Why not ecc.Implemented()?

ivokub (changes requested):

internal/backend/circuits/cmp.go
gbotrel: I suspect we missed it in PR #233 --> I like to run `go test -short` and with BW6 hard coded here, it takes 10x more time :)

Issue #254, opened 2022-02-09
bug lookup2 in plonk


yelhousni:
- lookup2 is not tested when inputs are constants. 
- It seems there is a bug when using lookup2 in plonk for fixed base scalar mul (see branch perf/EdDSA)

```
--- FAIL: TestScalarMulFixed (0.44s)
    --- PASS: TestScalarMulFixed/bn254/groth16 (0.04s)
    --- FAIL: TestScalarMulFixed/bn254/plonk (0.03s)
    --- PASS: TestScalarMulFixed/bls12_377/groth16 (0.04s)
    --- FAIL: TestScalarMulFixed/bls12_377/plonk (0.03s)
    --- PASS: TestScalarMulFixed/bls12_381/groth16 (0.05s)
    --- FAIL: TestScalarMulFixed/bls12_381/plonk (0.04s)
    --- PASS: TestScalarMulFixed/bw6_761/groth16 (0.08s)
    --- FAIL: TestScalarMulFixed/bw6_761/plonk (0.05s)
    --- PASS: TestScalarMulFixed/bls24_315/groth16 (0.06s)
    --- FAIL: TestScalarMulFixed/bls24_315/plonk (0.04s)
```

yelhousni:
There was no bug in `Lookup2` in plonk. The bug was related to a wrong op in sparse_r1cs constraint system (fixed in fd3ce5814f2781e4386ba0dacccb7470e45f15fb by @ThomasPiellard ). Closing this issue.

PR #257, opened 2022-02-14
EdDSA gadget

yelhousni:
Reduce the number of constraints of a single EdDSA verification (e.g. Groth16/BN254 `-2049` constraints).

### DIFF
--- a/std/algebra/twistededwards/bandersnatch/point_test.go
+++ b/std/algebra/twistededwards/bandersnatch/point_test.go
@@ -55,8 +55,8 @@ func TestIsOnCurve(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	witness.P.X = (params.BaseX)
-	witness.P.Y = (params.BaseY)
+	witness.P.X = (params.Base.X)
+	witness.P.Y = (params.Base.Y)
 
 	assert.SolvingSucceeded(&circuit, &witness, test.WithCurves(ecc.BLS12_381))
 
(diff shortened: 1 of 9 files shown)


yelhousni:
Modulo `Lookup2` for fixed inputs in PLONK #254

ThomasPiellard:
So the error was not the lookup table but a wrong op in the sparse_r1cs constraint system:
`r.ModInverse(&r, q).Mul(&l, &r).Mod(&l, q) `instead of
`r.ModInverse(&r, q).Mul(&l, &r).Mod(&r, q)`

gbotrel:
@yelhousni still a draft?

yelhousni:
> @yelhousni still a draft?

PR is ready after Thomas' fix.

std/algebra/twistededwards/bandersnatch/point_test.go
```diff
@@ -351,25 +361,39 @@ func TestNeg(t *testing.T) {
 
 }
 
-// benches
+// Bench
+func BenchmarkDouble(b *testing.B) {
```
gbotrel, line 365: mmhh I see what we do here, not very go-idiomatic (ie the benchmark is going to be run many times).

One way is to add this to the go test command line: ` -benchtime=1x`

gbotrel (approved):

std/algebra/twistededwards/bandersnatch/point_test.go
yelhousni: Got rid of benchmarks. Can be ran with `api.Counter` when needed.

ThomasPiellard (approved):

Issue #262, opened 2022-02-15
`std/signature/eddsa` is unnecessary complicated 


gbotrel:
Currently we force the circuit developer to instantiate the twisted edwards curve 

```
circuit.PublicKey.Curve, _ = twistededwards.NewEdCurve(api.Curve())
eddsa.Verify(api, circuit.Signature, circuit.Message, circuit.PublicKey)
```

1. It should not be inside the `eddsa.PublicKey` structure
2. It should probably be done behind `eddsa.Verify` 

--> only new parameter is that we now support `bandersnatch`; so for `BLS12-381` we may have multiple twisted edwards curves.

PR #263, opened 2022-02-16
ScalarMul on Bandersnatch

yelhousni:
`ScalarMul` on *Bandersnatch* now uses the D=-8 endomorphism with a hinted scalar decomposition. 

Groth16 bench:
- variable-base `ScalarMul` **2436** constraints
- fixed-base `ScalarMul` **1914** constraints
- variable-double-base `ScalarMul` **4844** constraints

For *Jubjub*, it costs:
- variable-base `ScalarMul` **3060** constraints
- fixed-base `ScalarMul` **2538** constraints
- variable-double-base `ScalarMul` **4844** constraints

Note that the variable-double-base `ScalarMul` algortihm is the same for both curves. It uses an interleaved simultaneous scanning of two scalars and a window table of 4 (`Lookup2`).  It is better constraint-wise to do so than using the endomorphism separately on two `ScalarMul`. Therefore, EdDSA verification cost is (almost) the same on both curve (*Bandersnatch* has 4C less as the cofactor clearing costs one doubling less), but the `ScalarMul` independently costs **624C** less.


std/algebra/twistededwards/bandersnatch/point.go
```diff
@@ -27,9 +29,16 @@ type Point struct {
...
+func (p *Point) Set(api frontend.API, p1 *Point) *Point {
+	p.X = p1.X
+	p.Y = p1.Y
+	return p
+}
+
 // Neg computes the negative of a point in SNARK coordinates
 func (p *Point) Neg(api frontend.API, p1 *Point) *Point {
-	p.X = api.Neg(p1.X)
+	p.X = api.Sub(0, p1.X)
```
gbotrel, line 41: is this actually any different than previous `Neg` call?

std/algebra/twistededwards/bandersnatch/point.go
```diff
@@ -98,41 +107,87 @@ func (p *Point) Double(api frontend.API, p1 *Point, curve EdCurve) *Point {
...
+	h := api.Sub(yy, curve.endo0)
+
+	p.X = api.DivUnchecked(f, xy)
+	p.Y = api.DivUnchecked(g, h)
+
+	return p
+}
+
+var scalarDecompositionHint = hint.NewStaticHint(func(curve ecc.ID, inputs []*big.Int, res []*big.Int) error {
+	// TODO: handle properly negative scalars
```
gbotrel, line 129: + check curveID is what is expected

std/algebra/twistededwards/bandersnatch/point.go
```diff
@@ -98,41 +107,87 @@ func (p *Point) Double(api frontend.API, p1 *Point, curve EdCurve) *Point {
...
+	var lambda, order big.Int
```
gbotrel, line 130: lambda, order and glvBasis shouldn't be initialized at each hint call --> if a circuit has multiple call to this hint, the overhead on perf will be significant in the solver. 

One way to do that would be; 
create a global variable (a new struct) with the 3 elements. 

Inside the hint, use `sync.Once` to do the init (SetString + PrecomputeLattice) only once 👍

std/algebra/twistededwards/bandersnatch/point.go
```diff
@@ -98,41 +107,87 @@ func (p *Point) Double(api frontend.API, p1 *Point, curve EdCurve) *Point {
...
+	// res[2].Mul(res[1], &lambda).Add(res[2], res[0])
+	res[2].Mul(res[1], &lambda).Sub(res[2], res[0])
+	res[2].Sub(res[2], inputs[0])
+	res[2].Div(res[2], &order)
+
+	return nil
+}, 1, 3)
+
+func init() {
+	hint.Register(scalarDecompositionHint)
```
gbotrel, line 149: hint func should be public, unless I missed something in the new pattern by @ivokub . 
Else, how would a prover (which may not import this package) provide the hint function to the backend calls?

std/algebra/twistededwards/bandersnatch/point.go
gbotrel, line 129: scalars can't be "negative" here. Inputs are big.Int , yes , but they are injected by the solver. So they were fr.Element converted to big.Int and will be 0 <= s < q .

std/algebra/twistededwards/bandersnatch/point.go
ivokub: As a side-note, this is the approach taken in WIP branch which merges different short Weierstrass implementations ([here](https://github.com/ConsenSys/gnark/pull/238/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-44b01c5b58d0588ce59dc5c74b729d43e99a85326fc34f919178f90b799bee2eR55)).

It would be great to follow similar approach, then it would be easier later to unify the implementations.

std/algebra/twistededwards/bandersnatch/point.go
ivokub: This is on me and sw implementation has the same problem. If the package isn't imported (both the CS and witness are deserialized), then the hint won't be provided.

In addition to exporting, I think doing something like #264 would also help with usability a bit?

std/algebra/twistededwards/bandersnatch/point.go
yelhousni: No there is no difference. I changed this while debugging and forgot to change it back.

std/algebra/twistededwards/bandersnatch/point.go
yelhousni: I followed the pattern in SW. Let's keep this in a separate ticket.

std/algebra/twistededwards/bandersnatch/point.go
yelhousni: yes, but at this point (as big.Int) they can be negative, and it seems `sp[0]` is always negative because `lambda` is 253-bit (max bitsize of scalar to decompose). What I did is negate the `sp[0]` before it is injected to guarantee it is the smallest within the range `[0,q[` and then negate the point inside the scalar mul function.

yelhousni: at which level should it be checked?

std/algebra/twistededwards/bandersnatch/point.go
```diff
@@ -98,41 +108,98 @@ func (p *Point) Double(api frontend.API, p1 *Point, curve EdCurve) *Point {
...
-	b := api.ToBinary(scalar)
+	s1, s2 := sd[0], sd[1]
 
-	res := Point{}
-	tmp := Point{}
-	A := Point{}
-	B := Point{}
+	// s1 + λ * s2 == s + k*r
+	// api.AssertIsEqual(api.Add(s1, api.Mul(s2, &curve.lambda)), api.Add(scalar, api.Mul(&curve.Order, sd[2])))
+	api.AssertIsEqual(api.Sub(api.Mul(s2, &curve.lambda), s1), api.Add(scalar, api.Mul(&curve.Order, sd[2])))
```
ThomasPiellard, line 183: Aren't there bounds to check as well? If `o` is the order of bandersnatch's prime group, in Z we have
`s1 + λ * s2 - s - k*o = 0`
`s1, s2 < sqrt(o)`
`o<r`
Here you only check that `s1 + λ * s2 = s + k*o mod r`, so it means that in Z `s1 + λ * s2 - s - k*o = c * r`. From the scalar decomposition we know that `s1, s2 < sqrt(o)<sqrt(r)`, but `λ ` is anywhere between `0` and `o`, so to me nothing guarantees that` c=0` here... I feel that something is missing, linked to the emulation of Z/oZ arithmetic in Z/rZ

std/algebra/twistededwards/bandersnatch/point.go
gbotrel: I think that's misleading; you can't inject a negative value here, because it's done at proving time, by the solver. So even if in the circuit definition, you "negate" it, it's going to be encoded in the constraint system, and when the solver will solve the constraints, it is going to be a fr.Element before it is a big int.

gbotrel: inside the hint I think, particularly if this hint works only for one curve; "if curve != ecc.BLS... return errors.New("....")"

yelhousni:
converting the PR to draft until we make sure of the hinted scalar decomposition.

std/algebra/twistededwards/bandersnatch/point.go
yelhousni: creating a ticket for this #268

std/algebra/twistededwards/bandersnatch/point.go
yelhousni: That's right. Also, if we reduce the scalar mod the Bandersnatch prime order (`s-o`), then the resulting scalar will bigger than 128-bit and we can't benefit from `api.ToBinary(s, 128)` and `for` loop of size 128, which is basically what reduces the number of constraints.

yelhousni: so as spotted and explained by @ThomasPiellard `sp[0]` is always negative because we always round above the determinant/2. It turns out that this is very suitable inside a circuit as we force `sp[0]` to be always negative. Thus we avoid incorrect reduction modulo `r` in the hint and we make sure of the scalars max bounds. However, this increases the bound from 128-bit to 129-bit costing +19C for a scalar mul.

gbotrel:
@yelhousni when is is going to be mergeable? I want to fix the weird apis in twistededwards / eddsa (see #262 ) and you may get quite a lot of conflicts if I do so before this PR :)

gbotrel:
@yelhousni ⬆️  ?

yelhousni:
> @yelhousni ⬆️ ?

I'll take care of that tomorrow. The GLV decomposition is correct (and scalar size bound is increased) but we may need to constrain few more things.

PR #265, opened 2022-02-17
constScalarMul for SW gadget

yelhousni:
When the scalar is constant (big.Int), there is no need to use `Lookup2` conditional. Instead we can just `Acc.DoubleAndAdd(Acc, table[k[0].Bit(i)+2*k[1].Bit(i)])`. This doesn't change the R1CS count (same mul complexity) but reduces the sparse_R1CS (PlonK) count by almost **2k** constraints.


gbotrel (approved):
👍

Issue #175, opened 2021-11-15
Solver hints with multiple outputs


gbotrel:
Solve hints currently forces the signature to return only one output.

However, if multiple variables are linked (bit decomposition, external-proof verification, ...) then it is wasteful (somethings prohibitively so) to compute the same hint n-times for each bit. 

Two ideas; 
1. from @AlexandreBelling  Make "stateful hints" solving order deterministic 
2. modify or add a new API `[N]Variable <- cs.NewHint(N, ..)`

ivokub:
Have you started working on it? If not, then I would have a look.

gbotrel:
All yours :) 
Ideally I'ld like to keep only 1 API for hints, no strong ideas on this topic 👍

ivokub:
For the second option - we can not return a variable-length array, but only slice. In that case, the arguments to NewHint wouldn't change, but the method would just return a slice of variables. So, the signature would be:
```
API.NewHint(f hint.Function, inputs... frontend.Variable) []frontend.Variable
```

We would also need to change the signature of `hint.Function` to:
```
type Function func(curveID ecc.ID, inputs []*big.Int, result []*big.Int) error
```

For me it seemed initially, that this wouldn't be a good approach, as then the user would need to know which elements in the returned `frontend.Variable` slice are which, but this can be documented with the hint function. I'll implement it and see how it looks.

gbotrel:
> Edit: Oh, I see - the variables are allocated before the function is evaluated and we do not know the number of resulting variables ahead of time.

Yes, that would force the user to pass `api.NewHint(n, ...)` , n being the size of the expected output. Or have hints functions be a struct implementing an interface rather than a function. Don't know.

PR #270, opened 2022-02-24
Simplify hints declaration and make frontend responsible for number of outputs, not hint / backend

gbotrel:
Creating this one as draft before cleaning it as it opens a discussion. It reverts some changes introduced by #183  #175  (cc @ivokub ). 

Essentially, this PR proposes to remove `nbInputs` and `nbOutputs` set in `StaticHintFunction` and the `NbOutputs()` function in the `Hint` interface. 

Trying to implement multi-outputs (with variable length) hints for binary operation, this declarative way is really not helping;

1. The circuit developer will provide a list of inputs to the hint function (example; `api.NewHint(myHint, input0, input1, ...)`. IF (and I don't think we need to) we want to check that the number of inputs matches what the hints can process, it's up to the hint to perform the check (ie if len(inputs) > ... --> error). Currently this was encapsulated in a `Call` function. 

2. Similarly, the circuit developer knows how many outputs it expects from a hint. He should set it like so `bits := api.NewHint(BitsHint, 220, value)` would return the first 220 bits for example in `[]bits`. Previously this was done declaratively at hint definition, through `NbOutputs()` method, which was called both in the frontend and the backend at solving time. --> it makes no sense as the backend will not decide the number of outputs, the frontend did at compile time and created the exact number of wires needed. 

@ivokub did you have specific future scenario in mind when doing this pattern?

### DIFF
--- a/backend/hint/builtin.go
+++ b/backend/hint/builtin.go
@@ -11,11 +11,11 @@ var initBuiltinOnce sync.Once
 
 func init() {
 	initBuiltinOnce.Do(func() {
-		IsZero = NewStaticHint(builtinIsZero, 1)
+		IsZero = NewStaticHint(builtinIsZero)
 		Register(IsZero)
-		IthBit = NewStaticHint(builtinIthBit, 2)
+		IthBit = NewStaticHint(builtinIthBit)
 		Register(IthBit)
-		NBits = NewStaticHint(builtinNBits, 1)
+		NBits = NewStaticHint(builtinNBits)
 		Register(NBits)
 	})
 }
--- a/backend/hint/builtin.go
+++ b/backend/hint/builtin.go
@@ -11,10 +11,12 @@ var initBuiltinOnce sync.Once
 
 func init() {
 	initBuiltinOnce.Do(func() {
-		IsZero = NewStaticHint(builtinIsZero, 1, 1)
+		IsZero = NewStaticHint(builtinIsZero, 1)
 		Register(IsZero)
-		IthBit = NewStaticHint(builtinIthBit, 2, 1)
+		IthBit = NewStaticHint(builtinIthBit, 2)
 		Register(IthBit)
+		NBits = NewStaticHint(builtinNBits, 1)
+		Register(NBits)
 	})
 }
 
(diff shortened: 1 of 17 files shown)


ivokub:
> Creating this one as draft before cleaning it as it opens a discussion. It reverts some changes introduced by #183 #175 (cc @ivokub ).
> 
> Essentially, this PR proposes to remove `nbInputs` and `nbOutputs` set in `StaticHintFunction` and the `NbOutputs()` function in the `Hint` interface.
> 
> Trying to implement multi-outputs (with variable length) hints for binary operation, this declarative way is really not helping;
> 
> 1. The circuit developer will provide a list of inputs to the hint function (example; `api.NewHint(myHint, input0, input1, ...)`. IF (and I don't think we need to) we want to check that the number of inputs matches what the hints can process, it's up to the hint to perform the check (ie if len(inputs) > ... --> error). Currently this was encapsulated in a `Call` function.
> 2. Similarly, the circuit developer knows how many outputs it expects from a hint. He should set it like so `bits := api.NewHint(BitsHint, 220, value)` would return the first 220 bits for example in `[]bits`. Previously this was done declaratively at hint definition, through `NbOutputs()` method, which was called both in the frontend and the backend at solving time. --> it makes no sense as the backend will not decide the number of outputs, the frontend did at compile time and created the exact number of wires needed.
> 
> @ivokub did you have specific future scenario in mind when doing this pattern?

`NbInputs` was already removed from the `hint.Function` interface before. It was really a leftover in `hint.StaticHintFunction` construction.

I added separate function for computing the number of output wires for cases when the number of outputs is non-trivial and is computed by some other function. For example, for ternary decomposition the number of outputs is ceil(log_3(modulus)) and it may be error-prone if the user computes it by hand. But, thinking now, we most probably do not want to encourage users to use such complicated hints directly and should rather refer them to use a gadget which provides (correctly) constrained version of them. I'm all good with removing `NbOutputs`.

In that context - with the creation of `std/math` / `std/math/bits`, would it make sense to make the built-in hints internal? In the context of [hints considered harmful](https://consensys.net/diligence/blog/2022/01/under-constrained-computation-a-new-kind-of-bug/)?

PR #272, opened 2022-03-01
feat: added std.GetHints for convenience. fixes #264

### LINKED ISSUE #264: hints in gadgets are not registered when packages are not imported
The problem appears when both the constraint system and witness are obtained by deserializing data.

Maybe add a new package which imports all hints in gadgets (`backend/hint/all` etc.) which has only a single side-effect of importing all hints from gadgets and registering. This package should then be imported by the compiler and provers.

gbotrel:
See #264. 

+ this PR adds a clear error message when a hint is missing for the solver; a compiled constraint system will be serialized with its hints dependencies; if any is missing at Solving time, the error message will contain the hint name.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -48,18 +48,6 @@ func (id ID) String() string {
 	}
 }
 
-// NewProverConfig returns a default ProverConfig with given prover options opts
-// applied.
-func NewProverConfig(opts ...ProverOption) (ProverConfig, error) {
-	opt := ProverConfig{LoggerOut: os.Stdout, HintFunctions: hint.GetAll()}
-	for _, option := range opts {
-		if err := option(&opt); err != nil {
-			return ProverConfig{}, err
-		}
-	}
-	return opt, nil
-}
-
 // ProverOption defines option for altering the behaviour of the prover in
 // Prove, ReadAndProve and IsSolved methods. See the descriptions of functions
 // returning instances of this type for implemented options.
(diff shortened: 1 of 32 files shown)


internal/generator/backend/template/representations/r1cs.go.tmpl
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/consensys/gnark/frontend/compiled"
 	"github.com/consensys/gnark/frontend/schema"
 	"github.com/consensys/gnark/backend"
+	"github.com/consensys/gnark/backend"
```
ivokub, line 15: Duplicate entry.

internal/generator/backend/template/representations/r1cs.sparse.go.tmpl
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/consensys/gnark/internal/backend/ioutils"
 	"github.com/consensys/gnark/frontend/compiled"
 	"github.com/consensys/gnark/backend"
+	"github.com/consensys/gnark/backend"
```
ivokub, line 16: Duplicate entry.

backend/backend.go
```diff
@@ -72,6 +60,18 @@ type ProverConfig struct {
 	LoggerOut     io.Writer       // defaults to os.Stdout
 }
 
+// NewProverConfig returns a default ProverConfig with given prover options opts
+// applied.
+func NewProverConfig(opts ...ProverOption) (ProverConfig, error) {
+	opt := ProverConfig{LoggerOut: os.Stdout, HintFunctions: hint.GetAll()}
```
ivokub, line 66: Maybe also `std.GetHints()`?

ivokub (changes requested):
Mostly looks good. I think the only point to consider if default prover config should include all hints from the gadget library or if there should be a prover option to include gadget hints.

backend/backend.go
gbotrel: import cycle hell :) std uses test package to test itself.

ivokub: :(
In that case - I think it is good as is. It will probably require too much refactoring?

Issue #243, opened 2022-01-27
perf: merging constraints in plonk


ThomasPiellard:
In plonk, when one creates a new variable like this for instance
`z := api.Add(api.Mul(x,y), 4)`
2 constraints are created, even if the intermediate result `api.Mul(x,y)` is not reused.

gbotrel:
I have an idea, which will be I think efficient to implement. (it accidentally solves #242 ). 

For a R1CS, idea is when you add a new constraint, if it’s in the form 1 * X == Y,  then, you always replace X by its original expression . 
2 cases:
1. a * b == X 
2. X * a == b 


1. we end up with a * b == Y 
2. we end up with Y * a == b

You keep a map of “orphans” constraints (ie a “new internal wire” is created, you add it to the map. When it’s referenced in any subsequent constraint, you remove it from the map. 

In a post compile phase, you delete all orphans and shift the ids accordingly (yet to figure out how to do that efficiently).

This approach make more sense for the Plonkish cs, since it avoids recursively “packing” linear/quadratic constraint into one. Currently with our api, if you write
y = x*x
-3y + yx + 2x == 0

it is going to create 4 constraints;
1. y = x *x
2. y * x = z
3. 2x - 3y = a
4. a + z == 0

with this proposition, it would be
1. y = x*x
2. y * x = z —> orphan
3. 2x - 3y = a —> orphan
4. 2x - 3y + y*x = 0

The idea is similar to R1CS, except, our API mostly creates either “linear” or “quadratic” plonkish constraints now. 

Note that since the parallel solver PR, we build a map of wire -> constraint (ie which constraint solved which wire) at post compile phase, and this data structure may be useful sooner 👍 .

gbotrel:
closing, superseded by plonk+custom gates wip.

Issue #242, opened 2022-01-27
perf: check for duplicate constraints during building CS


ivokub:
0 and 2 are essentially the same in http://play.gnark.io/?id=sdvlixeejg

gbotrel:
for R1CS builder; 

`newR1C` could do an optimistic check with a hashcode of the linear expressions against a hash map to see if we already have recorded a similar constraint.

But, we must know:
1. The new "internalVariable" is it in L, R or O ?
2. Is it reference previously in other constraints?

If answer to 2 is yes, then we probably need to maintain a sort of "alias" list, and do a post compile phase where we merge all variables that are "aliases" (technically interchangeable when they appear in constraints).

ivokub:
There is also a bit newer issue #243, which considers when two constraints can be merged in plonk (for example, can merge addition and multiplication). It seems that this issue should also cover the case of duplicate constraints in plonk.

In order to ensure that the behavior is uniform over all API operations, we should first ensure that all constraint insertions are performed through a single method (`system.addConstraint`). There are a some cases where the constraints are added directly (e.g. https://github.com/ConsenSys/gnark/blob/develop/frontend/cs/r1cs/api.go#L417, but need to verify the whole codebase) which could possibly pass the initial filtering step.

In order to hash the linear expression, we need to have some kind of fast set hash. I'd have to check, but maybe something like FNV+XOR would be good enough?
But, when hashing, we need to have a fixed value for the new internal variable added in the constraint as otherwise the set hashes of the same constraints (modulo the new internal variable) differ due to the difference in new internal variable ID. But for the hashing purposes, we can replace the ID of the new internal variable with a fixed one.

Regarding 2. - it seems a bit difficult problem. We do not want to have a full n-m mapping (between all constraints and all internal variables) as it may grow rather large. Maybe a kind of "transaction" for the internal variable (the phase where it can be recorded in new constraints as a "new" variable), such that we can eagerly perform the merging after every transaction closes.

PR #277, opened 2022-03-04
Expose low level constraint addition in Compiler interface

gbotrel:
- feat: exposed AddQuadraticConstraint and AddLinearConstraint
- test: missing file in previous commit

keeping as draft need to:

- [ ] decide what to expose exactly; we could expose a more expressive API for PlonK constraints. (cc @ivokub @ThomasPiellard )
- [ ] add tests


ivokub:
I'm not sure if there is any need to craft "full" (not quadratic nor linear) plonk constraints? Would there be any use cases where it would lead to fewer constraints in some circuit?

If at some point we would do custom gates, then we would need to expose plonk specific methods, but I would postpone for now.

gbotrel:
An example is this: http://play.gnark.io/?id=fevfbutxvg which takes 5 plonk constraints, but actually fits in 2. 

However, I agree, we should postpone for now --> tackling this problem, and #243  + #242 are actually linked... through I think a better data structure in the compiler. Rather than doing few patches here and there, I'ld prefer doing a clean approach once and for all.

PR #278, opened 2022-03-08
fixes #266 by adding constant path in Lookup2 and Select

### LINKED ISSUE #266: `api.Select` and `api.Lookup2` with constant booleans inputs should have special path
Lookup2 with 2 bits that are constants should be a "if" at compile time, likely not implemented as constraints.

gbotrel:


ivokub (approved):
Looks good.

PR #280, opened 2022-03-09
simplify post compile phase: shifting wire id is not needed 🙈 

gbotrel:
🙈 --> we used to do a weird dance in a post-compile phase, where we would loop through all constraints, logs, debuginfo, hints, and offset the ids of the wires. This was unnecessary complicated; new version just count the number of public and secret variables before calling `Define`, so, the r1cs builder and plonk builder both know the offset to add to new internal wires... voila. 

- feat: remove post-compile offset id in R1CS builder
- feat: remove offset shifts in plonk compile

### DIFF
--- a/frontend/compile.go
+++ b/frontend/compile.go
@@ -28,7 +28,7 @@ import (
 //
 // initialCapacity is an optional parameter that reserves memory in slices
 // it should be set to the estimated number of constraints in the circuit, if known.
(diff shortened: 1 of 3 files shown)


ThomasPiellard (commented):
Indeed it's simpler this way :) Is the Visibility field in Term still needed at all at this point? It was used when the public and secret variables were not a slice of []string, but a sort of wrapper around a variable...

PR #285, opened 2022-03-18
refactor: eddsa factorizing and code cleaning

### LINKED ISSUE #262: `std/signature/eddsa` is unnecessary complicated

gbotrel:
* This PR merges the `std/` code for bandersnatch and twistededwards in one single package.
* Refactors the tests. 
* Uses latest gnark-crypto to simplify eddsa signature; 

Twisted edwards curve usage is now, in a circuit (including bandersnatch);

```
curve, err := twistededwards.NewEdCurve(api, tedwards.BN254) // or tedwards.BLS12_381_BANDERSNATCH, ...
```
The `curve` object has point arithmetic defined on it, and if an endomorphism is available, will use scalarMulGLV. 

eddsa usage is now: 
```
err = eddsa.Verify(curve, signature, message, publicKey, hashFunction)
```

Both `std/.../eddsa` and `std/.../twistededwards` packages test all twisted edwards curves. 

Fixes #262


std/algebra/twistededwards/point.go
```diff
@@ -150,7 +145,7 @@ func (p *Point) ScalarMul(api frontend.API, p1 *Point, scalar frontend.Variable,
 // DoubleBaseScalarMul computes s1*P1+s2*P2
 // where P1 and P2 are points on a twisted Edwards curve
 // and s1, s2 scalars.
-func (p *Point) DoubleBaseScalarMul(api frontend.API, p1, p2 *Point, s1, s2 frontend.Variable, curve EdCurve) *Point {
+func (p *Point) DoubleBaseScalarMul(api frontend.API, p1, p2 *Point, s1, s2 frontend.Variable, curve *CurveParams) *Point {
```
ivokub, line 148: Couldn't this also use GLV?

yelhousni: It could but constraint-wise it's worse. It would need a `Lookup4`. The analysis is similar to a 2-bit window vs. a 4-bit window.

drakstik:
I'm new to Gnark and zk-proofs in general and I am trying to implement some of your examples, but for some reason they are full of little complexities that seem related to the fast-moving pace of this project.

Just wanted to comment that we noobs would greatly appreciate the examples to remain up to date and easily runnable. As fun as it is to rummage through the Issues tab, it is frustrating at times, considering how the topic itself is difficult to comprehend, let alone having to debug examples while you try to bring your ideas to life.

Thanks for all your hard work! It is because of people like you that noobs can even begin to dream up ideas for the new world.

PR #286, opened 2022-03-21
clean: remove PairingContext and Extension objects from api calls in std/.../pairing

gbotrel:
- fix: remove unused pairingContext.BTwistCoeff
- fix: remove unused pairingContext.BTwistCoeff
- clean: remove PairingContext and Extension objects from api calls in std/../pairing


gbotrel:
`Pairing` and `groth16.Verify` had `PairingContext` and `Extension` objects in all APIs, always initialized to the same values. 

Removed them in favor of a package private global var.

PR #289, opened 2022-03-22
refactor: remove StaticHint wrapper, log duplicate hints

gbotrel:
- refactor: remove statichint
- feat: log.Warn when duplicate hint registration.


backend/hint/builtin.go
```diff
@@ -10,7 +10,7 @@ var (
 	// IsZero computes the value 1 - a^(modulus-1) for the single input a. This
 	// corresponds to checking if a == 0 (for which the function returns 1) or a
 	// != 0 (for which the function returns 0).
-	IsZero = NewStaticHint(isZero)
+	IsZero = isZero
```
ivokub, line 13: I initially thought that it would be great to be able to collect all the built-in hints into a single `var` block for easier reference. Now there is only a single built-in hint left and now this doesn't make too much sense anymore. Maybe move the comment to `isZero` function and make that public instead?

std/hints.go
```diff
@@ -7,14 +7,13 @@ import (
...
-		sw_bls12377.DecomposeScalar,
-		bits.NTrits,
-		bits.NNAF,
-		bits.IthBit,
-		bits.NBits,
-	}
+// RegisterHints register std hints that are always injected in gnark solvers
+func RegisterHints() {
+	// note that importing these packages may already triggers a call to hint.Register(...)
+	hint.Register(sw_bls24315.DecomposeScalar)
```
ivokub, line 13: Even though `hint.Register` does not overwrite already registered hints, then it would be more safe to guard these registrations with a `sync.Once`.

std/hints.go
```diff
@@ -7,14 +7,13 @@ import (
...
-	return []hint.Function{
-		sw_bls24315.DecomposeScalar,
-		sw_bls12377.DecomposeScalar,
-		bits.NTrits,
-		bits.NNAF,
-		bits.IthBit,
-		bits.NBits,
-	}
+// RegisterHints register std hints that are always injected in gnark solvers
+func RegisterHints() {
```
ivokub, line 11: It seems that this function is not referred anywhere. Should there be an example, maybe? If i recall, then this function can not be called in compiler/prover due to import cycles.

ivokub (approved):
Looks good.

PR #294, opened 2022-03-25
fix: closes #293 and enables recursive hints solving

### LINKED ISSUE #293: bug: recursive hints not supported by R1CS solver
if we consider h1 and h2, two hint functions. 
h1 produces wire w1, h2 --> w2. 

Now, if h2 takes as input w1, and w1 didn't appear in a constraint preceding the one solving w2, then the solver needs to solve h1 first when encountering h2, which is not currently the case. 

- [x] add simple test case to reproduce the issue
- [x] restore panics in solver when a wire is unsolved (with debug tags only)
- [x] figure out how to deal with recursive hint solving (either when building dependencies in parallel levels, or at a later stage in the solver)

gbotrel:
- test: add failing test for #293
- fix: closes #293 and enables recursive hints solving


ivokub (approved):
Looks perfect. Also resolves my problem in feat/nonnative-ff.

PR #296, opened 2022-03-25
feat: add NoFuzzing() testing option

ivokub:
The new testing option allows to run assertions on `test.Assert` without fuzzing. My use case for running tests without fuzzing is in gadget where some errors panic to simplify the API.

Additionally, I was not sure if we should omit fuzzing when the tests are run in short mode? Should we?

### DIFF
--- a/test/assert.go
+++ b/test/assert.go
@@ -170,11 +170,13 @@ func (assert *Assert) ProverSucceeded(circuit frontend.Circuit, validAssignment 
 		}
 	}
 
-	// TODO may not be the right place, but ensures all our tests call these minimal tests
-	// (like filling a witness with zeroes, or binary values, ...)
-	assert.Run(func(assert *Assert) {
-		assert.Fuzz(circuit, 5, opts...)
-	}, "fuzz")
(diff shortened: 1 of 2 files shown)


gbotrel:
maybe it shouldn't be ran actually with "PRoverSucceeded" call. The point of this method is to test constraint system solver against the big.Int test engine -> if a witness fails/succeeds with one, it should with the others. 

What about just calling that separately on our integration tests?

ivokub:
Yup, removing fuzz testing in `ProverSucceeded` and then fuzzing separately in integration tests makes sense.
On a similar note -- maybe also remove marshalling tests from `ProverSucceeded` and move them into a new method? And add `assert.AllTests` etc. method for doing marshalling, compiling/solving, fuzzing etc.

ivokub:
(converted to draft until figuring what direction to take)

ivokub:
@gbotrel, do you have suggestion what to do with this PR? Someone stumbled against a similar problem why this PR was created (#336).

gbotrel (approved):

PR #298, opened 2022-03-28
fix: err instead of panic when recursively solving hints

ivokub:
Let the solver return the errors of the recursively solved hints instead of panicking. Otherwise fuzzer fails if hint returns an error on invalid input.

### DIFF
--- a/internal/generator/backend/template/representations/solution.go.tmpl
+++ b/internal/generator/backend/template/representations/solution.go.tmpl
@@ -139,18 +139,15 @@ func (s *solution) solveWithHint(vID int, h *compiled.Hint) error {
 	// is the output of another hint. 
 	// it is safe to recursively solve this with the parallel solver, since all hints-output wires 
 	// that we can solve this way are marked to be solved with the current constraint we are processing.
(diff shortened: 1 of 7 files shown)


internal/generator/backend/template/representations/solution.go.tmpl
```diff
@@ -139,18 +139,15 @@ func (s *solution) solveWithHint(vID int, h *compiled.Hint) error {
 	// is the output of another hint. 
 	// it is safe to recursively solve this with the parallel solver, since all hints-output wires 
 	// that we can solve this way are marked to be solved with the current constraint we are processing.
-	solveOrPanic := func(t compiled.Term) {
+	solveOrPanic := func(t compiled.Term) error {
```
gbotrel, line 142: semantically, maybe we shouldn't call it solveOrPanic then :)

gbotrel (approved):

internal/generator/backend/template/representations/solution.go.tmpl
ivokub: :), yup.

PR #302, opened 2022-04-07
feat: implement non-native field emulation

ivokub:
This PR adds a new standard gadget `std/math/nonnative` for performing non-native field operations in a circuit. It is trying to follow `big.Int`-like interface, but is not completely compatible.

The PR is mostly ready, but still some things to do:

- [x] add tests for more complex computations
- [x] implement `Select` and `Lookup2`
- [x] write a nice documentation which lists all the assumptions the library does to ensure the validness of the computation


ivokub:
Marking this version as ready to review as is. It is functionally complete, but lacks several optimisations I will implement in a new PR. When reviewing, I suggest to start from `std/math/nonnative/doc.go` (or look at the package documentation with `godoc`) for general description and then (in parallel) look at `std/math/nonnative/variable.go`.

SherLzp:
will it support `Sqrt`?

ivokub:
> will it support `Sqrt`? @ivokub

Currently `Sqrt` is not implemented in the PR, but it should be easy to add. Will add the `Sqrt` method soon.

ivokub:
> when will this PR be merged? @ivokub

I believe it is being reviewed right now. Cannot promise ETA but hopefully shortly.

ThomasPiellard (approved):

PR #321, opened 2022-05-24
Addition of a circuit that checks FRI proximity proofs

ThomasPiellard:
* Merkle proofs verification in circuit now use the binary decomposition of the leaf to know the order for hashing (instead of giving a sequence of "proof helpers" before)
* The rollup example has been fixed accordingly, and also each input is constrained now
* Addition of the fri verification circuit in `std/commitments/fri/`. The verification circuit checks radix-2 fri proximity proof as formatted in gnark-crypto. In particular, Fiat Shamir in gnark-crypto, when used with fri, has been modified so that the challenges seen as string outside a circuit and as frontend.Variable inside of the circuit match.


examples/rollup/circuit.go
```diff
@@ -108,12 +104,28 @@ func (circuit *Circuit) postInit(api frontend.API) error {
 		circuit.Transfers[i].SenderPubKey = circuit.PublicKeysSender[i]
 		circuit.Transfers[i].ReceiverPubKey = circuit.PublicKeysReceiver[i]
 
+		// allocate the slices for the Merkle proofs
+		// circuit.allocateSlicesMerkleProofs()
```
gbotrel, line 108: why is this commented?

examples/rollup/circuit_test.go
```diff
@@ -195,6 +213,7 @@ func TestCircuitUpdateAccount(t *testing.T) {
 	assert := test.NewAssert(t)
 
 	var updateAccountCircuit circuitUpdateAccount
+	(*Circuit)(&updateAccountCircuit).allocateSlicesMerkleProofs()
```
gbotrel, line 216: why the cast?

gbotrel (changes requested):
need to fix staticcheck errors + would have been nicer to have 3 separate PRs :-)

examples/rollup/circuit_test.go
ThomasPiellard: type `circuitUpdateAccount` inherits the fields but not the methods of `Circuit`

gbotrel:
👍  just staticcheck linter to fix and is good to merge

gbotrel:
@ThomasPiellard did a forge push on that one, gnark-crypto update was done in another branch 👍

gbotrel (approved):

Issue #312, opened 2022-05-10
refactoring: `frontend` APIs should take a field as input, not a curve


gbotrel:
* PlonK + FRI --> not curve dependent, but field dependent.
* Fuzz testing solvers against big int test engine --> not curve dependent, but field dependent.

It seems that the frontend APIs should not take a ecc.ID (ie a curveID) as parameter, but a “field id”. That would mean changing frontend.Compile api, but also the API interface (which has a CurveID() method) and the hint signature.

ivokub:
How will the actual curve be decided? Will there be any default mapping?

This proposal would actually make sense with #314 if we allow to give any field as an argument. But then we would need to decide separately which curve to work on.

PR #323, opened 2022-06-01
refactor: Compile(ecc.ID, ...) -> Compile(field.Field, ...)

gbotrel:
See #312 
- build: update to latest gnark-crypto
- refactor: frontend.Compile(ecc.ID) -> frontend.Compile(field.Field)

note: on a parallel branch, attempted to use generics to tackle this. Contaminated most of the packages. Perf issues in `R1CS` and `SparseR1CS` solvers made it a no-go 😞 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ func (circuit *CubicCircuit) Define(api frontend.API) error {
 
 // compiles our circuit into a R1CS
 var circuit CubicCircuit
-ccs, err := frontend.Compile(ecc.BN254, r1cs.NewBuilder, &circuit)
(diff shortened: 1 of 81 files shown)


ivokub:
> @ivokub here is a proposition to tackle #312 .
> 
> Unsolved issue; the `ConstraintSystem` now has a `field.Field` object which is an interface. Which fails the serialization round trip test -- what's the best way to proceed here in your opinion?

I think something like [this](https://github.com/ConsenSys/gnark/blob/develop/frontend/compiled/hint.go#L72) -- a custom CBOR marshaller/unmarshaller for `ConstraintSystem`. Right now it seems we can marshal `field.Field` as its modulus.

PR #326, opened 2022-06-06
feat: implement test engine options

ivokub:
Added options for the test engine and moved a an argument into the option. Additionally, added a new option for wrapping API and a boolean for defining the return value of the methods `ConstantValue` and `IsConstant`.

API wrapping was done to ease the testing of existing circuits using non-native field arithmetic.

Option for defining the constant values can be used to verify different code-paths in the circuits. I also changed the way fuzzer works -- now it checks that the test engine solves the circuit for both option and only if the test engine succeeds for both options (all vars constants or not), calls the actual prover.

### DIFF
--- a/test/assert.go
+++ b/test/assert.go
@@ -132,7 +132,7 @@ func (assert *Assert) ProverSucceeded(circuit frontend.Circuit, validAssignment 
 				checkError(err)
 
 				// must not error with big int test engine (only the curveID is needed for this test)
-				err = IsSolved(circuit, validAssignment, curve, backend.UNKNOWN)
+				err = IsSolved(circuit, validAssignment, curve)
 				checkError(err)
 
 				assert.t.Parallel()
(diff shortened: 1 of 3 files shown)


test/assert.go
```diff
@@ -383,9 +383,10 @@ func (assert *Assert) fuzzer(fuzzer filler, circuit, w frontend.Circuit, b backe
 	// fuzz a witness
 	fuzzer(w, curve)
 
-	err := IsSolved(circuit, w, curve, b)
+	errVars := IsSolved(circuit, w, curve, WithBackend(b))
+	errConsts := IsSolved(circuit, w, curve, WithBackend(b), SetAllVariablesAsConstants())
 
-	if err == nil {
+	if errVars == nil && errConsts == nil {
```
gbotrel, line 389: @ivokub what useful information can we (systematically) deduce when `(errVars == nil) != (errConsts == nil)` ?

ivokub: As an example, the cause for my bug was https://github.com/ConsenSys/gnark/blob/master/std/math/bits/conversion_binary.go#L68. Here the bit decomposition for constants and variables is slightly different. For variables the decomposed bits have to add up to the variable value, but not for constants.

I wouldn't say the behavior is wrong for constants, but it definitely is unexpected.

PR #328, opened 2022-06-06
refactor: Compile(ecc.ID) -> Compile(field *big.Int)

gbotrel:
This PR makes `frontend.Compile` `hints` `witness` and `cs` packages dependent on a `field` instead of being dependent on a `curveID`. 

This PR adds a `tinyfield` package (modulus == 47), for test purposes.

I believe the only algorithmic change is `witness` Serialization; instead of using `curve.Encoder` it directly uses `fr.Element [Set]Bytes()` methods. 

- build: update to latest gnark-crypto
- refactor: frontend.Compile(ecc.ID) -> frontend.Compile(field.Field)
- refactor: field.Field -> *big.Int
- build: update to latest gnark-crypto
- feat: added internal/tinyfield
- refactor: NewWitness(ecc.ID) -> NewWitness(field big.Int)
- feat: build R1CS with tinyfield
- test: ignore range circuit for tinyfield r1cs serialization
- refactor: hint now takes *big.Int scalar field instead of curveID


internal/generator/backend/main.go
```diff
@@ -102,6 +109,22 @@ func main() {
 				panic(err)
 			}
 
+			// groth16 & plonk
+			if d.Curve == "tinyfield" {
```
ivokub, line 113: Maybe instead of manually checking here add field `noBackend` field to `templateData` and check it?

std/algebra/fields_bls12377/e12_test.go
```diff
@@ -294,7 +294,9 @@ func TestFrobeniusFp12(t *testing.T) {
 	_, _ = a.SetRandom()
 	c.Frobenius(&a)
 	d.FrobeniusSquare(&a)
-	e.FrobeniusCube(&a)
+	// TODO @yelhousni restore
```
ivokub, line 297: Still relevant?

std/fiat-shamir/transcript_test.go
```diff
@@ -88,7 +88,7 @@ func getChallenges(curveID ecc.ID) (string, string, string) {
 	// it writes the domain separators as bytes
 	// in gnark, we write them as field element
 	// to ensure consistency in this test, we ensure the challengeIDs have a fix byte len (the one of fr.Element)
-	frSize := curveID.Info().Fr.Bytes
+	frSize := len(curveID.ScalarField().Bits()) * 8
```
ivokub, line 91: Division?

std/fiat-shamir/transcript_test.go
```diff
@@ -124,7 +124,8 @@ func TestFiatShamir(t *testing.T) {
 				bindings[i][j].SetUint64(uint64(i * j))
 			}
 		}
-		buf := make([]byte, curveID.Info().Fr.Bytes)
+		frSize := len(curveID.ScalarField().Bits()) * 8
```
ivokub, line 127: Division?

std/math/bits/conversion_binary.go
```diff
@@ -54,7 +53,7 @@ func fromBinary(api frontend.API, digits []frontend.Variable, opts ...BaseConver
 func toBinary(api frontend.API, v frontend.Variable, opts ...BaseConversionOption) []frontend.Variable {
 	// parse options
 	cfg := baseConversionConfig{
-		NbDigits:             api.Compiler().Curve().Info().Fr.Bits,
+		NbDigits:             api.Compiler().Field().BitLen(),
```
ivokub, line 56: If implement `BitLen()` in interface (described in other comment), then can replace.

std/math/bits/conversion_ternary.go
```diff
@@ -56,7 +55,7 @@ func fromTernary(api frontend.API, digits []frontend.Variable, opts ...BaseConve
 
 func toTernary(api frontend.API, v frontend.Variable, opts ...BaseConversionOption) []frontend.Variable {
 	// parse options
-	nbBits := api.Compiler().Curve().Info().Fr.Bits
+	nbBits := api.Compiler().Field().BitLen()
```
ivokub, line 58: If implement `BitLen()` in interface (described in other comment), then can replace.

ivokub (changes requested):
I think that in general the changes make sense. I'm just thinking that maybe it would be better to name the modulus variable as `primeFieldModulus` etc? If it is `field`, then may be ambiguous?

internal/utils/field_to_curve.go
```diff
@@ -0,0 +1,27 @@
...
+
+func init() {
+	curves = make(map[string]ecc.ID)
+	for _, c := range gnark.Curves() {
+		fHex := c.ScalarField().Text(16)
+		curves[fHex] = c
+	}
+}
+
+func FieldToCurve(q *big.Int) ecc.ID {
```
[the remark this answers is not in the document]
gbotrel: might be in a very distant future; however this method is internal and here to make this refactor smoother (ie we should kill the Curve() API anyway -- )

backend/witness/witness.go
```diff
@@ -121,15 +123,16 @@ func (w *Witness) MarshalBinary() (data []byte, err error) {
 // UnmarshalBinary implements encoding.BinaryUnmarshaler
 func (w *Witness) UnmarshalBinary(data []byte) error {
 
+	snarkFieldSize := len(w.CurveID.ScalarField().Bits()) * 8
```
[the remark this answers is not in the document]
gbotrel: it's the byte size; Bits() on big int returns the words, so len(bits()) == nbWords .

frontend/cs/r1cs/builder.go
```diff
@@ -58,21 +60,19 @@ type r1cs struct {
 
 	// map for recording boolean constrained variables (to not constrain them twice)
 	mtBooleans map[uint64][]compiled.LinearExpression
+
+	q *big.Int
```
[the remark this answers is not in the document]
gbotrel: Issue is ConstraintSystem doesn't (and shouldn't expose it). Field() method does an allocation since it duplicates the value for safety.

backend/witness/witness.go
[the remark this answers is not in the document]
gbotrel: that being said, it is not 32-bit compatible, will fix.

internal/backend/bls12-377/witness/witness.go
```diff
@@ -39,13 +37,18 @@ func (witness *Witness) WriteTo(w io.Writer) (int64, error) {
 		return 0, err
 	}
 
-	enc := curve.NewEncoder(w)
+	n := int64(4)
```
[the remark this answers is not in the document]
gbotrel: 4 is the size of the uint32 written at the previous line (len(witness))

internal/generator/backend/main.go
```diff
@@ -20,44 +20,60 @@ func main() {
...
 
 	var wg sync.WaitGroup
 
+	{
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			// generate R1CS on tiny field
+		}()
+	}
```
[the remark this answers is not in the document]
gbotrel: yes dead code from refactoring

std/algebra/fields_bls12377/e12_test.go
gbotrel: yes

drakstik:
Hey, I know this is a closed issue, but it was not updated in your Go package documentations https://pkg.go.dev/github.com/consensys/gnark#section-readme. Specifically, the first example given of a cubic equation. Would save time for beginners to just add the ecc.BN254.ScalarField() to the example! I had to look around the GitHub for this, which is good practice, but I'd rather save other's time just a little.

Thanks for all the hard work!

PR #331, opened 2022-06-23
perf: allocate less in test engine

ivokub:
At every API call utils.FromInterface method was called. This method always
allocates new *big.Int, which is very stressful for the runtime. Instead, we
first check if the variable is already *big.Int and return it otherwise. Only
if type assertion fails do we allocate new *big.Int.

Now, as we do not allocate every time, we had to change the API calls to
allocate a new result, so that we wouldn't mutate the passed in values.

### DIFF
--- a/test/engine.go
+++ b/test/engine.go
@@ -133,71 +133,73 @@ func IsSolved(circuit, witness frontend.Circuit, field *big.Int, opts ...TestEng
 }
 
 func (e *engine) Add(i1, i2 frontend.Variable, in ...frontend.Variable) frontend.Variable {
-	b1, b2 := e.toBigInt(i1), e.toBigInt(i2)
-	b1.Add(&b1, &b2)
+	res := new(big.Int)
+	res.Add(e.toBigInt(i1), e.toBigInt(i2))
 	for i := 0; i < len(in); i++ {
-		bn := e.toBigInt(in[i])
-		b1.Add(&b1, &bn)
+		res.Add(res, e.toBigInt(in[i]))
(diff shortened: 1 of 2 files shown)


test/engine.go
```diff
@@ -433,10 +442,17 @@ func (e *engine) AddCounter(from, to frontend.Tag) {
 	// do nothing, we don't measure constraints with the test engine
 }
 
-func (e *engine) toBigInt(i1 frontend.Variable) big.Int {
-	b := utils.FromInterface(i1)
-	b.Mod(&b, e.modulus())
-	return b
+func (e *engine) toBigInt(i1 frontend.Variable) *big.Int {
+	switch vv := i1.(type) {
+	case *big.Int:
```
gbotrel, line 447: shouldn't we still do a mod reduce if it's a big int? may slow things down (or we could have a fast path with a comparaison first) --> you can have circuit inputs / constants at this stage, so it may trigger weird edge cases if we allow some values in the test engine to not be mod reduced?

ivokub: I tried to rewrite the API methods such that `toBigInt` is only called for method inputs and I always recreate a new `*big.Int` for temp variables and result. The operations and results are always mod reduced.

I also tried to mod reduce the inputs in `toBigInt` method, but as I will be modifying inputs, I started having a few failing edge cases for the integration tests. Another approach would be to allocate a new `*big.Int`, set its value from the input and mod reduce it, but this goes against the goal of this PR which was to prevent unnecessary allocations.

PR #332, opened 2022-06-23
refactor: allow builder to interact with schema parser

ivokub:
In the upcoming fake-API for field emulation branch I have implemented a wrapper for `frontend.Builder` which initialises the variables in the circuit witness. The builder also needs to modify the parsed schema (for example, we need a native element for every limb). For that, the `LeafHandler` now takes as input `*schema.Field` which it can modify in-place.

I think this PR may not be optimal. Will try to rebase fake-API branch on top of this one to see how it would be used in practice.

### DIFF
--- a/frontend/builder.go
+++ b/frontend/builder.go
@@ -2,6 +2,7 @@ package frontend
 
 import (
 	"math/big"
+	"reflect"
 
 	"github.com/consensys/gnark/backend/hint"
 	"github.com/consensys/gnark/frontend/schema"
@@ -65,6 +66,10 @@ type Builder interface {
 
 	// Compile is called after circuit.Define() to produce a final IR (CompiledConstraintSystem)
 	Compile() (CompiledConstraintSystem, error)
+
+	// VariableCount returns the number of native elements required to represent
+	// the given reflected type as a witness.
+	VariableCount(reflect.Type) int
 
 	// SetSchema is used internally by frontend.Compile to set the circuit schema
 	SetSchema(*schema.Schema)
--- a/frontend/builder.go
+++ b/frontend/builder.go
@@ -76,9 +76,9 @@ type Builder interface {
 
 	// AddPublicVariable is called by the compiler when parsing the circuit schema. It panics if
 	// called inside circuit.Define()
-	AddPublicVariable(name string) Variable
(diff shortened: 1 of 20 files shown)


gbotrel (approved):
looks OK to me, not sure where this is going but doesn't look like to be introducing any regressions 👍 

```
// VariableCount returns the number of native elements required to represent
 	// the given reflected type as a witness.
 	VariableCount(reflect.Type) int
 ```
 
Don't understand why this is in Builder interface.

ivokub:
> looks OK to me, not sure where this is going but doesn't look like to be introducing any regressions 👍

Just in case on how I would be using it. When I have defined variables in a circuit as type `frontend.Variable` (for compatibility purposes), but in practice them with `nonnative.Elements`, then I need to modify the schema field in-place in `Builder.Add{Public/Secret}Variable` method. See for example https://github.com/ConsenSys/gnark/blob/1fcebc0dcae8a0d7f06c9efe869834141cdc94be/std/math/nonnative/api.go#L62. I'm modifying in-place the schema field from `Leaf` type to `Slice` type to account that nonnative element needs several variables for storing the limb values.

> 
> ```
> // VariableCount returns the number of native elements required to represent
>  	// the given reflected type as a witness.
>  	VariableCount(reflect.Type) int
> ```
> 
> Don't understand why this is in Builder interface.

So, if I wrap the native builder (r1cs or scs) for using  non-native field emulation (`nonnative.Builder`), then I need several native elements/circuit variables to represent nonnative elements. Currently compiler runs schema parser in two rounds. In first round to get number of (public/secret) wires and to shift the wire IDs correctly when initialising variable in second round. I didn't see another nice option to count the number of variables properly otherwise.

For example usage, see https://github.com/ConsenSys/gnark/blob/1fcebc0dcae8a0d7f06c9efe869834141cdc94be/std/math/nonnative/api.go#L37

Or actually, maybe I could do without `Builder.VariableCount`, but then I would have to compute the wire IDs lazily or somehow remove the necessary shifting for private variables?
I can probably figure it out but don't know if it makes sense. What do you think?

gbotrel:
`VariableCount` is fine, this is not a user-facing API, so it's OK if we end up refactoring in the future 👍

PR #347, opened 2022-07-21
feat: added serialization header to CS and debug info to all constraints with -tags=debug

gbotrel:
- feat: added GnarkVersion semver tag in cs serialized objects
- feat: store debug info stack for all constraints if debug tag is set
- doc: added TODO
- fix: round trip serialization test of cs
- fix: stack parsing

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -17,7 +17,12 @@
 // https://docs.gnark.consensys.net
 package gnark
 
-import "github.com/consensys/gnark-crypto/ecc"
+import (
+	"github.com/blang/semver/v4"
+	"github.com/consensys/gnark-crypto/ecc"
+)
+
+var Version = semver.MustParse("0.8.0-alpha")
 
 // Curves return the curves supported by gnark
 func Curves() []ecc.ID {
--- a/frontend/compiled/cs.go
+++ b/frontend/compiled/cs.go
@@ -3,6 +3,7 @@ package compiled
 import (
 	"fmt"
 	"math/big"
+	"runtime"
 	"strings"
 
 	"github.com/blang/semver/v4"
@@ -39,7 +40,10 @@ type ConstraintSystem struct {
 
 	// debug info contains stack trace (including line number) of a call to a cs.API that
 	// results in an unsolved constraint
-	DebugInfo []LogEntry
+	DebugInfo       []LogEntry
+	DebugStackPaths map[int]string    // maps unique id to a path (optimized for reading debug info stacks)
+	debugPathsIds   map[string]uint32 // maps a path to an id (optimized for storing debug info stacks)
+	debugPathId     uint32
 
 	// maps constraint id to debugInfo id
 	// several constraints may point to the same debug info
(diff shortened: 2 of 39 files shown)


frontend/compiled/cs.go
```diff
@@ -34,12 +41,17 @@ type ConstraintSystem struct {
 	// debug info contains stack trace (including line number) of a call to a cs.API that
 	// results in an unsolved constraint
 	DebugInfo []LogEntry
+	// maps unique id to a path (optimized for reading debug info stacks)
+	DebugStackPaths map[uint32]string
```
ivokub, line 45: Hmm. Would it make sense to prevent serializing `DebugInfo`, `Logs` like `debugPathsIds` and `debugPathId`? I think the use-case for debug info is debugging the circuit locally and then we can assume that we haven't had serialized->deserialized step?

frontend/compiled/cs.go
```diff
@@ -147,3 +192,61 @@ func (cs *ConstraintSystem) AddDebugInfo(errName string, i ...interface{}) int {
...
+
+func (cs *ConstraintSystem) GetDebugInfo() ([][]uint64, map[uint32]string) {
+	r := make([][]uint64, len(cs.DebugInfo))
+	for _, l := range cs.DebugInfo {
+		r = append(r, l.Stack)
+	}
+	return r, cs.DebugStackPaths
+}
+
+func (cs *ConstraintSystem) stack() (r []uint64) {
```
ivokub, line 204: Method very similar to `debug.WriteStack()`. Would make sense to refactor the common part into separate method and then `debug.WriteStack()` would just expand the stack entries into string builder?

ivokub (commented):
I think the PR makes sense.

doc.go
```diff
@@ -17,7 +17,12 @@
 // https://docs.gnark.consensys.net
 package gnark
 
-import "github.com/consensys/gnark-crypto/ecc"
+import (
+	"github.com/blang/semver/v4"
+	"github.com/consensys/gnark-crypto/ecc"
+)
+
+var Version = semver.MustParse("0.8.0-alpha")
```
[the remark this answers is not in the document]
gbotrel: 1. Experimented with the `debug.BuildInfo` (go 1.18) but what it embeds is where you are at. So if you are on develop, you'll have "devel". And not in test mode either, only if you build a binary. 

2. That's what we did for `goff` (now merged into gnark-crypto). Can do something similar 👍

frontend/compiled/cs.go
gbotrel: No, `DebugInfo` and `Logs` can be used after deserialization; say you have a prover service, and it solves with invalid witness, these 2 objects will help produce a readable error. 

+ logs is for the `api.Println` that you may wish to log or what not.

frontend/compiled/cs.go
gbotrel: yes thought about that but the "debugStack" new version, just needs file and line number. The other debug stack for assertion failure, also collects function names. And depending on build tag, filter gnark internals or not.  

Will revisit after building the constraint viewer tool.

ivokub (approved):

PR #354, opened 2022-08-03
replaces full bit decomposition by rshift in emulated/assertIsEqual

gbotrel:
In emulated/assertLimbEquality, previously we had:

```
		diffBits := bits.ToBinary(api, diff, bits.WithNbDigits(int(nbBits+nbCarryBits+1)), bits.WithUnconstrainedOutputs())
		for j := uint(0); j < nbBits; j++ {
			api.AssertIsEqual(diffBits[j], 0)
		}
		carry = bits.FromBinary(api, diffBits[nbBits:nbBits+nbCarryBits+1])
```

but the bits in the binary decomposition comes from a hint; so having the "lowest bits" always constrain to 0 and not using them later on in the circuit, is a waste; it's "we have a input wire in the circuit that must always be 0" --> we don't need it. 

Instead this PR replaces this by a "clean" rshift; it is essentially the same code but discard the low bits in the binary decomposition.


ivokub (approved):
Looks 👍.

PR #361, opened 2022-09-07
Feat/polynomial

Tabaie:
Just some polynomial related functionalities that are used by a GKR verifier:

* Evaluating a multilinear polynomials at any point given its evaluation on the hypercube
* Evaluating a polynomial at any point given its coefficients
* Evaluating (interpolating) a low-degree extension at any point given its evaluation on small nonnegative integers
* Evaluating a multilinear polynomial that tells if two binary vectors are equal, at any point

### DIFF
--- a/std/hints.go
+++ b/std/hints.go
@@ -21,7 +21,7 @@ func RegisterHints() {
 }
 
 func registerHints() {
-	// note that importing these packages may already triggers a call to hint.Register(...)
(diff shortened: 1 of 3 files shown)


gbotrel:
@Tabaie  why is this not in `std/math/polynomial` ?

Tabaie:
It should be. I'll do it in the next PR.

PR #368, opened 2022-09-13
fix(emulated): enforce widths of packed limbs

ivokub:
When constructing a new emulated element using `PackLimbs`, we manually checked that the whole width of the constructed element corresponds to the modulus. However, `PackLimbs` was only used in contexts where we constructed an element from a single limb. Then it makes more sense to check only that it is constructed from a limb which is in range.


ivokub:
The tests failed due to stats difference. It seems that this fix increases the number of internal variables and constraints for field emulation stats circuit. But after reviewing it seems that it is correct as previously the limbs output by hints were not width-constrained. Now they are implicitly constrained by the `PackLimbs` function.

gbotrel (approved):

PR #372, opened 2022-09-21
feat: add ECDSA signature verification

ivokub:
Initial PR for discussion. If the API looks OK, will add documentation and merge for now.

This PR implements also emulated short-Weierstrass package, which for now assumes the special case for secp256k1 (a4=0), but in general can be generalised to any a4. I didn't want to do it now because then we would need to have more parameters etc. and I didn't want to eagerly generalise.

A few things to do before merge:
* [x] this PR is slightly larger and contains other fixes to field emulation. Should we separate this into three - this, emulated SW and field emulation bugs?
* [x] add package documentation and code examples
* [ ] add ECRecover?
* [x] i'm leaking the internal implementation of the field emulation package by type asserting `frontend.Variable` to `emulated.Element[T]`. I think long-term would be better to have something which returns actual types not empty interface.

This implementation right now is very heavy and requires 3.7M constraints without range checks, but I think it is a good base for optimisation. There are several optimisation options:
* implement separate variable and base point scalar mul. Base point scalar mul can take advantage of huge precomputation tables + lookups.
* variable point scalar multiplication requires building lookup tables per-point. For a single signature this allows to implement windowed scalar multiplication and reduce the cost several times. See https://0xparc.org/blog/zk-ecdsa-2.
* [0xPARC blog](https://0xparc.org/blog/batch-ecdsa) describes an interesting idea for checking only a random linear combination of signatures. This would require applying Fiat-Shamir and a lot of hashing, but may probably be more efficient than point operations.
* When we combine two lookup tables, then can build window over two scalars at a time for ECDSA.

**DRAFT**: should be merged after #395.


gbotrel:
@yelhousni can you do a review of that one before merge?

ivokub:
Merged develop and fixed conflicts. It is good to merge from my side.

@yelhousni said that will post optimisations in separate PR.

yelhousni (approved):
Looks great! 👍
We can merge this PR and track optimizations in separate PRs.

gbotrel:
@yelhousni should we merge now or wait for replacing the external dependency for out-of-circuit ecdsa with latest `gnark-crypto` ?

yelhousni:
> @yelhousni should we merge now or wait for replacing the external dependency for out-of-circuit ecdsa with latest `gnark-crypto` ?

To do so we have to wait for https://github.com/ConsenSys/gnark-crypto/pull/310 to be merged I guess. So either we wait or we merge this and follow with a quick PR for the dependency.

PR #376, opened 2022-10-04
Optimize bn254/groth16 solidity verifier.

citizen-stig:
# Problem

Existing implementation of Pairing library allocated additional memory on each addition and multiplication, which can quadratically increase gas cost with larger public input.

This PR addresses several things:

1. Allocates single array for each operation and reuses them between
2. Changes declaration of `input` from `memory` to `calldata`. Ideally public input should be constructed off-chain and remain only in call data. It is important to notice, that caller smart contract might already do it, but EVM implicitly copy call data to memory, which is not needed in this case.

### DIFF
--- a/internal/backend/bn254/groth16/solidity.go
+++ b/internal/backend/bn254/groth16/solidity.go
@@ -242,7 +242,7 @@ contract Verifier {
         uint256[{{sub $lenK 1}}] calldata input
     ) public view returns (bool r) {
 
-        Proof memory proof
+        Proof memory proof;
         proof.A = Pairing.G1Point(a[0], a[1]);
         proof.B = Pairing.G2Point([b[0][0], b[0][1]], [b[1][0], b[1][1]]);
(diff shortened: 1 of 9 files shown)


internal/backend/bn254/groth16/solidity.go
```diff
@@ -217,10 +263,42 @@ contract Verifier {
...
+        {{- range $i, $ki := .G1.K -}}
+            {{- if lt $i 1}}
+        // last
+        add_input[0] = add_return.X;
+        add_input[1] = add_return.Y;
+        add_input[2] = uint256({{$ki.X.String}});
+        add_input[3] = uint256({{$ki.Y.String}});
+        Pairing.plus_raw(add_input, add_return);
+             {{- end -}}
+        {{- end }}
```
gbotrel, line 301: test were failing because; vk_x is used in the pairing computation, while the result from add_return is not. 
Will merge and do the quick fix on develop

PR #377, opened 2022-10-06
style: remove one add in groth16 solidity verifier, minor refactor

gbotrel:
Following #376 , some minor changes, see diff.

* more comments
* remove unused IC (vk.K) from the VerifyingKey structure
* remove "//last" iteration; initialize the input to accumulate function with vk.K[0] instead.

@citizen-stig : questions
1. Does it matter to inline the for loop? (ie use code generation for each iteration, rather than have them in the for loop to have a compact size if we have many inputs?)
2. Does it matter to use code generation to set the values of the vk.K (IC) verifying key directly before calling the scalar mul / point addition? (previously was refering the verifyingKey data structure in memory)

### DIFF
--- a/internal/backend/bn254/groth16/solidity.go
+++ b/internal/backend/bn254/groth16/solidity.go
@@ -197,7 +197,7 @@ contract Verifier {
         Pairing.G2Point beta2;
         Pairing.G2Point gamma2;
         Pairing.G2Point delta2;
-        Pairing.G1Point[{{$lenK}}] IC;
+        // []G1Point IC (K in gnark) appears directly in verifyProof
     }
 
     struct Proof {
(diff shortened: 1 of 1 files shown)


citizen-stig:
> Does it matter to inline the for loop? (ie use code generation for each iteration, rather than have them in the for loop to have a compact size if we have many inputs?)

Could you please clarify which for loop?

> Does it matter to use code generation to set the values of the vk.K (IC) verifying key directly before calling the scalar mul / point addition? (previously was refering the verifyingKey data structure in memory)

What would be alternative? With code generation, all the data is stored in code of smart contract, and `vk_x` memory is just re-used between calls.

gbotrel:
I meant this for loop:
```
// Make sure that every input is less than the snark scalar field
        for (uint256 i = 0; i < input.length; i++) {
            require(input[i] < SNARK_SCALAR_FIELD,"verifier-gte-snark-scalar-field");
        }
```
Before your PR, the scalar multiplication + point addition was done in it. After your PR, the for loop is unrolled through code generation and each iteration is code generated. Does it help with perf? If not I'll put it back in the for loop, since some circuits could have much more that 1 public input the code will be very verbose.

The second point was about how the vk values are referenced, before your PR, it was verifyingKey.IC[i], after your PR, it's uint256(x), with x being code generated directly to the correct value. Same question, does it matter for perf ?

citizen-stig:
Oh, I remember. Both of these are related to problems in solidity, when solc wasn't able to compile such contracts because of optimisations was taking enormous amount of time: https://github.com/ethereum/solidity/issues/12406

I don't remember if I had verified that improvement works. But I think it is worth trying. If bringing forloop back, then minimal solidity version should be `0.8.15`, when fix was merged.

PR #381, opened 2022-10-10
Changes on the tensor commitment resulting from the integration with the Wizard

AlexandreBelling:
# Changes on the tensor commitment

Goes together with : https://github.com/ConsenSys/gnark-crypto/pull/247

* Update the test which are impacted by the PR on gnark-crypto
* Move the FFT inverse out of the loop because otherwise it is a huge loss in performances

# Remaining issues

## BUG : the padding strategy of the gnark/std/sis.RSisHash is to preprend with zeroes, while the gnark-crypto's strategy is to append.

(Also pointed out in the gnark-crypto's PR)
* There should be a test for this edge-case

## In `api.Println(v)`

If `v` is a slice of `frontend.Variable()` and it contains a constant, it will panic.
Note that `api.Println(v...)` and `api.Println(frontend.Variable(0))` will work fine.

## `tensorcommitment.Proof.GenBigDomainTensorCommitment` is a big.Int

You can be sure that the user will systematically call `api.ToBigInt` instead of `api.ToBigIntRegular`. It would be worth finding a way to not let the user get a chance to make this mistake.

## ~~`api.Println` will not print in my tests unless I pass the `-tags=debug` flag~~

~~I can expand more on this if you think that this is worth spending time on, it does not seem to be an intended behaviour. It seems that Azam does not have it on her tests, so the error probably comes from my setup. Could be Linux vs Mac related.~~


std/commitments/tensor-commitment/verifier.go
```diff
@@ -173,13 +180,7 @@ func Verify(api frontend.API, proof Proof, digest [][]frontend.Variable, l []fro
 			linCombEncoded = api.Add(linCombEncoded, tmp)
 		}
 
-		// entry i of the encoded linear combination
-		// first we express proof.LinearCombination in canonical form
-		// then we evaluate it at the required queries
-		linCombCanonical := FftInverse(api, proof.LinearCombination, proof.GenInvSmallDomainTensorCommitment, proof.SizeSmallDomainTensorCommitment)
-
-		var encodedLinComb frontend.Variable
-		encodedLinComb = evalAtPower(
```
AlexandreBelling, line 182: Just my linter suggesting this

PR #387, opened 2022-10-22
feat: add inherit circuit tag

### LINKED ISSUE #385: Variable tag to inherit parent visibility
Reported in https://github.com/ConsenSys/gnark/discussions/384.

I think the approach is to define flag 'inherited', which applies the visibility of the parent object to the current object. If parent is unset, then returns an error.

@gbotrel, I recall having similar discussion about this in the past. Was there any reason not to add this tag?

ivokub:
Resolves #385.

This PR adds new "inherit" tag to the circuit which allows inherit the visibility of the parent object. Useful for custom defined types.

Meanwhile, updated also some comments to take advantage of new go doc formatting rules (links and lists).

### DIFF
--- a/frontend/circuit.go
+++ b/frontend/circuit.go
@@ -16,28 +16,24 @@ limitations under the License.
 
 package frontend
 
-// Circuit must be implemented by user-defined circuits
+// Circuit must be implemented by user-defined circuit. The best way to define a
+// circuit is define a type which contains all the witness elements as fields
+// and declare `Define` method on the type.
 //
-// the tag format is as follow:
-//
-//	type MyCircuit struct {
-//		Y frontend.Variable `gnark:"name,option"`
-//	}
-//
-// if empty, default resolves to variable name (here "Y") and secret visibility
-// similarly to json or xml struct tags, these are valid:
-//
-//	`gnark:",public"` or `gnark:"-"`
-//
-// using "-" marks the variable as ignored by the Compile method. This can be useful when you need to
-// declare variables as aliases that are already allocated. For example
+// For example, the following is a minimal valid circuit:
 //
 //	type MyCircuit struct {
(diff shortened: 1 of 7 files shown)


ivokub:
There was before some inconsistency how the tags were parsed. If there was not tag, then we set the visibility from the parent (https://github.com/ConsenSys/gnark/blob/develop/frontend/schema/schema.go#L256 and https://github.com/ConsenSys/gnark/blob/develop/frontend/schema/schema.go#L281). If there was only a name tag `gnark:"X"`, then we set the visibility to secret (https://github.com/ConsenSys/gnark/blob/develop/frontend/schema/schema.go#L264). The tests didn't have the edge case.

I'm trying to make the behaviour consistent and keep the existing promise. So, on top-level if there is no tag, then the visibility is left to `Unset`, but this resolves (even if the top-level element has sub-elements) in the variable parser to `Secret`. In every other case when there is not tag, we assume the visibility of the parent.

If there is only name tag, then for top-level variables we assume `Secret` visibility and lower levels inherit from the parent (this is the difference, previously we set the visibility to `Secret`).

Added also explicit "inherit" tag option which is like empty one, but which can not be set for the top-level element. I made this change before figuring out the name-only tag cases and now seems to be slightly redundant. But dunno, maybe is useful.

Question -- shouldn't we in the case when there is not tag and we are top-level, set the visibility to `Secret` eagerly? E.g. when we hit the case https://github.com/ConsenSys/gnark/blob/develop/frontend/schema/schema.go#L280. It should work now also nicely as it expected, but we take slightly different paths (compared to when we have name-only tag) and this could at some point create inconsistent behaviour.

gbotrel (approved):

drakstik:
This should be mentioned in the turotial here:   https://docs.gnark.consensys.io/HowTo/write/circuit_structure

PR #388, opened 2022-10-25
feat: add lookup table

ivokub:
Adds AS-Waksman permutation network (for internal use only for building gadgets).

We implement a simple append-only lookup table using it. We could probably also to a full RW memory possibly.


ivokub:
FYI @Tabaie, I updated the PR description about possible improvements.

Tabaie:
Does this implement a particular paper?

ivokub:
> Does this implement a particular paper?

The AS Waksman routing network is described in [On Arbitrary Waksman Networks and their
Vulnerability](https://hal.inria.fr/inria-00072871/document) by Beauquier and Darrot. It describes how to connect the gates in layers, the tricks for the lower gates and splitting to obtain the arbitrary-sizeness (compared to previous results where the number of inputs had to be power of two). It also describes bipartite coloring for deciding the routing.

Tabaie (dismissed):
Apologies for the long delay!

ivokub:
Oh - I liked the PR, but there is probably no need for it anymore when we have a lot more efficient methods. It didn't work well neither.

PR #392, opened 2022-11-09
fix: fixed a bug in the emulated field's _select function

kevjue:
Gnark will throw an array access out of bounds error in the __select_ function if the _a_ and _b_ parameter has more than one limb.


ivokub:
Thanks for the PR, @kevjue! I had also realised a problem with Select. There is actually slightly more complexity involved when the overflows and limb lengths are different. In PR #395 there is a [rewrite of Select](https://github.com/ConsenSys/gnark/blob/c6951b22b31d325fc0d6c10b112bf76cdb4b0b6b/std/math/emulated/field_ops.go#L285) which works for any inputs. We will hopefully merge that soon.

PR #399, opened 2022-11-17
keccakf permutation

Soleimani193:
This is the implementation of keccakF the permutation in the keccak hash function (in gnark API).
keccakhash divides the messages into blocks, for each block it calls the permutation keccakF and absorbs the block into the permutation. 
 The implementation only concerns the permutation of  keccakF (but not the full keccakHash).

The file keccak includes
keccakf: this is keccakF is ganrk API
keccakf_test: for testing keccakf
purekeccakf: this is the original keccakF (a copy-past, since the function was not exportable) that is used only for testing aim
atomic_test: this are set of small tests for functions used in the keccakf


ivokub:
Taking this PR. I'll rewrite it a little and push onto a new branch.

ivokub:
I rewrote the permutation in #401. I introduced new type `xuint64` which is like `uint64`, but as circuit variable. In practice, it is actually array of field elements of length 64. As we perform only binary ops on them then know they are boolean and do not have to re-assert.

The compiler got a bit further now, but it still doesn't compiler the full circuit (at least in a few minutes, then I stopped the process). We have actually already struggled with similar problem in the past, but were unsuccessful. Now it becomes more urgent.

PR #401, opened 2022-11-19
feat: keccak-f permutation function

ivokub:
Based on #399, but rewrote a bit to adapt for possible future byte-wise boolean ops.

PR #403, opened 2022-11-25
feat: parametrise frontend

ivokub:
A quick try at parametrising the frontend types.

I ran into problems when working with large linear expressions which appear in zk-unfriendly circuits (for example #401). The problem was that linear expressions are huge and in additions we have to add coefficients wire-by-wire. And this created two issues:

1. the coefficients were stored in a lookup table for avoiding redundancy and having compact terms (in linear expressions). However, when reducing LEs, then some coefficients became unused. But we cannot remove them from the lookup table as the compiler has no knowledge if the inputs are used in other operations. In #401 context it meant that the size of lookup table grew to 10M entries very quickly (after 15K constraints) and lookups became very expensive.

2. we perform a lot of reductions which incur super-linear cost in the size of linear expressions.

So, I was trying to solve both of the issues. To get rid of the lookup tables, I redefined `Term` to include the coefficient directly (not through ID and table). Additionally, I parametrised everything so that the coefficients would be the actual underlying type (`fr.Element` for scalar fields).

For the second problem, I implemented a feature to record a linear expression in the circuit to replace it with a single term. I did this by creating a new variable `h` and then defining constraint `(\sum_{i=0}^{huge} c_i a_i) * 1 = h` and then returning `h` instead. The parameter when to do this is configurable (and the default behaviour is not to "compress" LEs) and can be provided to compiler as `frontend.WithCompressThreshold` option. This provides satisfactory result - for keccak-f permutation with threshold 500 the incurred overhead (in the number of constraints) is around 9%. But compared to the previous situation where I was unable to compile the circuit at all (and now compiles in 30s), is satisfactory (at least for now).

PR is not yet ready:
- [ ] some tests fail (but they seem easy to fix),
- [ ] I leak the type parameters slightly into backend (I'm not sure if it actually hurts, because the type parameters are fixed for the curves. In any case would be easy to fix with code generation)
- [ ] I lost some optimisations in the backend (I use the same `Term` definition as for frontend, but for the backend we can actually build the coefficient table and use it. I also lost batch-invert in plonk backend because lacking coefficient table)
- [ ] lacking documentation for new definitions
- [ ] could use more pool for element reuse (but need to look at the assembly to figure out when allocated on heap and when not)
- [ ] frontend has still a lot of *big.Ints, but want to get rid of them (for constants etc.).

I haven't benchmarked on larger circuits yet, but tests seem to be working not-too-slow (subjectively, at least). @gbotrel, can give some feedback if is worth fixing.


gbotrel:
Will visit this PR and experiment locally today :) 

> I lost some optimisations in the backend (I use the same Term definition as for frontend, but for the backend we can actually build the coefficient table and use it. I also lost batch-invert in plonk backend because lacking coefficient table)

--> that's a must have 👍

frontend/compiled/term.go
```diff
@@ -175,3 +171,26 @@ func (t Term) string(sbb *strings.Builder, coeffs []big.Int) {
 	}
 	sbb.WriteString(strconv.Itoa(t.WireID()))
 }
+
+func (t Term[E, ptE]) HashCode() uint64 {
+	return t.Var
```
gbotrel, line 176: @ivokub that seems too little? can we afford to do a better hashcode with the actual coeff limbs values ?

frontend/compiled/term.go
```diff
@@ -175,3 +171,26 @@ func (t Term) string(sbb *strings.Builder, coeffs []big.Int) {
...
+
+func (t Term[E, ptE]) IsZero() bool {
+	return ptE(&t.Coeff).IsZero()
+}
+
+func (t Term[E, ptE]) IsOne() bool {
+	return ptE(&t.Coeff).IsOne()
+}
+
+func (t Term[E, ptE]) IsNegOne() bool {
```
gbotrel, line 187: this probably does 1 heap alloc per call right?

frontend/cs/r1cs/builder.go
```diff
@@ -88,80 +114,88 @@ func newBuilder(field *big.Int, config frontend.CompileConfig) *r1cs {
...
 
-	mod := system.Field()
-	c := new(big.Int)
 	for i := 1; i < len(l); i++ {
 		pcID, pvID, pVis := l[i-1].Unpack()
 		ccID, cvID, cVis := l[i].Unpack()
 		// if the coefficient is zero, we remove it
-		if ccID == compiled.CoeffIdZero {
-			l = append(l[:i], l[i+1:]...)
+		if l[i].IsZero() {
```
gbotrel, line 168: why do we have coefficients at zero here? I can understand this to happen **after** the reduce, since we could add / sub coeffs with the result == 0, but before ?

gbotrel:
Did add couple of comments -- not sure this patch is not hurting other cases; for example, in `std/groth16_bls12377`, running `BenchmarkCompile` against develop:

```
benchmark               old ns/op     new ns/op      delta
BenchmarkCompile-10     394783403     1292244583     +227.33%

benchmark               old allocs     new allocs     delta
BenchmarkCompile-10     2952001        15556641       +426.99%

benchmark               old bytes     new bytes      delta
BenchmarkCompile-10     210047573     1595247472     +659.47%
```

frontend/compiled/term.go
gbotrel: never mind I recall now that we only care about sorting linear expressions with respect to variable id

frontend/compiled/term.go
ivokub: Yup, does allocate. Due to type parameter cannot init once easily. In many cases we use this method in the constraint system, I guess could store -1 there?

frontend/cs/r1cs/builder.go
ivokub: Not fully sure, didn't think about it. I guess this may have happened after reduce but for some reason the check was here? Makes sense to move it to post-reduce. Then would have to count nb of zeros though because now construct the new linear expression in one pass/allocation.

ivokub:
> Did add couple of comments -- not sure this patch is not hurting other cases; for example, in `std/groth16_bls12377`, running `BenchmarkCompile` against develop:
> 
> ```
> benchmark               old ns/op     new ns/op      delta
> BenchmarkCompile-10     394783403     1292244583     +227.33%
> 
> benchmark               old allocs     new allocs     delta
> BenchmarkCompile-10     2952001        15556641       +426.99%
> 
> benchmark               old bytes     new bytes      delta
> BenchmarkCompile-10     210047573     1595247472     +659.47%
> ```

Yup, for the same benchmark I also had worse results. But I think there is still room for improvement -- could use pool for `fr.Element` which are used in `Term` and pass them back after reduce. I was still thinking about experimenting with using `fr.Element` pointers in `Term` as I had a hunch that we allocate too much with `Term.Unpack()` etc etc. But didn't start looking too much into it for now.

ivokub:
> Will visit this PR and experiment locally today :)
> 
> > I lost some optimisations in the backend (I use the same Term definition as for frontend, but for the backend we can actually build the coefficient table and use it. I also lost batch-invert in plonk backend because lacking coefficient table)
> 
> --> that's a must have 👍

Definitely! I think actually that this would be a great opportunity to fully split frontend and backend. Make types `frontend/compiled` private for frontend only and then define the corresponding non-generic generated types in backend (with the table-based coefficients etc). Then can easily reintroduce the optimisations I removed.

ivokub:
This PR is superseded by #412. Keeping it now for porting bda92771794405ebbd43dbb482d6c6443bc8d7b9 on top.

PR #405, opened 2022-12-01
Feat/g16 enormous fs hint

Tabaie:


frontend/cs/r1cs/api.go
```diff
@@ -576,3 +576,19 @@ func (system *r1cs) negateLinExp(l compiled.LinearExpression) compiled.LinearExp
 func (system *r1cs) Compiler() frontend.Compiler {
 	return system
 }
+
+func (system *r1cs) Commit(v ...frontend.Variable) (frontend.Variable, error) {
+
```
gbotrel, line 581: need to ensure we do it only once? since it uses a singleton (system.CommitmentInfo). But ... technically we could have (if needed) a list of []CommitmentInfo right?

frontend/compiled/commitment.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, fmt.Errorf("commitment variable index smaller than some committed variable indices")
+	}
+
+	i.HintID = hint.UUID(bsb22CommitmentComputePlaceholder)
+
+	return commitment, nil
+}
+
+func bsb22CommitmentComputePlaceholder(*big.Int, []*big.Int, []*big.Int) error {
+	return fmt.Errorf("placeholder function: to be replaced by commitment computation")
```
gbotrel, line 68: why is this just a placeholder? Would be nice to have a default impl, so that calling R1CS.Solve on a R1CS with commitment info actually works?

gbotrel: (that is, without depending on Groth16)

frontend/compiled/commitment.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	return res
+}
+
+func (i *Info) SerializeCommitment(privateCommitment []byte, publicCommitted []*big.Int, fieldByteLen int) []byte {
+	buf := bytes.NewBuffer(privateCommitment)
+	for _, inI := range publicCommitted {
+		inIBytes := inI.Bytes()
+		slack := fieldByteLen - len(inIBytes)
+		buf.Write(make([]byte, slack))
```
gbotrel, line 84: in the case of many public commited values, would probably be beneficial to allocate an empty []byte of len fieldByteLen before the loop, and write `slack[:slackIdx]` here

frontend/compiled/commitment.go
Tabaie: The problem is, this function computes the commitment, so it requires the Commitment Key, created during setup. But this hint is created in compilation time, so the Commitment Key is not available.

The only way out of this that I can think of is to instead randomize the Commitment Key during compilation. But that would break modularity by making setup-type tasks bleed into compilation, and it would also bring a big code-generated curve switch-case into the compilation codebase, which as I understand currently doesn't have any.

frontend/cs/r1cs/api.go
Tabaie: It's definitely ideal for different subcircuits to be able to get different commitments. Maybe something like
```go
func Commit(uid []byte, v ...frontend.Variable) (frontend.Variable, error)
```
that could be called multiple times.
I think it's worth trying it but would involve even more modifying-a-hint-after-it's-created type shenanigans. I'll create an issue and once I'm a bit more secure that my current hint hacks are okay, I'll take a stab at it.

But in the meantime I agree there should be an error if `Commit` is called more than once.

frontend/compiled/commitment.go
Tabaie: Pushed something in that spirit!

frontend/compiled/commitment.go
gbotrel: well, what I'm saying is that, technically, the frontend should build a constraint system, that's solvable as is, without going to the next step (proof systems).
So why not just put a hint that do a sha256 or something deterministic with the inputs, without a "key"?

Tabaie: Sha256 is only possible for public variables to which the verifier has direct answer. For private variables we will need that Pedersen commitment whose basis comes from the verifying key. Unfortunately this spaghetti quality, that weakens the modular frontend/backend separation is inherent to the scheme, and actually part of the reason why it works.

PR #410, opened 2022-12-08
Update Builder.go O(N^2) ---> O(N)

niconiconi:
Works way faster when there are a lot of api.Add() at the same time (and a lot of redundancy)


gbotrel:
hi -- thanks for the PR; we did notice the same issue and a fix will be merged on develop in the coming days. It is part of a refactoring that greatly reduce memory pressure at Compile() time.

PR #421, opened 2022-12-20
feat: add debug.SymbolTable into constraint system for storage efficiency of debug info

gbotrel:
Clearing that from backend; don't store stack trace in string (debug info) in constraint systems, just store ids to a symbol table and format at run time.

Additionally, don't store the stack at all for `AssertIsBoolean` unless debug build tag is set since we can have a huge amount of these.

Example impact on bn254 pairing:
```
benchmark                                  old ns/op        new ns/op       delta
BenchmarkFinalExponentiationCompile-10     148012806625     42604127375     -71.22%

benchmark                                  old allocs     new allocs     delta
BenchmarkFinalExponentiationCompile-10     919164251      534771532      -41.82%

benchmark                                  old bytes        new bytes       delta
BenchmarkFinalExponentiationCompile-10     121579914912     78784978536     -35.20%
```


debug/symbol_table.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+type SymbolTable struct {
+	Locations  []Location
+	Functions  []Function
+	mFunctions map[string]int `cbor:"-"` // frame.File+frame.Function to id in Functions
```
ivokub, line 12: frame.Function is fully qualified name of the function. Do we need to store also the file name?

ivokub (approved):
Looks good. Only the map for functions in the debug symbol table has filename+methodname as keys, but imo as methodname is fully qualified, then I think can use only methodname. But I think it doesn't make any measurable difference in memory usage, as the debug stacks are way larger anyway.

debug/symbol_table.go
gbotrel: this is derived from the `profile/` package which is itself derived from go std, don't think it matters too much indeed :)

PR #423, opened 2022-12-20
Restore frontend.WithCapacity (to merge after perf/debugInfo)

gbotrel:


ivokub (approved):
Looks good. I think it would be possible to tweak the 0.1 multiplier for number of boolean linear expressions and coefficients, but this is quite circuit-specific.

PR #439, opened 2023-01-18
fix: pack full limbs for quotient

### LINKED ISSUE #426: bug: PackLimbs in field emulation assumes input is less than emulated modulus
PackLimbs is used to construct a new emulated element + enforce the limb widths. However, it is also used to enforce the output of `QuoHint` hint function, but its output may be larger than modulus.

It is better to have two versions of this method `PackElementLimbs` and `PackFullLimbs` where former assumes that input is smaller than modulus and otherwise that limbs are just smaller than `NbBits` parameter.

Ref: https://github.com/ConsenSys/gnark/discussions/420

ivokub:
Closes #426.

### DIFF
--- a/std/math/emulated/field_assert.go
+++ b/std/math/emulated/field_assert.go
@@ -122,6 +122,9 @@ func (f *Field[T]) enforceWidth(a *Element[T], modWidth bool) {
 			panic("constant limb width doesn't match parametrized field")
 		}
 	}
+	if modWidth && len(a.Limbs) != int(f.fParams.NbLimbs()) {
+		panic("enforcing modulus width element with inexact number of limbs")
+	}
 
 	for i := range a.Limbs {
 		limbNbBits := int(f.fParams.BitsPerLimb())
--- a/std/math/emulated/field_assert.go
+++ b/std/math/emulated/field_assert.go
@@ -111,10 +111,11 @@ func (f *Field[T]) AssertLimbsEquality(a, b *Element[T]) {
 	}
 }
 
-// EnforceWidth enforces that the bitlength of the value is exactly the
-// bitlength of the modulus. Any newly initialized variable should be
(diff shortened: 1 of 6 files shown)


std/math/emulated/field_assert.go
```diff
@@ -111,22 +111,24 @@ func (f *Field[T]) AssertLimbsEquality(a, b *Element[T]) {
 	}
 }
 
-// EnforceWidth enforces that the bitlength of the value is exactly the
-// bitlength of the modulus. Any newly initialized variable should be
-// constrained to ensure correct operations.
-func (f *Field[T]) EnforceWidth(a *Element[T]) {
+// enforceWidth enforces the width of the limbs. When modWidth is true, then the
+// limbs are asserted to be the width of the modulus (highest limb may be less
```
gbotrel, line 115: I am a bit confused by the comments; 

"highest limb may be less than full limb width" --> may or must? 

Also, it's unclear; does this method guarantee that the constrain emulated element `e` is mod reduced? (`e < q`) ?

ivokub: May be. So usually the maximum width of the limb is given by the NbBits parameter (64 bits etc.). However, it usually doesn't divide emulated modulus evenly. Then the highest limb has fewer than NbBits bits. But if divides evenly, then is exactly NbBits.

This method doesn't ensure that is mod reduced. It only ensures (if argument `modWidth=true`) that the width of the element is same as the modulus. So in general it could either be `x` or `x-p`, but for the arithmetic either case is sufficient.

For ensuring that value is less than modulus, there is the `AssertIsLessEqualThan` method.

gbotrel (approved):
ok for the scope of fixing the issue 👍 We probably can do better in term of api clarity / conciseness for the `emulated` package but that will wait for a future refactoring;

PR #440, opened 2023-01-20
Refactor: remove geth dependency

yelhousni:
Use gnark-crypto instead of geth/crypto for secp256k1 and ecdsa.
This needs https://github.com/ConsenSys/gnark-crypto/pull/310 to be merged first.


yelhousni:
needs to be updated after https://github.com/ConsenSys/gnark-crypto/pull/313 and https://github.com/ConsenSys/gnark-crypto/pull/314 get merged in gnark-cypto.

ivokub:
So, right now there is a slight problem - when we sign with gnark-crypto, as a signature we get a `[]byte`, but it is difficult to get `R` and `S` from the signature. We would have to instantiate fr-specific `ecdsa.Signature`, unmarshal it from bytes and assign from it. It involves a type-switch, but I think it is ok.

I have some ideas and will try to push the changes.

ivokub (approved):
Rebased on top of develop, added a few more changes (avoid explicit conversion to *big.Int, rename package import, update go.mod).

Previously I totally overthought it. I think the simple approach is better (read signature directly into secp256k1 ECDSA signature struct and get R and S from there).

As far as it seems then the failing tests are due to MiMC hashing. Imo are unrelated to this PR.

Issue #390, opened 2022-10-29
Backend fails when subslices are of different length


Tabaie:
Code provided [here](https://github.com/ConsenSys/gnark/blob/bug/slices/test/bug_test.go).

Seems like somewhere in the backend it is assumed that all multi-dimensional slices have consistent lengths. i.e. If there is a witness variable `Slice [][]frontend.Variable` of length at least 2, it is assumed that `len(Slice[0]) == len(Slice[1])`.

One of the test routines provided to check this, [`TestInconsistentSliceLengthProving`](https://github.com/ConsenSys/gnark/blob/bug/slices/test/bug_test.go#L25), fails the `marshal/json` subtest with `panic: runtime error: index out of range [3] with length 3`.

```goroutine 5 [running]:
testing.tRunner.func1.2({0x102d1bc00, 0x1400002c7f8})
	go/1.19/libexec/src/testing/testing.go:1396 +0x1c8
testing.tRunner.func1()
	go/1.19/libexec/src/testing/testing.go:1399 +0x378
panic({0x102d1bc00, 0x1400002c7f8})
	go/1.19/libexec/src/runtime/panic.go:884 +0x204
github.com/consensys/gnark/internal/backend/bn254/witness.(*Witness).ToAssignment.func1.1(0x14000386940?, {0x102d5d4e0?, 0x14000396538?, 0x102d28b60?})
	project-path/gnark/internal/backend/bn254/witness/witness.go:147 +0x250
github.com/consensys/gnark/frontend/schema.parse({0x0?, 0x0, 0x0}, {0x14000386940?, 0x14000396538?}, {0x102d70ae8?, 0x102d5d4e0}, {0x1400039c030, 0x9}, {0x1400039c030, ...}, ...)
	project-path/gnark/frontend/schema/schema.go:223 +0x71c
github.com/consensys/gnark/frontend/schema.parse({0x14000399440?, 0x1, 0x1}, {0x14000386900?, 0x14000396530?}, {0x102d70ae8?, 0x102d5d4e0}, {0x1400039c019, 0x7}, {0x1400039c019, ...}, ...)
	project-path/gnark/frontend/schema/schema.go:341 +0xcf4
github.com/consensys/gnark/frontend/schema.parse({0x0?, 0x0, 0x0}, {0x140003868c0?, 0x14000396520?}, {0x102d70ae8?, 0x102d5d4e0}, {0x1400032df53, 0x5}, {0x1400032df53, ...}, ...)
[...]

PR #444, opened 2023-01-21
fix #442: use `reflectwalk` to walk through circuit structures without building a `Schema`

### LINKED ISSUE #442: perf: `frontend/schema.Parse` makes compile time unnecessary slow and doesn't handle some common cases well
Calls to `schema.Parse` build a schema of the circuit, counts the number of secret / public variables, and if set, calls a handler on each leaf (of type `frontend.Variable`). 

Couple of issues; 
* #390 : current version doesn't handle slices of varying length (so `[N][]frontend.Variable` where the slice lens varies currently produce an incorrect result in the schema). #433  fixes that but upon closer inspection, realize we have some technical luggage here that warrants a deeper work;
* most of the times, we don't need to create / store the graph ( schema ), we just want to walk through the structure. The ONLY use of the schema itself is to marshal / unmarshal witnesses in JSON form (currently only used in play.gnark.io) 
* performance wise, using reflection on very large complex structures can cost a lot, particularly given that this is called several time during compile / test process. consider 

```go
type struct1 struct {
[N][M][]frontend.Variable
}
type struct2 struct {
[N][][]complex // complex --> struct { [] frontend.Variable , x,y frontend.Variable}
}
```

--> we can for `struct1` have some special handling of array or slices of `frontend.Variable`. For `struct2` we need to walk through every single leaf. We may not handle this case well in a first step.

Proposed solution is to use a lightweight fork of https://github.com/mitchellh/reflectwalk which gives us a cleaner code to walk the structure, have a "fast path" for walking the circuit, and a "slow path" to build a schema (JSON witness only).

gbotrel:
Closes #442 #433 #390. 

## Description

Introduces `schema.Walk` which behaves similarly to `schema.Parse` except it doesn't build a `Schema`, just visit the leaves. 

Known issues:
* we still need `schema.Parse` to handle `json` witnesses (play.gnark.io), and it won't handle all circuit structures well
* `schema.Walk` doesn't like it if the leaf we are looking for is a `reflect.Ptr`. Only happens in... `witness` where we want to identify zero values in `json` will probably revisit when it's warranted.  

## Benchmark

Old (parse full schema) vs new (walk):
```
BenchmarkLargeSchema/walk-10                        4074            277284 ns/op           15456 B/op       4000 allocs/op
BenchmarkLargeSchema/parse-10                          1        2127391458 ns/op        3490130176 B/op 66707400 allocs/op
BenchmarkArrayOfSliceOfStructSchema/walk-10                    1        6265857250 ns/op        802062272 B/op  100257706 allocs/op
BenchmarkArrayOfSliceOfStructSchema/parse-10                   1        13797889083 ns/op       21742293688 B/op        351920063 allocs/op
```


gbotrel:
Yes behavior is unchanged, we may have some perf issues with very large complex circuits (with multi dimension slices and structs) at compile time, but nothing urgent to fix for now.

`ParseDeprecated` I will try to kill it and reuse a `walker` pattern to ensure we use the same exact code (reflection part). The only part where we use this method (probably better named as `schema.New`) is to marshal/unmarshall json witness. So for us, in play.gnark.io . -->  not critical if slow for large circuits.

PR #449, opened 2023-01-26
refactor: emulated clean up fixes #448

### LINKED ISSUE #448: cleanup `std/math/emulated`
Before focusing on "plonk-specific" optimizations in the package, would like to reduce its scope. For example the implementation of `frontend.Builder` and `frontend.API` may not be strictly necessary, since using `emulated.Element` directly seems to be more convenient in all use cases so far.

gbotrel:
- refactor: step 1 clean up api and builder wrapper
- refactor: PackElementLimbs PackFull -> packLimbs
- refactor: added field.NewElement
- refactor: remove deadcode from builder stuff, update stats with constraining inputs


std/math/emulated/field.go
```diff
@@ -86,6 +86,32 @@ func NewField[T FieldParams](native frontend.API) (*Field[T], error) {
 	return f, nil
 }
 
+// NewElement builds a new Element[T] from input v.
+//   - if v is a Element[T] or *Element[T] it clones it
+//   - if v is a constant this is equivalent to calling emulated.NewConstant[T]
```
ivokub, line 91: emulated.ValueOf

std/math/emulated/element_test.go
```diff
@@ -633,34 +633,36 @@ type ComputationCircuit[T FieldParams] struct {
...
 		x13 = f.Reduce(x13)
 	}
-	x13 = f.Mul(x13, c.X1)
+	x13 = f.Mul(x13, &c.X1)
 	if !c.noReduce {
 		x13 = f.Reduce(x13)
 	}
 
-	fx2 := f.Mul(5, c.X2)
+	// TODO @gbotrel better way to deal with constants?
```
ivokub, line 650: f.NewElement?

std/math/emulated/element.go
```diff
@@ -100,7 +91,7 @@ func (f *Field[T]) newInternalElement(limbs []frontend.Variable, overflow uint)
 // GnarkInitHook describes how to initialise the element.
 func (e *Element[T]) GnarkInitHook() {
```
ivokub, line 92: Maybe won't need GnarkInitHook anymore?

ivokub (changes requested):
In general looks good. I think we do not need GnarkInitHook anymore though?

And now there is only a single use of `Field.NewElement`. In some tests we use `emulated.ValueOf` where we actually initialise constant in-circuit and then it would be better to use `Field.NewElement` to set `internal=true`

std/math/emulated/element.go
```diff
@@ -33,26 +33,17 @@ type Element[T FieldParams] struct {
...
+// The input is converted to *big.Int and decomposed into limbs and packed into new Element[T].
+func ValueOf[T FieldParams](constant interface{}) Element[T] {
+	if constant == nil {
 		r := newConstElement[T](0)
-		r.internal = false
-		return *r
-	} else {
-		r := newConstElement[T](v)
-		r.internal = false
+		// r.internal = false
```
ivokub, line 41: Is the line intentionally commented? I think we still need for enforcing the widths of the limbs on the first use of Element?

std/math/emulated/element.go
```diff
@@ -33,26 +33,17 @@ type Element[T FieldParams] struct {
...
 		return *r
 	}
+	r := newConstElement[T](constant)
+	// r.internal = false
```
ivokub, line 45: Uncomment.

std/math/emulated/element.go
gbotrel: I'm afraid we do, at least for convenience... so a circuit like 
```
type circuit struct {
A,B,C emulated.Element
}
```
if we remove gnark init hook, then before compiling, user has to do:
```
c := circuit{}
c.A = emulated.SomethingElementInit(... nb limbs ...) 
c.B = ...
// etc 
```

ivokub: Oh, you are right.

std/math/emulated/element.go
gbotrel: why though; this path just init constants. That goes through decompose big int method. We are already guaranteed the width of the element is correct ,since it's constant and built by the package. 
Now if someone does 
```
myConstant := emulated.NewElment[T](42)
myConstant.Limbs[0] = someVariable
```
then ...

ivokub (approved):
🚀

Issue #445, opened 2023-01-23
restore `ccs.GetConstraints` and `ccs.GetSchema`


gbotrel:

gbotrel:
+ for PlonK cc @AlexandreBelling

PR #450, opened 2023-01-27
refactor: clean up witness package, introduces clean `witness.Witness` interface

gbotrel:
# Motivation for this PR

#445 and #442 . + decouple *schema.Schema from witness data structure; we only need a `schema.Schema` when doing `JSON` (un)marshaling, which happens for pretty printing or in the playground. This had a significant cost in cpu + in storage, and in 99% of cases we didn't use it. 

See `witness.ExampleWitness` for usage and `witness.Witness` interface for the gist of the PR. 

Additionally, `groth16` `plonk` and `constraint` packages now directly reason with `[]fr.Element` (as they should). 

Breaking changes;
* binary serialization of a witness now encodes `[uint32(nbPublic)|uint32(nbSecret)|uint32(nbElements)|elements...]` instead of `[uint32(nbElements)|elements...]`. `fr.Vector` manage its serialization (in gnark-crypto), but a witness needs these 2 extra uint32 to ensure consistency when serializing / deserializing. 
* tried to keep user facing API untouched.


backend/witness/vector.go
```diff
@@ -1,92 +1,288 @@
...
-	io.ReaderFrom
-	FromAssignment(assignment interface{}, leafType reflect.Type, publicOnly bool) (*schema.Schema, error)
-	ToAssignment(assigment interface{}, leafType reflect.Type, publicOnly bool)
-	Len() int
-	Type() reflect.Type
-}
-
-func newVector(field *big.Int) (Vector, error) {
-	var w Vector
+func newVector(field *big.Int, size int) (any, error) {
```
ivokub, line 20: For understanding the design - in witness we use a pointer to the fr.Vector (which is a slice). Imo slice is struct{head *ptr, cap, len int} (https://pkg.go.dev/reflect#SliceHeader) and I think it isn't a big difference if we work with `*Vector` or `Vector`.

backend/witness/witness.go
```diff
@@ -199,50 +261,62 @@ func (w *Witness) UnmarshalJSON(data []byte) error {
...
 
-func (w *Witness) toAssignment(to interface{}, toLeafType reflect.Type) error {
-	if w.Schema == nil {
-		return errMissingSchema
-	}
-	if w.Vector == nil {
-		return fmt.Errorf("%w: empty witness", ErrInvalidWitness)
-	}
+	{
+		chValues := make(chan any)
```
ivokub, line 300: I don't follow. Why do we have to walk again with only public values now?

backend/witness/witness_test.go
```diff
@@ -114,26 +84,40 @@ func TestPublic(t *testing.T) {
...
+	rw, err := witness.New(ecc.BN254.ScalarField())
+	assert.NoError(err)
+	err = rw.UnmarshalBinary(data)
+	assert.NoError(err)
+
+	assert.True(reflect.DeepEqual(rw, w), "witness binary round trip serialization")
 
-func init() {
-	tVariable = reflect.TypeOf(circuit{}.E)
 }
```
ivokub, line 123: I would add test case for `FromJSON`

internal/generator/backend/template/zkpschemes/plonk/plonk.prove.go.tmpl
```diff
@@ -38,7 +37,7 @@ type Proof struct {
 }
 
 // Prove from the public data
-func Prove(spr *cs.SparseR1CS, pk *ProvingKey, fullWitness {{ toLower .CurveID }}witness.Witness, opt backend.ProverConfig) (*Proof, error) {
+func Prove(spr *cs.SparseR1CS, pk *ProvingKey, fullWitness []fr.Element, opt backend.ProverConfig) (*Proof, error) {
```
ivokub, line 40: fr.Vector instead of []fr.Element? Then is more consistent.

internal/generator/backend/template/zkpschemes/plonk/plonk.verify.go.tmpl
```diff
@@ -20,7 +19,7 @@ var (
 	errWrongClaimedQuotient = errors.New("claimed quotient is not as expected")
 )
 
-func Verify(proof *Proof, vk *VerifyingKey, publicWitness {{ toLower .CurveID }}witness.Witness) error {
+func Verify(proof *Proof, vk *VerifyingKey, publicWitness []fr.Element) error {
```
ivokub, line 22: Type consistency fr.Vector.

internal/generator/backend/template/zkpschemes/plonkfri/plonk.prove.go.tmpl
```diff
@@ -47,7 +46,7 @@ type Proof struct {
 	OpeningsId1Id2Id3mp [3]fri.OpeningProof
 }
 
-func Prove(spr *cs.SparseR1CS, pk *ProvingKey, fullWitness {{ toLower .CurveID }}witness.Witness, opt backend.ProverConfig) (*Proof, error) {
+func Prove(spr *cs.SparseR1CS, pk *ProvingKey, fullWitness []fr.Element, opt backend.ProverConfig) (*Proof, error) {
```
ivokub, line 49: Type consistency fr.Vector

internal/generator/backend/template/zkpschemes/plonkfri/plonk.verify.go.tmpl
```diff
@@ -5,14 +5,13 @@ import (
...
 	{{- template "import_fr" . }}
-    {{ template "import_witness" . }}
 	
     fiatshamir "github.com/consensys/gnark-crypto/fiat-shamir"
 )
 
 var ErrInvalidAlgebraicRelation = errors.New("algebraic relation does not hold")
 
-func Verify(proof *Proof, vk *VerifyingKey, publicWitness {{ toLower .CurveID }}witness.Witness) error {
+func Verify(proof *Proof, vk *VerifyingKey, publicWitness []fr.Element) error {
```
ivokub, line 14: Type consistency fr.Vector

internal/generator/backend/template/zkpschemes/groth16/groth16.prove.go.tmpl
```diff
@@ -35,7 +34,7 @@ func (proof *Proof) CurveID() ecc.ID {
 }
 
 // Prove generates the proof of knowledge of a r1cs with full witness (secret + public part).
-func Prove(r1cs *cs.R1CS, pk *ProvingKey, witness {{ toLower .CurveID }}witness.Witness, opt backend.ProverConfig) (*Proof, error) {
+func Prove(r1cs *cs.R1CS, pk *ProvingKey, witness []fr.Element, opt backend.ProverConfig) (*Proof, error) {
```
ivokub, line 37: Type consistency fr.Vector

internal/generator/backend/template/zkpschemes/groth16/groth16.verify.go.tmpl
```diff
@@ -19,7 +19,7 @@ var (
 )
 
 // Verify verifies a proof with given VerifyingKey and publicWitness
-func Verify(proof *Proof, vk *VerifyingKey, publicWitness {{ toLower .CurveID}}witness.Witness) error {
+func Verify(proof *Proof, vk *VerifyingKey, publicWitness []fr.Element) error {
```
ivokub, line 22: Type consistency fr.Vector

ivokub (approved):
Looks all good. A few minor comments but otherwise good to go. I think this change makes a lot of sense and makes the whole binary representation of witness so much more clear and practical to use.

I don't understand why the tests are failing though. May it due to double filling in FromJSON?

backend/witness/vector.go
gbotrel: yes may be a clear way to avoid it now; at some point in the refactoring it was just easier to do this for deserialization and growing the slice in-place.

backend/witness/witness.go
gbotrel: can be skipped now that this is separated; but need to identify before if we errored because of missing assignment in the public or secret witness, and discard any elements we read from the secret witness. Will fix.

backend/witness/witness_test.go
gbotrel: round trip JSON is tested in intergration_test (assert.go). Though it doesn't cost much to add something similar here too.

PR #478, opened 2023-02-18
Fix the example in `README.md`

aybehrouz:
Unfortunately, the example in `README.md` still does not compile. Apparently `BN254` has a `ScalarField()` method now, and should not be passed as a parameter directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,14 +99,14 @@ func (circuit *CubicCircuit) Define(api frontend.API) error {
 func main() {
 	// compiles our circuit into a R1CS
 	var circuit CubicCircuit
-	ccs, _ := frontend.Compile(ecc.BN254, r1cs.NewBuilder, &circuit)
+	ccs, _ := frontend.Compile(ecc.BN254.ScalarField(), r1cs.NewBuilder, &circuit)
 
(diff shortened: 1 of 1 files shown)


ivokub (approved):

ivokub:
@gbotrel - It is README only change. I think can be merged into master but pinging to be sure that it won't invoke any automatic releases etc.

gbotrel:
Yep it's fine 👍  thanks for the PR. We should probably find a way to have these examples in a `_test.go` file to ensure we update them, not the first time...

ivokub:
> Yep it's fine 👍 thanks for the PR. We should probably find a way to have these examples in a `_test.go` file to ensure we update them, not the first time...

Yep. Maybe we could use package documentation for the actual README, where the examples are documentation examples (`Example...` function in test files) and then use https://pkg.go.dev/github.com/posener/goreadme to compile into README.md.

Only ugly part with it is that then Go package documentation site has duplicate content (pkg.go.dev/...)

aybehrouz:
Also, this Github Action from the marketplace could be useful, though I don't know if using it is safe or not:

https://github.com/marketplace/actions/markdown-embed-code-from-file

PR #480, opened 2023-02-20
feat: add simple key-value store to the builders

ivokub:
With new gadgets, we may have some state we want to store over different calls. Instead of using singletons (essentially global variables), we instead add a `map[any]any` to the builders which allow to store builder-local variables.

To prevent overuse of the map, we dont publish the methods publicly in the `frontend.API`, but instead embed an instance in the builders themselves. If a gadget wants to use the storage, then they first have to type-assert the frontend.API instance to `kvstore.Store` and then on the new variable call `SetKeyValue` and `GetKeyValue`.

Another great aspect of having `any` keys is that we can use any comparable types (comparable because map implementation requires so). This allows us to avoid any collisions by enforcing that the keys are of different type. See the example below:
```go

type gadgetKey struct{}

func (c *Circuit) Define(api frontend.API) error {
    kv, ok := api.(kvstore.Store)
    if !ok {
        return fmt.Errorf("api doesn't implement Store")
    }
    kv.SetKeyValue(gadgetKey{}, "aaa") // only this package can set this key as they do not have the type
    // ...
    res := kv.GetKeyValue(gadgetKey{}) //
    // use res = "aaa"
}
```

In the previous example, we could also have for example `type gadgetKey int` and then have different values for the keys if the gadget needs to store different values.

Initially I thought about just using `context.Context` as it provides similar features, but it doesn't seem to fit in our use cases. See comments in #472.

### DIFF
--- a/frontend/cs/r1cs/builder.go
+++ b/frontend/cs/r1cs/builder.go
@@ -29,6 +29,7 @@ import (
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/frontend/internal/expr"
 	"github.com/consensys/gnark/frontend/schema"
+	"github.com/consensys/gnark/internal/kvstore"
 	"github.com/consensys/gnark/internal/tinyfield"
 	"github.com/consensys/gnark/internal/utils"
 	"github.com/consensys/gnark/logger"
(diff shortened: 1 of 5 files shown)


Tabaie (approved):
Very nice! I for one find this very useful for GKR and Commitments

PR #481, opened 2023-02-20
refactor: define Committer interface for builders

ivokub:
Right now `Commit` is part of the `frontend.Compiler` API. But this doesn't work well in case where the builders actually do not implement commitment (e.g. PLONK). We will get a runtime panic but this is really difficult to detect in a way which is sustainable long term (string handling etc.).

So instead, I add a new interface `Committer` with a single method `Commit` (as previously in `frontend.Compiler`). If the builder implements commitments, then any user can type assert and use the method and otherwise rely on some locally-defined fallback.

I also added a simple and hopefully fast enough commitment to the test engine. And removed the commitment from PLONK for now until we implement augmented-PLONK

### DIFF
--- a/frontend/builder.go
+++ b/frontend/builder.go
@@ -48,13 +48,6 @@ type Compiler interface {
 
 	// FieldBitLen returns the number of bits needed to represent an element in the scalar field
 	FieldBitLen() int
-
-	// Commit returns a commitment to the given variables, to be used as initial randomness in
-	// Fiat-Shamir when the statement to prove is particularly large.
-	// TODO cite paper
-	// ! Experimental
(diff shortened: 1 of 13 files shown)


ivokub:
> I'm currently working on commitments for Plonk/scs. If R1CS and test.engine already have it, doesn't that obviate the extra interface?

I think it doesn't. I believe there still may be cases where the builder may not have the capability:
- wizard? I'm not sure.
- whatever we may integrate in the future.

However, imo the biggest benefit this additional interface provides is that the "augmenting" (LegoSNARK definition) is completely optional and we still have the option of doing vanilla-PLONK/Groth16. This requirement may rise in the context of external verifiers which cannot check commitments or preserving compatibility with other prover backends.

Tabaie (approved):

PR #482, opened 2023-02-21
fix: allocate new variable in engine.MulAcc

ivokub:
We previously modified the result inline, but the behaviour may be inconsistent with actual frontend/compiler. When variable is a Term or LinearExpression, then
```
   b := a
```
copies value of a into b and after that these variables are independent. For example, with MulAcc(b, x, y) we won't modify the variable a. But, as test engine modified variable inline, then doing MulAcc(b, x, y) also changed value of a (as they are just *big.Int). This lead to inconsistent results.

### DIFF
--- a/test/engine.go
+++ b/test/engine.go
@@ -161,8 +161,9 @@ func (e *engine) MulAcc(a, b, c frontend.Variable) frontend.Variable {
 	bc := pool.BigInt.Get()
 	bc.Mul(e.toBigInt(b), e.toBigInt(c))
 
+	res := new(big.Int)
 	_a := e.toBigInt(a)
-	_a.Add(_a, bc).Mod(_a, e.modulus())
+	res.Add(_a, bc).Mod(res, e.modulus())
 
 	pool.BigInt.Put(bc)
 	return _a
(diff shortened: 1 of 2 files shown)


gbotrel:
well, if I understood correctly, the real behavior difference is not from the `MulAcc` but from the ... `a := b` right?

ivokub:
> well, if I understood correctly, the real behavior difference is not from the `MulAcc` but from the ... `a := b` right?

Yup. If in-circuit there is `a := b`, then the behaviour for test engine and prover is different (copy vs pointer)

gbotrel (approved):
seems the failing test is unrelated to the PR right? lgtm 👍

ivokub:
> seems the failing test is unrelated to the PR right? lgtm 👍 

I'll have a look. May actually be related as other PRs are going through checks just fine. Or it may be that issue https://github.com/ConsenSys/gnark/issues/484 is amplified with the changes.

test/engine.go
```diff
@@ -161,8 +161,9 @@ func (e *engine) MulAcc(a, b, c frontend.Variable) frontend.Variable {
 	bc := pool.BigInt.Get()
 	bc.Mul(e.toBigInt(b), e.toBigInt(c))
 
+	res := new(big.Int)
 	_a := e.toBigInt(a)
-	_a.Add(_a, bc).Mod(_a, e.modulus())
+	res.Add(_a, bc).Mod(res, e.modulus())
 
 	pool.BigInt.Put(bc)
 	return _a
```
aybehrouz, line 169: I think you should `return res` here.

ivokub: Thanks, good catch! I believe this was it.

PR #483, opened 2023-02-21
feat: add defer to the Compiler interface

ivokub:
See discussion in #472. This PR adds `PreCompileHook(cb func(API) error)` method to the `frontend.Compiler`. The intent of the method is to register all callback which should be called after the `Define` method of the circuit is called. These callbacks can be used by gadgets to safely defer any necessary closing tasks (e.g. range checking builds actual constraints, permutation networks build the permutations etc.) without requiring any interaction from the user.

I don't like the implementation though, I think it is really fragile. However, the problem is that I do not want to make `frontend.Compile` interface any more complex than it barely should, and this also means that there isn't a method to get all the callbacks which the `frontend.Compile` method can use. That is why I have embedded calling the callback in every builder's `Compile` method. But this is really really ugly solution.
A better way would definitely be to make list of all callback available to `frontend.Compile` without making `frontend.Compiler` interface bigger. One option, for example, would be to use the key-value store with some internal key. For example the implementation for registering callback:
```go

// type CallbackKey is in frontend/internal/XXX

func (b *builder) PreCompileHook(cb func(frontend.API) error) {
    cbs := b.GetKeyValueStore(CallbackKey{})
    b.SetKeyValueStore(callbackKey{}, append(cbs, cb))
}
```

and then in `frontend.Compile`:
```go
function Compile() {
// ...
    cbs := builder.GetKeyValue(CallbackKey{})
    // call all callbacks
    builder.Compile()
}
```

### DIFF
--- a/debug/debug.go
+++ b/debug/debug.go
@@ -75,5 +75,8 @@ func writeStack(sbb *strings.Builder, forceClean ...bool) {
 		if strings.HasSuffix(function, "Define") {
 			break
 		}
+		if strings.HasSuffix(function, "callDeferred") {
+			break
+		}
 	}
 }
--- a/debug/symbol_table.go
+++ b/debug/symbol_table.go
@@ -76,6 +76,9 @@ func (st *SymbolTable) CollectStack() []int {
 		if strings.HasSuffix(function, "Define") {
 			break
 		}
+		if strings.HasSuffix(function, "callDeferred") {
+			break
+		}
 	}
 	return r
 }
--- a/frontend/builder.go
+++ b/frontend/builder.go
@@ -55,6 +55,11 @@ type Compiler interface {
 	// ! Experimental
 	// TENTATIVE: Functions regarding fiat-shamir-ed proofs over enormous statements  TODO finalize
 	Commit(...Variable) (Variable, error)
+
+	// Defer is called after circuit.Define() and before Compile(). This method
+	// allows for the circuits to register callbacks which finalize batching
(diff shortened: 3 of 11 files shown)


gbotrel:
follow up on our discussion; my vote is to call it `api.Compiler().Defer(...)` . I think it's OK to assume users calling a method on a `compiler` object should read the doc (ie that `Defer` on the compiler object is not limited to the scope of a function like `go` `defer` ) .

ivokub:
> follow up on our discussion; my vote is to call it `api.Compiler().Defer(...)` . I think it's OK to assume users calling a method on a `compiler` object should read the doc (ie that `Defer` on the compiler object is not limited to the scope of a function like `go` `defer` ) .

Makes sense. I also think it seems nicer.

I'll also try to append all the deferred methods into the key-value storage to allow `frontend.Compile` to call the callbacks directly instead of doing in builder.

gbotrel (changes requested):
lgtm, w/ `PreCompileHook(..)` --> `Defer(...)`

ivokub:
> lgtm, w/ `PreCompileHook(..)` --> `Defer(...)`

Renamed, but it still doesn't feel right. I'll try with the kvstore to see if is better.

ivokub:
It is done but rebased on top of #480 (kvstore PR, not yet merged). Imo looks a bit nicer and I think a bit more future-proof.

gbotrel (approved):

PR #486, opened 2023-02-22
feat: add a partition selector

aybehrouz:
This is a simple partition selector. It selects the first count number of the input array, keeps them and zeros all other entries or vice versa. By cascading these gadgets a slice can be selected. It works using a step function as a mask.

I just had this idea and tried to implement it. I'm not sure if my implementation is good or if this is useful. Currently some tests are failing for a specific curve. I have no idea why, and I don't know how I can fix that!

### DIFF
--- a/std/selector/slice.go
+++ b/std/selector/slice.go
@@ -1,0 +1,79 @@
+package selector
+
+import (
+	"github.com/consensys/gnark/backend/hint"
+	"github.com/consensys/gnark/frontend"
+	"math/big"
+)
+
+func init() {
+	// register hints
+	hint.Register(StepOutput)
+}
+
+// Partition selects left or right side of the input array, with respect to the pivotPosition.
+// More precisely when rightSide is false, for each i:
+//
+//	if i < pivotPosition
+//	    out[i] = input[i]
+//	else
+//	    out[i] = 0
+//
+// and when rightSide is true, for each i we have:
+//
+//	if i >= pivotPosition
+//	    out[i] = input[i]
+//	else
+//	    out[i] = 0
+//
+// We must have pivotPosition >= 1 and pivotPosition <= len(input)-1, otherwise no proof can be generated.
+func Partition(api frontend.API, pivotPosition frontend.Variable, rightSide bool,
+	input []frontend.Variable) []frontend.Variable {
+	out := make([]frontend.Variable, len(input))
+	var mask []frontend.Variable
+	if rightSide {
+		mask = StepMask(api, len(input), pivotPosition, 0, 1)
+	} else {
+		mask = StepMask(api, len(input), pivotPosition, 1, 0)
+	}
+	for i := 0; i < len(out); i++ {
+		out[i] = api.Mul(mask[i], input[i])
+	}
+	return out
+}
+
+func StepMask(api frontend.API, outputLen int,
+	stepPosition, startValue, endValue frontend.Variable) []frontend.Variable {
+	if outputLen < 2 {
+		panic("output len must be >= 2")
+	}
+	// Get the output as a hint
+	out, _ := api.Compiler().NewHint(StepOutput, outputLen, stepPosition, startValue, endValue)
+
+	// Add boundary constraints
+	api.AssertIsEqual(out[0], startValue)
+	api.AssertIsEqual(out[len(out)-1], endValue)
+
+	// Add constraints for the correct form of a step function that steps at the stepPosition
+	for i := 1; i < len(out); i++ {
+		// (out[i] - out[i-1]) * (i - stepPosition) == 0
+		api.AssertIsEqual(api.Mul(api.Sub(out[i], out[i-1]), api.Sub(i, stepPosition)), 0)
+	}
+	return out
+}
+
+// StepOutput is a hint function used within [StepMask] function. It must be
+// provided to the prover when circuit uses it.
+func StepOutput(_ *big.Int, inputs, results []*big.Int) error {
+	stepPos := inputs[0]
+	startValue := inputs[1]
+	endValue := inputs[2]
+	for i := 0; i < len(results); i++ {
+		if i < int(stepPos.Int64()) {
+			results[i].Set(startValue)
+		} else {
+			results[i].Set(endValue)
+		}
+	}
+	return nil
+}
--- a/std/selector/slice.go
+++ b/std/selector/slice.go
@@ -26,15 +26,16 @@ func init() {
 //	else
 //	    out[i] = 0
 //
-// We must have pivotPosition >= 1 and pivotPosition <= len(input)-1, otherwise no proof can be generated.
+// We must have pivotPosition >= 1 and pivotPosition <= len(input), otherwise a proof cannot be generated.
 func Partition(api frontend.API, pivotPosition frontend.Variable, rightSide bool,
 	input []frontend.Variable) []frontend.Variable {
 	out := make([]frontend.Variable, len(input))
 	var mask []frontend.Variable
+	// we create a mask of len(input)+1 to be able to handle the case pivotPosition == len(input)
 	if rightSide {
-		mask = StepMask(api, len(input), pivotPosition, 0, 1)
+		mask = StepMask(api, len(input)+1, pivotPosition, 0, 1)
 	} else {
-		mask = StepMask(api, len(input), pivotPosition, 1, 0)
+		mask = StepMask(api, len(input)+1, pivotPosition, 1, 0)
 	}
 	for i := 0; i < len(out); i++ {
 		out[i] = api.Mul(mask[i], input[i])
--- a/std/selector/slice.go
+++ b/std/selector/slice.go
@@ -26,16 +26,16 @@ func init() {
 //	else
 //	    out[i] = 0
 //
-// We must have pivotPosition >= 1 and pivotPosition <= len(input), otherwise a proof cannot be generated.
+// We must have pivotPosition >= 1 and pivotPosition <= len(input)-1, otherwise a proof cannot be generated.
 func Partition(api frontend.API, pivotPosition frontend.Variable, rightSide bool,
 	input []frontend.Variable) (out []frontend.Variable) {
 	out = make([]frontend.Variable, len(input))
 	var mask []frontend.Variable
 	// we create a mask of len(input)+1 to be able to handle the case pivotPosition == len(input)
 	if rightSide {
-		mask = StepMask(api, len(input)+1, pivotPosition, 0, 1)
+		mask = StepMask(api, len(input), pivotPosition, 0, 1)
 	} else {
-		mask = StepMask(api, len(input)+1, pivotPosition, 1, 0)
+		mask = StepMask(api, len(input), pivotPosition, 1, 0)
 	}
 	for i := 0; i < len(out); i++ {
 		out[i] = api.Mul(mask[i], input[i])
--- a/std/selector/slice.go
+++ b/std/selector/slice.go
@@ -12,7 +12,7 @@ func init() {
 }
 
 // Partition selects left or right side of the input array, with respect to the pivotPosition.
-// More precisely when rightSide is false, for each i:
+// More precisely when rightSide is false, for each i we have:
 //
 //	if i < pivotPosition
 //	    out[i] = input[i]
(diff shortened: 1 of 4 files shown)


aybehrouz:
tests that fail:
```
=== RUN   TestPartition/bls24_317/plonkFRI#03
=== PAUSE TestPartition/bls24_317/plonkFRI#03
=== CONT  TestPartition/bls24_317/plonkFRI#03
    assert.go:517: 
        	Error Trace:	/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:517
        	            				/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:227
        	            				/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:269
        	            				/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:74
        	Error:      	did not error (but should have) plonkFRI(bls24_317)
        	            	witness:{"Pivot":6,"In":[10,20,30,40,50,60],"Left":[10,20,30,40,50,60],"Right":[0,0,0,0,0,0]}
        	Test:       	TestPartition/bls24_317/plonkFRI#03
    --- FAIL: TestPartition/bls24_317/plonkFRI#03 (0.63s)
```
```
=== RUN   TestPartition/bls24_317/plonkFRI#04
=== PAUSE TestPartition/bls24_317/plonkFRI#04
=== CONT  TestPartition/bls24_317/plonkFRI#04
    assert.go:517: 
        	Error Trace:	/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:517
        	            				/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:227
        	            				/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:269
        	            				/home/aybehrouz/GolandProjects/gnark/std/selector/assert.go:74
        	Error:      	did not error (but should have) plonkFRI(bls24_317)
        	            	witness:{"Pivot":0,"In":[10,20,30,40,50,60],"Left":[0,0,0,0,0,0],"Right":[10,20,30,40,50,60]}
        	Test:       	TestPartition/bls24_317/plonkFRI#04
    --- FAIL: TestPartition/bls24_317/plonkFRI#04 (1.02s)
    ```

**Comment by aybehrouz** [comment]
I'll close this for now. Later it could be reopened, if needed.

**Comment by ivokub** [comment]
Sorry for not reviewing the PR in time. We definitely appreciate the contribution. Skimming through the implementation it seems to be of high quality and I wouldn't mind merging. I was just postponing the review after other tasks :)

From my side, it can stay open because imo otherwise the PR will be buried into other closed (and merged) PRs.

**Comment by aybehrouz** [comment]
Not at all, actually I realized that you guys were busy and because it wasn't something high priority, I didn't want to keep the PR list crowded, so I closed it to reopen later.

File: std/selector/slice.go [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+		// (out[i] - out[i-1]) * (i - stepPosition) == 0
+		api.AssertIsEqual(api.Mul(api.Sub(out[i], out[i-1]), api.Sub(i, stepPosition)), 0)
+	}
+	return out
+}
+
+// StepOutput is a hint function used within [StepMask] function. It must be
+// provided to the prover when circuit uses it.
+func StepOutput(_ *big.Int, inputs, results []*big.Int) error {
+	stepPos := inputs[0]
```
ivokub, line 76: Just to be safe, check the input and output lengths. It may happen that someone uses this hint function incorrectly. Then it's better to have smoother errors.

std/selector/slice.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// We must have stepPosition >= 1 and stepPosition <= outputLen, otherwise a proof cannot be generated.
+// This function panics when outputLen is less than 2.
+func StepMask(api frontend.API, outputLen int,
+	stepPosition, startValue, endValue frontend.Variable) []frontend.Variable {
+	if outputLen < 2 {
+		panic("the output len of StepMask must be >= 2")
+	}
+	// get the output as a hint
+	out, _ := api.Compiler().NewHint(StepOutput, outputLen, stepPosition, startValue, endValue)
```
ivokub, line 59: Handle error. It should be sufficient to do `panic(fmt.Sprintf("step hint error: %v", err))`. I do not recommend adding error return to the function as it will be more difficult to use and safe recovery is impossible anyway (there is something wrong internally if err happens).

std/selector/slice.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// and when rightSide is true, for each i we have:
+//
+//	if i >= pivotPosition
+//	    out[i] = input[i]
+//	else
+//	    out[i] = 0
+//
+// We must have pivotPosition >= 1 and pivotPosition <= len(input)-1, otherwise a proof cannot be generated.
+func Partition(api frontend.API, pivotPosition frontend.Variable, rightSide bool,
```
ivokub, line 30: Looks good.

What do you think about having additionally `Slice` method a la
```go
func Slice(api frontend.API, start, end frontend.Variable, input []frontend.Variable) ([]frontend.Variable) {
    out := Partition(api, start, true, input)
    out = Partition(api, end, false, out)
    return out
}
```

Imo it is more natural to the way people use slices in Go as `out := in[start:end]`.

Additionally, maybe we can even have modified methods `PartitionCut` and `SliceCut` with additional argument `outPutlen` which returns an array of length `outPutlen` (int type), omitting the zeros. So for example:
```go
in := []frontend.Variable{1,2,3,4,5}
out := SliceCut(api, 1, 3, in, 2) // == []frontend.Variable{2,3}
out = SliceCut(api, 1, 3, in, 3) // == []frontend.Variable{2,3,0} as outputLen is larger than end-start
out = SliceCut(api, 1, 3, in, 1) // == []frontend.Variable{2}, result incomplete
out = SliceCut(api, 1, 3, in, 6) // panic as outLen > len(in)
```

std/selector/slice.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// StepMask generates a step like function into an output array of a given length.
+// The output is an array of length outputLen,
+// such that its first stepPosition elements are equal to startValue and the remaining elements are equal to
+// endValue. Note that outputLen cannot be a circuit variable.
+//
+// We must have stepPosition >= 1 and stepPosition <= outputLen, otherwise a proof cannot be generated.
+// This function panics when outputLen is less than 2.
+func StepMask(api frontend.API, outputLen int,
```
ivokub, line 53: I get the function and its use. However I am not sure if it makes sense to have it public. It took me a minute to understand what it does (from method doc). Is there any potential use outside of `Partition` function?

std/selector/slice.go
ivokub: But I guess if there is a `Slice` function then we can do it slightly more efficiently when we do it in a single pass instead of two in my proposal.

ivokub (changes requested):
Thanks for the contribution! I think that the gadget is useful and implementation seems very good.

I have some few inline comments though. I think error should be handled but I'm not sure if would be good to add new methods.

std/selector/slice.go
aybehrouz: Can we make these hint functions unexported or it may cause internal problems? They should not really appear on the public API. I wanted to make them unexported but I was afraid that may break something internally.

std/selector/slice.go
aybehrouz: Very good point.

std/selector/slice.go
ivokub: There has to be a way to make them available to the prover somehow. An approach I have been using elsewhere is to have a function which returns all used hints in a package a la
```go
func hintfn(_ *big.Int, []*big.Int, []*big.Int) error {
//
}

func GetHints() []solver.HintFunction {
    return []solver.HintFunction{hintfn}
}
```

It would also be good to register hints in [`std/hints.go`](https://github.com/ConsenSys/gnark/blob/develop/std/hints.go). Then the prover can get all hints by doing `std.RegisterHinds()`

std/selector/slice.go
aybehrouz: Actually, I think we should implement a `Slice` method. The thing is, after a while I realized that cascading is not a good idea and working with `mask` and using addition and subtraction can be more efficient. That's why I made `StepMask` public (exported). For example:
```
mask1         = [0, 0, 0, 1, 1, 1, 1, 1]
mask2         = [0, 0, 0, 0, 0, 0, 1, 1]
mask1 - mask2 = [0, 0, 0, 1, 1, 1, 0, 0]
```
So we will avoid an array multiplication.

I think implementing `SliceCut` won't be so easy, since `end` and `start` are wires. We would need to use `Mux` to *shift* every element.

std/selector/slice.go
aybehrouz: Actually, I think we should implement a `Slice` method. The thing is, after a while I realized that cascading is not a good idea and working with `mask` and using addition and subtraction can be more efficient. That's why I made `StepMask` public (exported). For example:
```
mask1         = [0, 0, 0, 1, 1, 1, 1, 1]
mask2         = [0, 0, 0, 0, 0, 0, 1, 1]
mask1 - mask2 = [0, 0, 0, 1, 1, 1, 0, 0]
```
So we will avoid an array multiplication.

I think implementing `SliceCut` won't be so easy, since `end` and `start` are wires. We would need to use `Mux` to *shift* every element.

std/selector/slice.go
aybehrouz: It's tricky, I'll try to explain in the next comment.

std/selector/slice.go
aybehrouz: Should I also implement `GetHints`? How come is it working now? 😅

std/selector/slice.go
aybehrouz: Now I'm not sure do we really need a `Slice` function? Cascading seems pretty straightforward.

std/selector/slice.go
ivokub: I agree. It seems tricky. Lets omit `..Cut` versions.

std/selector/slice.go
ivokub: See suggested edit.

std/selector/slice.go
ivokub: Actually, now looking at it, I think `Slice` function makes a bit more sense that `Partition`. I really do not like the `rightSide` argument in the argument of `Partition`. I think it is error-prone for the users, while `Slice` is really natural when working with slices (except that the returned array has the same length).

ivokub (commented):
Changes wrt. hints. See also the comment vs `Slice`/`Partition`.

std/selector/slice.go
aybehrouz: So maybe we should make `Partiion` unexported and replace it with a `Slice` function?

aybehrouz:
I agree with changes wrt. hints. Let's all packages have the same method for handling hints.

std/selector/slice.go
aybehrouz: ops, that doesn't work. It will make one sided selection inefficient. 
What about having two functions `RightPartition` and `LeftPartition`?

aybehrouz: There is some special limitations because we are working with a circuit.

ivokub: In that case, I think lets keep it as is.

aybehrouz: OK then, Can you please push your changes for hints?

aybehrouz: I'll also add a `Slice` function it could be removed later if it was not useful.

ivokub: > OK then, Can you please push your changes for hints?

Will do

ivokub (commented):
Hmm. I think it almost looks perfect. How many constraints we add when we also want to handle pivot at beginning of end of the slice? If it is two contstraints per partition then I would definitely add (just add boundary cases).

std/selector/slice.go
```diff
@@ -75,9 +74,11 @@ func stepMask(api frontend.API, outputLen int,
...
 	}
 
-	// add the boundary constraints
-	api.AssertIsEqual(out[0], startValue)
-	api.AssertIsEqual(out[len(out)-1], endValue)
+	// add the boundary constraints:
+	// (out[0] - startValue) * stepPosition == 0
+	api.AssertIsEqual(api.Mul(api.Sub(out[0], startValue), stepPosition), 0)
+	// (out[len(out)-1] - endValue) * (len(out) - stepPosition) == 0
+	api.AssertIsEqual(api.Mul(api.Sub(out[len(out)-1], endValue), api.Sub(len(out), stepPosition)), 0)
```
aybehrouz, line 81: Actually I think I found a cheap way for handling boundary conditions.

ivokub (approved):
Looks good! Merge ahead!

PR #488, opened 2023-02-23
feat: add EVM precompiles

### LINKED ISSUE #466: feat: zk-evm precompiles
Each of the precompile generated through gnark frontend should have:

* clean function signature that matches what's expected in the inner prover
* a constraint count with PlonK+bypassing the range checks, and PlonK+with api.Commit() enabled range checks.

- [ ] Pairing
- [ ] ECMul
- [ ] ECAdd
- [ ] Modexp
- [ ] ECRecover
- [ ] Sha256

### LINKED ISSUE #484: bug: fuzzer failures with non-native arithmetic
Fuzzer occasionally finds inputs to the field emulation gadget where the test engine succeeds but the prover fails. I have reduced the problem to the way we enforce width constraints for the witness elements. In automatic width enforcement we have:
```go
if !frontend.IsCanonical(a.Limbs[i]) {
	// this is not a variable. This may happen when some limbs are
	// constant and some variables. A strange case but lets try to cover
	// it anyway.
	continue
}
```
But `frontend.IsCanonical` only checks if the variable is `cs.Term` or `cs.LinearExpression` (or old PLONK versions of them). In test engine variables are `*big.Int` and we pass the automatic width enforcement.

So, what happens is that fuzzer creates some input which is not valid for field emulation (more than 64 bits) but the test engine doesn't fail. Then this invalid input is provided to the prover which fails and we get a fuzzing error.

Imo the options are:
* in test engine, wrap `*big.Int` into some struct with some additional metadata. Then we can also start differentiating between variables and constants (for now we have the test option to set all variables as consts or non-consts, which is far from good). Then add the new struct into the check in `frontend.IsCanonical` function.
* Make `IsCanonical` part of Compiler interface. Right now all places we use it has access to api, so it would technically work. But there may be cases where we want to check without API?

ivokub:
Closes #466.

Even if we have many primitives, then this PR is introducing tightly coupled circuits for using with EVM precompiles. The goal is that we can take verbatim precompile call arguments and return whatever implementations would return.

Still WIP, but starting a draft PR for discussions on interfaces etc. Need to fix #484 as with test engine right now get errors where shouldn't.

More specifically:
- [x] ECRecover - will have several variations with and without Keccak hashing the public key. Without hashing will rely on the prover to perform the hashing. But for consistency with the package promise will add variant with Keccak hashing. Mostly done, need to add a few in-circuit checks.
- [ ] sha256 - lower priority right now.
- [x] identity - pointless
- [ ] expmod - definitely needs rangechecking. I assume will be most complicated. Plan is to use Montgomery multiplication and textbook square-and-mul algorithm.
- [x] BN254 add - need to implement compatible interface.
- [x] BN254 scalar mul - need to implement compatible interface.
- [x] BN254 pairing - need to implement compatible interface.
- [ ] Blake2

Tasks:
- [x] right now for testing ECRecover I have a local version of gnark-crypto which outputs compatible signatures (with `v`). Need to polish the implementation and merge in gnark-crypto for better maintainability.

### DIFF
--- a/std/evmprecompiles/01-ecrecover.go
+++ b/std/evmprecompiles/01-ecrecover.go
@@ -33,7 +33,7 @@ func ECRecover(api frontend.API, msg emulated.Element[emulated.Secp256k1Fr],
 	if err != nil {
 		panic(fmt.Sprintf("new curve: %v", err))
 	}
-	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
+	// we cannot directly use the field emulation hint calling wrappers as we work between two fields.
 	Rlimbs, err := api.Compiler().NewHint(recoverPointHint, 2*int(emfp.NbLimbs()), recoverPointHintArgs(v, r)...)
 	if err != nil {
 		panic(fmt.Sprintf("point hint: %v", err))
@@ -42,7 +42,7 @@ func ECRecover(api frontend.API, msg emulated.Element[emulated.Secp256k1Fr],
 		X: *fpField.NewElement(Rlimbs[0:emfp.NbLimbs()]),
 		Y: *fpField.NewElement(Rlimbs[emfp.NbLimbs() : 2*emfp.NbLimbs()]),
 	}
-	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
+	// we cannot directly use the field emulation hint calling wrappers as we work between two fields.
 	Plimbs, err := api.Compiler().NewHint(recoverPublicKeyHint, 2*int(emfp.NbLimbs()), recoverPublicKeyHintArgs(msg, v, r, s)...)
 	if err != nil {
 		panic(fmt.Sprintf("point hint: %v", err))
--- a/std/evmprecompiles/01-ecrecover.go
+++ b/std/evmprecompiles/01-ecrecover.go
@@ -1,0 +1,79 @@
+package evmprecompiles
+
+import (
+	"fmt"
+
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/std/algebra/emulated/sw_emulated"
+	"github.com/consensys/gnark/std/math/bits"
+	"github.com/consensys/gnark/std/math/emulated"
+)
+
+// ECRecover implements [ECRECOVER] precompile contract at address 0x01.
+//
+// [ECRECOVER]: https://ethereum.github.io/execution-specs/autoapi/ethereum/paris/vm/precompiled_contracts/ecrecover/index.html
+func ECRecover(api frontend.API, msg emulated.Element[emulated.Secp256k1Fr],
+	v frontend.Variable, r, s emulated.Element[emulated.Secp256k1Fr]) *sw_emulated.AffinePoint[emulated.Secp256k1Fp] {
+	// EVM uses v \in {27, 28}, but everyone else v >= 0. Convert back
+	v = api.Sub(v, 27)
+	var emfp emulated.Secp256k1Fp
+	var emfr emulated.Secp256k1Fr
+	fpField, err := emulated.NewField[emulated.Secp256k1Fp](api)
+	if err != nil {
+		panic(fmt.Sprintf("new field: %v", err))
+	}
+	frField, err := emulated.NewField[emulated.Secp256k1Fr](api)
+	if err != nil {
+		panic(fmt.Sprintf("new field: %v", err))
+	}
+	// with the encoding we may have that r,s < 2*Fr (i.e. not r,s < Fr). Apply more thorough checks.
+	frField.AssertIsLessOrEqual(&r, frField.Modulus())
+	frField.AssertIsLessOrEqual(&s, frField.Modulus())
+	curve, err := sw_emulated.New[emulated.Secp256k1Fp, emulated.Secp256k1Fr](api, sw_emulated.GetSecp256k1Params())
+	if err != nil {
+		panic(fmt.Sprintf("new curve: %v", err))
+	}
+	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
+	Rlimbs, err := api.Compiler().NewHint(recoverPointHint, 2*int(emfp.NbLimbs()), recoverPointHintArgs(v, r)...)
+	if err != nil {
+		panic(fmt.Sprintf("point hint: %v", err))
+	}
+	R := sw_emulated.AffinePoint[emulated.Secp256k1Fp]{
+		X: *fpField.NewElement(Rlimbs[0:emfp.NbLimbs()]),
+		Y: *fpField.NewElement(Rlimbs[emfp.NbLimbs() : 2*emfp.NbLimbs()]),
+	}
+	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
+	Plimbs, err := api.Compiler().NewHint(recoverPublicKeyHint, 2*int(emfp.NbLimbs()), recoverPublicKeyHintArgs(msg, v, r, s)...)
+	if err != nil {
+		panic(fmt.Sprintf("point hint: %v", err))
+	}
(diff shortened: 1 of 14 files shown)


ivokub:
Precompiles done. Will update the stats in the corresponding issues.

Will do rest of the precompiles later (modexp, SHA256, Blake, RipeMD)

ivokub:
@yelhousni, ping

std/evmprecompiles/08-bnpairing.go
```diff
@@ -0,0 +1,21 @@
...
+
+// ECPair implements [ALT_BN128_PAIRING_CHECK] precompile contract at address 0x08.
+//
+// [ALT_BN128_PAIRING_CHECK]: https://ethereum.github.io/execution-specs/autoapi/ethereum/paris/vm/precompiled_contracts/alt_bn128/index.html#alt-bn128-pairing-check
+func ECPair(api frontend.API, P *sw_bn254.G1Affine, Q *sw_bn254.G2Affine) *sw_bn254.GTEl {
+	pair, err := sw_bn254.NewPairing(api)
+	if err != nil {
+		panic(err)
+	}
+	R, err := pair.Pair([]*sw_bn254.G1Affine{P}, []*sw_bn254.G2Affine{Q})
```
yelhousni, line 16: I think the precompile does not return the multi-pairing value `R` but rather checks if `e(P,Q)==1`. So maybe instead we should use 
https://github.com/ConsenSys/gnark/blob/6939cf5ba04cd000024b7120a69ea9cfcf9b2f6d/std/algebra/emulated/sw_bn254/pairing.go#L219

std/evmprecompiles/01-ecrecover.go
```diff
@@ -0,0 +1,79 @@
...
+	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
+	Rlimbs, err := api.Compiler().NewHint(recoverPointHint, 2*int(emfp.NbLimbs()), recoverPointHintArgs(v, r)...)
+	if err != nil {
+		panic(fmt.Sprintf("point hint: %v", err))
+	}
+	R := sw_emulated.AffinePoint[emulated.Secp256k1Fp]{
+		X: *fpField.NewElement(Rlimbs[0:emfp.NbLimbs()]),
+		Y: *fpField.NewElement(Rlimbs[emfp.NbLimbs() : 2*emfp.NbLimbs()]),
+	}
+	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
```
yelhousni, line 45: typo: *two* fields

std/evmprecompiles/01-ecrecover.go
```diff
@@ -0,0 +1,79 @@
...
+		panic(fmt.Sprintf("new field: %v", err))
+	}
+	// with the encoding we may have that r,s < 2*Fr (i.e. not r,s < Fr). Apply more thorough checks.
+	frField.AssertIsLessOrEqual(&r, frField.Modulus())
+	frField.AssertIsLessOrEqual(&s, frField.Modulus())
+	curve, err := sw_emulated.New[emulated.Secp256k1Fp, emulated.Secp256k1Fr](api, sw_emulated.GetSecp256k1Params())
+	if err != nil {
+		panic(fmt.Sprintf("new curve: %v", err))
+	}
+	// we cannot directly use the field emulation hint calling wrappers as we work between to fields.
```
yelhousni, line 36: typo: *two* fields

std/evmprecompiles/01-ecrecover.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/std/algebra/emulated/sw_emulated"
+	"github.com/consensys/gnark/std/math/bits"
+	"github.com/consensys/gnark/std/math/emulated"
+)
+
+// ECRecover implements [ECRECOVER] precompile contract at address 0x01.
+//
+// [ECRECOVER]: https://ethereum.github.io/execution-specs/autoapi/ethereum/paris/vm/precompiled_contracts/ecrecover/index.html
+func ECRecover(api frontend.API, msg emulated.Element[emulated.Secp256k1Fr],
```
yelhousni, line 15: maybe add ECRecover function to `std/signatures/ecdsa` as in gnark-crypto and use function here? but tbh the only use of ECDSA I can see is Ethereum.

yelhousni (changes requested):
LGTM! apart from the `08-bnpairing.go` where I think the precompile does not return the pairing value but just checks it is equal to 1 (there is no support of Fp12 arithmetic in EVM).

std/evmprecompiles/08-bnpairing.go
ivokub: Good catch! Changed. Also changed the corresponding test.

std/evmprecompiles/01-ecrecover.go
ivokub: I tried and it almost works until I got to hints used in ECrecover. The hints cannot be generic and thus I have to pass all the field emulation parameters (modulus, nb limbs, limb width) into the hint. Then, I would have to switch between different gnark-crypto `ECrecover` methods (we have in `bn254` and `secp256k1` package) depending on the modulus. It got quite messy.

I would rather postpone and keep it as an issue for future.

ivokub:
Resolved the comments. Right now didn't refactor ECRecover into `std/signature/ecdsa` as it would have made the implementation a lot more messy.

If re-checking code, then I would particularly point to ECPair test. I tried to define a succeeding test.

yelhousni:
> If re-checking code, then I would particularly point to ECPair test. I tried to define a succeeding test.

Suggested edit for ECPair test to use random point each time:
```diff
diff --git a/std/evmprecompiles/bn_test.go b/std/evmprecompiles/bn_test.go
index ed1cfb24..74815562 100644
--- a/std/evmprecompiles/bn_test.go
+++ b/std/evmprecompiles/bn_test.go
@@ -133,9 +133,8 @@ func TestECMulCircuitFull(t *testing.T) {
 }
 
 type ecpairCircuit struct {
-	P        [2]sw_bn254.G1Affine
-	Q        [2]sw_bn254.G2Affine
-	Expected sw_bn254.GTEl
+	P [2]sw_bn254.G1Affine
+	Q [2]sw_bn254.G2Affine
 }
 
 func (c *ecpairCircuit) Define(api frontend.API) error {
@@ -148,10 +147,19 @@ func (c *ecpairCircuit) Define(api frontend.API) error {
 func TestECPairCircuitShort(t *testing.T) {
 	assert := test.NewAssert(t)
 	_, _, p1, q1 := bn254.Generators()
+
+	var u, v fr.Element
+	u.SetRandom()
+	v.SetRandom()
+
+	p1.ScalarMultiplication(&p1, u.BigInt(new(big.Int)))
+	q1.ScalarMultiplication(&q1, v.BigInt(new(big.Int)))
+
 	var p2 bn254.G1Affine
 	var q2 bn254.G2Affine
 	p2.Neg(&p1)
 	q2.Set(&q1)
+
 	err := test.IsSolved(&ecpairCircuit{}, &ecpairCircuit{
 		P: [2]sw_bn254.G1Affine{sw_bn254.NewG1Affine(p1), sw_bn254.NewG1Affine(p2)},
 		Q: [2]sw_bn254.G2Affine{sw_bn254.NewG2Affine(q1), sw_bn254.NewG2Affine(q2)},

```

yelhousni (approved):

yelhousni:
@ivokub I applied the edit in the last commit + removed `Expected sw_bn254.GTEl` which is not used anymore after `PairingCheck()` use.

yelhousni:
Should we merge this PR or wait for the remaining precompiles?

ivokub:
> Should we merge this PR or wait for the remaining precompiles?

Looks good, I'll merge. Rest of the precompiles takes a bit time and is less important, so can do in another PR.

PR #490, opened 2023-02-23
sha256: groth16 tests pass, plonk tests fail

jplaui:
Hi gnark team,

concerning the sha256 implementation of this pull request, all tests pass when using the groth16 backend. However, I have issues with tests when using Plonk.
It would be amazing if you could take a look at my sha256 implementation at line 149: where I must negate a bit and AND it with another bit. 
I was using the following implementation as a reference (https://github.com/akosba/jsnark/blob/master/JsnarkCircuitBuilder/src/examples/gadgets/hash/SHA256Gadget.java#L213)

I guess I must work with hints at this point. However, I cannot find sufficient docs on how I must apply that.

Any feedback is highly appreciated!


gbotrel:
hi!; 
* after merging your branch with `develop` / latest PlonK refactor (#493 ), I still see an issue where indeed test seems to be OK with big.Int and `groth16` and failing with `plonk`; didn't manage to isolate the culprit yet.
* why do you need it to negate + AND ? (`api.Neg` and `api.And` are not enough? ) ; in any cases, there is a short exposition here https://docs.gnark.consensys.net/HowTo/write/hints and you can also checkout couple of hint usage in the gnark code base (look for calls to `NewHint`).

jplaui:
Hi Gautam,
I will try to solve the issues with the plonk backend. I tried more direct API calls api.Neg etc as well but I get the same errors until now. Thanks for referencing where I can find hints docs and samples. I'll commit again once I get it to work.

gbotrel:
(just a quick thought by "negating a bit", you mean the "not" operation? in which case you can probably validate with something like: 
`api.Select(a, 0, 1)` 
)

ivokub:
`api.Neg` is not optimal as it doesn't mark output boolean. Then in the next boolean operation we are going to assert that variable is a bit. In Keccak gadget I used `api.Xor(input, 1)` for emulating `api.Not`.

jplaui:
I am currently using `api.Xor(input, 1)` too and it works. Thanks for `api.Select(a, 0, 1)`, that is something I have not tried. Will check it.

gbotrel:
merged latest develop branch on top of this, seems tests are ok 👍

ivokub:
@jplaui, @gbotrel - I suggest closing this PR as #689 already introduces SHA2. At first look it seems that this PR doesn't introduce any additional features? If so, then I can consider porting them over.

PR #492, opened 2023-02-23
Feat/export solution

ThomasPiellard:
Addition of `GetTrace(witness witness.Witness, opts ...backend.ProverOption) (Trace, error)` on the `ConstraintSystem` interface.
For Groth16 -> it returns `w, a, b, c `such the `Aw o Bw - Cw = 0` and` a=Aw, b=Bw, c=Cw`
For Plonk -> it returns L, R, O.

`Trace` is an interface implementing `WriteTo` and `ReadFrom`.

The solving part in `Prove` for both Groth16 and Plonk has been modified to call GetTrace directly.

The goal is to extract a vector of solutions from a constraint system so that the constraint system + the corresponding solution can used in another tool (like the wizard iop).

Breaking change (internal only)
-------------------------------

In Groth16 and Plonk, `Prove` now takes the witness not casted, and a list of prover options (instead of prover config).


gbotrel:
Initial thoughts:

- [ ] The fix on the size of the multiexp in Groth16 should be a separate PR
- [ ] in a future PR, backend.ProverOptions / ProverConfig should likely be refactored into `constraint.SolverOption` (tracking issue: )
- [ ] since now the `constraint/XXX` packages are exposed, I'm thinking, the `ConstraintSystem` interface should expose one function (`Solve(witness, opts) any, error`) instead of `Solve` `GetTrace` and `IsSolved` . 
- [ ] in that spirit, `TraceR1CS` should be `R1CSSolution` and `TraceSparseR1CS` `SparseR1CSSolution`. We may even want to have subpackages for `r1cs` and `scs` under typed curved `constraint` packages;.. future PR. 
- [ ] please add a benchmark impact of this PR on the provers; in particular I suspect `groth16` will be impacted since previous version allocated the `a,b,c` vector with the FFT domain len as capacity to avoid extra mem moves.

gbotrel (changes requested):
Initial thoughts:

- [x] The fix on the size of the multiexp in Groth16 should be a separate PR
- [x] in a future PR, backend.ProverOptions / ProverConfig should likely be refactored into `constraint.SolverOption` (tracking issue: )
- [ ] since now the `constraint/XXX` packages are exposed, I'm thinking, the `ConstraintSystem` interface should expose one function (`Solve(witness, opts) any, error`) instead of `Solve` `GetTrace` and `IsSolved` . 
- [ ] in that spirit, `TraceR1CS` should be `R1CSSolution` and `TraceSparseR1CS` `SparseR1CSSolution`. We may even want to have subpackages for `r1cs` and `scs` under typed curved `constraint` packages;.. future PR. 
- [ ] please add a benchmark impact of this PR on the provers; in particular I suspect `groth16` will be impacted since previous version allocated the `a,b,c` vector with the FFT domain len as capacity to avoid extra mem moves.

gbotrel (approved):

Issue #504, opened 2023-03-01
bug: BSB22 commitment in Groth16 is not overloaded when running only the solver


ivokub:
So the commitment hint uses proving key which the solver doesn't have access to. When we run the prover then the prover overloads the hint with a function which has access to the proving key and then the solver has access to it to generate the full witness.

However, when we only run solver (in fuzzer/assertions etc.), then we do not overload and because the dummy commitment panics, the solver returns error. But actually the circuit may be solvable.

@Tabaie, @gbotrel

PR #507, opened 2023-03-01
fix: allow unreplaced BSB22 commitment hint in solver

ivokub:
Related #504.

In order to enable running full test suite where we use commitments we need to have the commitment available for the solver. I don't know what is a good way to do this, but this patch allows to continue with #472.

First, I register the hint in `frontend/r1cs` init. This gives solver access to the hint when run in the same process as the compiler. The hint is private so it isn't possible to provide it manually.

We also check if either debug tag is set or we run as a Go test (i.e. the executable name ends in `.test`). Only so we have a dummy commitment (zero).

Otherwise, we return an error, as before.

### DIFF
--- a/frontend/cs/r1cs/api.go
+++ b/frontend/cs/r1cs/api.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"os"
 	"path/filepath"
 	"reflect"
 	"runtime"
@@ -27,9 +28,11 @@ import (
 
 	"github.com/consensys/gnark/constraint"
 	"github.com/consensys/gnark/constraint/solver"
+	"github.com/consensys/gnark/debug"
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/frontend/internal/expr"
 	"github.com/consensys/gnark/frontend/schema"
+	"github.com/consensys/gnark/logger"
 	"github.com/consensys/gnark/std/math/bits"
 )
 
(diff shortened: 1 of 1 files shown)


gbotrel:
so, right now we have a "CommitInfo" bits in the ... `constraint/System` --> so we know: 
```go
HintID                 solver.HintID
``` 
at solve time, in all constraint system. 

Shouldn't the R1CS.Solve() and SparseR1CS.Solve() methods check
```
if (commitment) and !alreadyReplaced(commitment.HintID) {
replace with something
}
```
? (commitmnet could already be repalced by the groth16 / plonk prover). Or did that change it latest design cc @Tabaie  ?

Tabaie:
> so, right now we have a "CommitInfo" bits in the ... `constraint/System` --> so we know: 
> ```go
> HintID                 solver.HintID
> ``` 
> at solve time, in all constraint system. 
> 
> Shouldn't the R1CS.Solve() and SparseR1CS.Solve() methods check
> ```
> if (commitment) and !alreadyReplaced(commitment.HintID) {
> replace with something
> }
> ```
> ? (commitmnet could already be repalced by the groth16 / plonk prover). Or did that change it latest design cc @Tabaie  ? 
> 

Yes I think that makes sense. Ideally the prover replaces the hint using proving key, but if that hasn't happened the solver would put a fake hint there. That at least is my understanding of what this PR does.
One other thing we could do is to actually use the latter version of the hint in the api.Commit function and register that hint by default so that we get the fallback case for free.

ivokub:
> > so, right now we have a "CommitInfo" bits in the ... `constraint/System` --> so we know:
> > ```go
> > HintID                 solver.HintID
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > at solve time, in all constraint system.
> > Shouldn't the R1CS.Solve() and SparseR1CS.Solve() methods check
> > ```
> > if (commitment) and !alreadyReplaced(commitment.HintID) {
> > replace with something
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ? (commitmnet could already be repalced by the groth16 / plonk prover). Or did that change it latest design cc @Tabaie  ?
> 
> Yes I think that makes sense. Ideally the prover replaces the hint using proving key, but if that hasn't happened the solver would put a fake hint there. That at least is my understanding of what this PR does. One other thing we could do is to actually use the latter version of the hint in the api.Commit function and register that hint by default so that we get the fallback case for free.

Actually this is exactly what this PR is doing. frontend package registers the hint, but we do not register it in the `constraint.RegisterHints` (so the hint is only available if the frontend package is imported).

gbotrel (approved):

PR #511, opened 2023-03-01
fix: closes #509 api did not handle AssertIsLessOrEqual with constant as first param

### LINKED ISSUE #509: `AssertIsLessOrEqual` panics when first argument is an int literal
One of my circuits attempts to do the following:
```go
api.AssertIsLessOrEqual(0, v)
```
But when I try to test this, it panics with the message:
```
interface conversion: frontend.Variable is uint, not expr.LinearExpression
```
Exact message depends on the way the circuit is compiled (e.g. r1cs or scs, groth16 or plonk, which curve, etc), e.g. may say:
```
interface conversion: frontend.Variable is int, not expr.TermToRefactor
```

---

For (R1CS, Groth16, BN254), I tracked this down to the `mustBeLessOrEqVar` function in `frontend/cs/r1cs/api_assertions.go`, on line 131, where we attempt to do a type assertion of `aBits[i].(expr.LinearExpression)`. When I was debugging this, it seemed that `aBits[i]` was a `frontend.Variable` wrapped around a `uint`, not a `expr.LinearExpression`.

Please let me know what other info I can provide to help with this, I'm not much of a Go guru so I'm not sure why this type assertion is failing, or what the fix should be

gbotrel:


akirillo:
dang you are fast dude

frontend/cs/r1cs/api_assertions.go
```diff
@@ -187,7 +209,7 @@ func (builder *builder) mustBeLessOrEqCst(a expr.LinearExpression, bound big.Int
 			l = builder.Sub(l, aBits[i])
 
 			added = append(added, builder.cs.AddConstraint(builder.newR1C(l, aBits[i], builder.cstZero())))
-			builder.MarkBoolean(aBits[i].(expr.LinearExpression))
+			builder.MarkBoolean(aBits[i])
```
ivokub, line 212: I'm trying to understand why do we mark `aBits[i]` boolean at all? Imo marking something boolean is useful when we are going to use it later in a boolean operation and then we do not have to assert. But in this method this doesn't seem to happen anyway?

frontend/cs/r1cs/api_assertions.go
```diff
@@ -128,16 +144,22 @@ func (builder *builder) mustBeLessOrEqVar(a, bound expr.LinearExpression) {
 		// note if bound[i] == 1, this constraint is (1 - ai) * ai == 0
 		// → this is a boolean constraint
 		// if bound[i] == 0, t must be 0 or 1, thus ai must be 0 or 1 too
-		builder.MarkBoolean(aBits[i].(expr.LinearExpression)) // this does not create a constraint
+		builder.MarkBoolean(aBits[i]) // this does not create a constraint
```
ivokub, line 147: Similarly to comment below, why do we mark `aBits[i]` boolean. It looks the output is not returned or used (except for the call to `builder.newR1C` below, but I think it doesn't check that input is boolean)

frontend/cs/scs/api_assertions.go
```diff
@@ -147,18 +147,29 @@ func (builder *builder) mustBeLessOrEqVar(a expr.Term, bound expr.Term) {
 		// note if bound[i] == 1, this constraint is (1 - ai) * ai == 0
 		// → this is a boolean constraint
 		// if bound[i] == 0, t must be 0 or 1, thus ai must be 0 or 1 too
-		builder.MarkBoolean(aBits[i].(expr.Term)) // this does not create a constraint
+		builder.MarkBoolean(aBits[i]) // this does not create a constraint
```
ivokub, line 150: A pointer to the same case of marking boolean here.

ivokub (approved):
Looks good. I added a few comments about marking bits of decomposed values as boolean. It definitely doesn't hurt, but if the bits are not used, then the table of boolean variables gets bigger.

frontend/cs/r1cs/api_assertions.go
gbotrel: I don't know why we do that, as you pointed out, it's not useful.
The only case where I could see this being useful, is for prover side optimizations. Since it "tags" the wire as being a boolean, we could in theory identify a lot of 0 and 1 in the vector of field elements at proving time to do specialized multi exps or ffts. 

Going to remove it for now, the day we add these kind of optim it's easy to restore.

ivokub (approved):
Changes all good 👍

PR #518, opened 2023-03-03
fix: restrict constants in field emulation to width

### LINKED ISSUE #484: bug: fuzzer failures with non-native arithmetic

ivokub:
Closes #484.

### DIFF
--- a/std/math/emulated/field.go
+++ b/std/math/emulated/field.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/internal/utils"
 	"github.com/consensys/gnark/logger"
 	"github.com/rs/zerolog"
 	"golang.org/x/exp/constraints"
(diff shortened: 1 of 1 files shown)


gbotrel (approved):
Non-blocking remarks;
* `utils.FromInterface` don't `mod-reduce` the constant. I think we're fine but in case I missed a case where we are not... let me know :) .

ivokub:
> Non-blocking remarks;
> 
> * `utils.FromInterface` don't `mod-reduce` the constant. I think we're fine but in case I missed a case where we are not... let me know :) .

Good point. But I think we are fine. If the user gets a panic if they are not supposed to, then they could reduce the constant by hand. Otherwise we would have to modify limbs of the method arguments and this may lead to unexpected behaviour.

PR #522, opened 2023-03-04
docs: correct `WithNbDigits` description

aybehrouz:
When `nbDigits` is set to a number that is less than the length of the binary representation of the input, a proof can not be generated. That's because the `api.AssertIsEqual(Σbi, v)` constraint can not be satisfied. The description of `WithNbDigits` needs to be fixed.

### DIFF
--- a/std/math/bits/conversion.go
+++ b/std/math/bits/conversion.go
@@ -56,9 +56,10 @@ type baseConversionConfig struct {
 // BaseConversionOption configures the behaviour of scalar decomposition.
 type BaseConversionOption func(opt *baseConversionConfig) error
 
-// WithNbDigits set the resulting number of digits to be used in the base conversion.
-// nbDigits must be > 0. If nbDigits is lower than the length of full
-// decomposition, then nbDigits least significant digits are returned. If the
+// WithNbDigits sets the resulting number of digits (nbDigits) to be used in the
+// base conversion. nbDigits must be > 0. If nbDigits is lower than the length of full
+// decomposition and WithUnconstrainedOutputs option is not used,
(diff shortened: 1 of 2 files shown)


std/math/bits/conversion.go
```diff
@@ -56,9 +56,10 @@ type baseConversionConfig struct {
 // BaseConversionOption configures the behaviour of scalar decomposition.
 type BaseConversionOption func(opt *baseConversionConfig) error
 
-// WithNbDigits set the resulting number of digits to be used in the base conversion.
-// nbDigits must be > 0. If nbDigits is lower than the length of full
-// decomposition, then nbDigits least significant digits are returned. If the
+// WithNbDigits sets the resulting number of digits (nbDigits) to be used in the
+// base conversion. nbDigits must be > 0. If nbDigits is lower than the length of full
+// decomposition and WithUnconstrainedOutputs option is not used,
+// then a proof can not be generated. If WithNbDigits
```
gbotrel, line 62: replace "proof can not be generated" by something more "constraint" oriented (ie this generates an unsatisfiable constraint)

gbotrel (changes requested):

std/math/bits/conversion.go
aybehrouz: looks good?
```
// WithNbDigits sets the resulting number of digits (nbDigits) to be used in the base conversion.
// nbDigits must be > 0. If nbDigits is lower than the length of full decomposition and
// WithUnconstrainedOutputs option is not used, then this function generates an unsatisfiable
// constraint. If WithNbDigits option is not set, then the full decomposition is returned.
```

PR #526, opened 2023-03-06
Refactor: std/algebra

yelhousni:
refactors algebra into `native` that uses (2-chains) and `emulated`


yelhousni:
todo:
- [ ] add tests for `fields_bn254`

gbotrel:
👍 
+ maybe:

- [ ] add a `doc.go` in `std/algebra` & `emulated` / `native` to brievely introduce differences between these pacakges

yelhousni:
> todo:
> 
> * [x]  add tests for `fields_bn254`

@ivokub @gbotrel Done in #532 but in a separate PR as I'm doing some refactoring/cleaning there.  We can merge this PR.

std/algebra/emulated/sw_bn254/pairing.go
```diff
@@ -0,0 +1,283 @@
...
+	ba, err := emulated.NewField[emulated.BN254Fp](api)
+	if err != nil {
+		return nil, fmt.Errorf("new base api: %w", err)
+	}
+	return &Pairing{
+		Ext12: fields_bn254.NewExt12(ba),
+	}, nil
+}
+
+func (pr Pairing) DoubleStep(p *g2Projective) (*g2Projective, *lineEvaluation) {
```
ivokub, line 25: Imo if not necessary to export, then I would make private. It would be easier for the users to find.

std/algebra/emulated/sw_bn254/pairing.go
```diff
@@ -0,0 +1,283 @@
...
+	px := &Q.X            // p.x.Set(&Q.X)
+	py := &Q.Y            // p.y.Set(&Q.Y)
+	return &g2Projective{ // return p
+		X: *px,
+		Y: *py,
+		Z: *pz,
+	}
+}
+
+func (pr Pairing) NegAffine(a *G2Affine) *G2Affine {
```
ivokub, line 86: Make private?

std/algebra/emulated/sw_bn254/pairing.go
```diff
@@ -0,0 +1,283 @@
...
+	px := &a.X              // p.X = a.X
+	py := pr.Ext2.Neg(&a.Y) // p.Y.Neg(&a.Y)
+	return &G2Affine{       // return p
+		X: *px,
+		Y: *py,
+	}
+}
+
+func (pr Pairing) AddStep(p *g2Projective, a *G2Affine) (*g2Projective, *lineEvaluation) {
```
ivokub, line 95: Make private?

std/algebra/emulated/sw_bn254/pairing_test.go
```diff
@@ -71,7 +72,7 @@ func (c *FinalExponentiationCircuit) Define(api frontend.API) error {
 		return fmt.Errorf("new pairing: %w", err)
 	}
 	res := pairing.FinalExponentiation(&c.InGt)
-	pairing.ext12.assertIsEqual(res, &c.Res)
+	pairing.Ext12.AssertIsEqual(res, &c.Res)
```
ivokub, line 75: I previously added `AssertIsEqual` to the `Pairing` type, so it should be sufficient to do:
```go
pairing.AssertIsEqual(res, &c.Res)
```

std/algebra/emulated/sw_bn254/pairing_test.go
```diff
@@ -42,7 +43,7 @@ func (c *MillerLoopCircuit) Define(api frontend.API) error {
 	if err != nil {
 		return fmt.Errorf("pair: %w", err)
 	}
-	pairing.ext12.assertIsEqual(res, &c.Res)
+	pairing.Ext12.AssertIsEqual(res, &c.Res)
```
ivokub, line 46:
```go
pairing.AssertIsEqual(res, &c.Res)
```

std/algebra/emulated/fields_bn254/e12_pairing.go
```diff
@@ -0,0 +1,136 @@
+package fields_bn254
+
+import (
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark/std/math/emulated"
+)
+
+type GTEl = E12
```
ivokub, line 8: I would move this type alias and `NewGTEl` to `sw_bn254` package. Then we have all related types (or aliases) in a single package and it is easier to comprehend.

std/algebra/emulated/fields_bn254/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package fields_bn254 implements the fields arithmetic of the Fp12 tower
+// used to compute the pairing over the BN254 curve.
+//
+//	𝔽p²[u] = 𝔽p/u²+1
+//	𝔽p⁶[v] = 𝔽p²/v³-9-u
+//	𝔽p¹²[w] = 𝔽p⁶/w²-v
+package fields_bn254
```
ivokub, line 7: What do you think about making this package `fields_bn254` internal? Under `std/algebra/emulated/internal/fields_bn254`. When we have type alias `GTEl = E12` in `sw_bn254`, then it should be sufficient?

My reasoning for making it internal is that then we have more freedom in reimplementing, optimising it without having to worry that there are any external users. I would still keep `sw_emulated` and `sw_bn254` non-internal.

std/algebra/native/fields_bls12377/doc.go
```diff
@@ -0,0 +1,9 @@
+// Package fields_bls12377 implements the fields arithmetic of the Fp12 tower
+// used to compute the pairing over the BLS12-377 curve.
+//
+//	𝔽p²[u] = 𝔽p/u²+5
+//	𝔽p⁶[v] = 𝔽p²/v³-u
+//	𝔽p¹²[w] = 𝔽p⁶/w²-v
+//
+// Reference: https://eprint.iacr.org/2022/1162
+package fields_bls12377
```
ivokub, line 9: Similarly for emulated `fields_bn254` package, maybe make internal?

ivokub (changes requested):
A few comments. Otherwise looks good.

std/algebra/emulated/sw_bn254/pairing_test.go
```diff
@@ -28,9 +29,9 @@ func randomG1G2Affines(assert *test.Assert) (bn254.G1Affine, bn254.G2Affine) {
 }
 
 type MillerLoopCircuit struct {
-	InG1 weierstrass.AffinePoint[emulated.BN254Fp]
+	InG1 sw_emulated.AffinePoint[emulated.BN254Fp]
```
ivokub, line 32: Can use G1Affine (which is alias to `sw_emulated.AffinePoint`)

std/algebra/emulated/sw_bn254/doc_test.go
```diff
@@ -9,21 +9,22 @@ import (
...
+	"github.com/consensys/gnark/std/algebra/emulated/sw_bn254"
 )
 
 type PairCircuit struct {
-	InG1 pairing_bn254.G1Affine
-	InG2 pairing_bn254.G2Affine
-	Res  pairing_bn254.GTEl
+	InG1 sw_bn254.G1Affine
+	InG2 sw_bn254.G2Affine
+	Res  fields_bn254.GTEl
```
ivokub, line 19: After alias can use `sw_bn254.GTEl`

std/algebra/emulated/sw_bn254/pairing.go
yelhousni: This function will change anyway in the upcoming perf PR. Let's keep this refactoring PR minimal and will make the suggested change then.

std/algebra/emulated/sw_bn254/pairing.go
yelhousni: same as DoubleStep

std/algebra/emulated/sw_bn254/pairing.go
yelhousni: Same as DoubleStep and NegAffine

std/algebra/emulated/fields_bn254/e12_pairing.go
yelhousni: I would rather keep field aliases separate from curve aliases. `fields_bn254` package could be used independently of `sw_bn254` package.

std/algebra/emulated/fields_bn254/doc.go
yelhousni: that can make sense for now but I suspect we would have later use-cases of `fields_bn254` as an independent package (e.g. Protocols that use GT-arithmetic like SnarkPack or SSLE would need to export e12.go and e6.go for compression).

std/algebra/native/fields_bls12377/doc.go
yelhousni: same

std/algebra/emulated/fields_bn254/e12_pairing.go
ivokub: Definitely, we can use it separately. What I was proposing is to keep `E12` type in this package as is. All the methods on `E12` will stay here as is. We will only move the alias `type GTEl = E12` into `sw_bn254` package (as `type GTEl = fields_bn254.E12`. Actually it would be similar how it currently is for the 2-chains (see https://github.com/ConsenSys/gnark/blob/f2b7910c6a50c6272806bf21576e61efb2667167/std/algebra/native/sw_bls12377/pairing.go#L28)

yelhousni: I see! ok that makes sense

ivokub (approved):
LGTM

PR #528, opened 2023-03-06
Test/hints in tests

Tabaie:
The following simple test
```go
func idHint(_ *big.Int, in []*big.Int, out []*big.Int) error {
	if len(in) != len(out) {
		return fmt.Errorf("in/out length mismatch %d≠%d", len(in), len(out))
	}
	for i := range in {
		out[i].Set(in[i])
	}
	return nil
}

type idHintCircuit struct {
	X frontend.Variable
}

func (c *idHintCircuit) Define(api frontend.API) error {
	x, err := api.Compiler().NewHint(idHint, 1, c.X)
	if err != nil {
		return err
	}
	api.AssertIsEqual(x[0], c.X)
	return nil
}

func TestIdHint(t *testing.T) {
	solver.RegisterHint(idHint)
	assignment := idHintCircuit{0}
	test.NewAssert(t).SolvingSucceeded(&idHintCircuit{}, &assignment)
}
```
should pass, but currently gets a nondeterministic circuit error. This PR fixes that by ignoring the unexported fields of `constraint.System`.

### DIFF
--- a/constraint/bls12-377/r1cs.go
+++ b/constraint/bls12-377/r1cs.go
@@ -465,3 +465,11 @@ func (cs *R1CS) ReadFrom(r io.Reader) (int64, error) {
 
 	return int64(decoder.NumBytesRead()), nil
 }
+
+func (cs *R1CS) Equal(o constraint.ConstraintSystem) bool {
+	O, ok := o.(*R1CS)
+	if !ok {
+		return false
+	}
+	return cmp.Equal(*cs, *O, cmp.AllowUnexported(debug.SymbolTable{}, CoeffTable{}, R1CS{}), cmpopts.IgnoreUnexported(constraint.System{}))
+}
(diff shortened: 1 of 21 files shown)


gbotrel:
I'm ok with the PR on principle, just as discussed a bit uncomfortable with the direct dependence of `go-cmp` (ie in non test packages). Now a package that imports `constraint/xxx` will pull `go-cmp`. 
That being said, not super familiar with `go-cmp`, @ivokub  any thoughts ?

ivokub:
> I'm ok with the PR on principle, just as discussed a bit uncomfortable with the direct dependence of `go-cmp` (ie in non test packages). Now a package that imports `constraint/xxx` will pull `go-cmp`. 
> That being said, not super familiar with `go-cmp`, @ivokub  any thoughts ?

I'll have to take a look later. My non-deterministic compiles previously were caused by actual bugs. Maybe its possible to avoid it here without go-cmp. I agree it is a huge package and would avoid at if possible.

I'd propose to keep open for a bit.

Tabaie:
Closing this in favor of another simpler solution at https://github.com/ConsenSys/gnark/pull/529

PR #529, opened 2023-03-06
Test basic hint

Tabaie:
Alternative to https://github.com/ConsenSys/gnark/pull/528
Copy-pasting the problem statement

The following simple test

```go
func idHint(_ *big.Int, in []*big.Int, out []*big.Int) error {
	if len(in) != len(out) {
		return fmt.Errorf("in/out length mismatch %d≠%d", len(in), len(out))
	}
	for i := range in {
		out[i].Set(in[i])
	}
	return nil
}

type idHintCircuit struct {
	X frontend.Variable
}

func (c *idHintCircuit) Define(api frontend.API) error {
	x, err := api.Compiler().NewHint(idHint, 1, c.X)
	if err != nil {
		return err
	}
	api.AssertIsEqual(x[0], c.X)
	return nil
}

func TestIdHint(t *testing.T) {
	solver.RegisterHint(idHint)
	assignment := idHintCircuit{0}
	test.NewAssert(t).SolvingSucceeded(&idHintCircuit{}, &assignment)
}
```
should pass, but currently gets a nondeterministic circuit error. This PR fixes that by ignoring the unexported fields of constraint.System.

### DIFF
--- a/constraint/bls12-377/r1cs_test.go
+++ b/constraint/bls12-377/r1cs_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/frontend/cs/r1cs"
 	"github.com/consensys/gnark/internal/backend/circuits"
+	"github.com/consensys/gnark/test"
 	"reflect"
 	"testing"
 
@@ -74,15 +75,7 @@ func TestSerialization(t *testing.T) {
 
 				// compare original and reconstructed
 				if diff := cmp.Diff(r1cs1, &reconstructed,
-					cmpopts.IgnoreFields(cs.R1CS{},
-						"System.q",
-						"arithEngine",
-						"CoeffTable.mCoeffs",
-						"System.lbWireLevel",
-						"System.lbHints",
-						"System.SymbolTable",
-						"System.lbOutputs",
-						"System.bitLen")); diff != "" {
+					cmpopts.IgnoreFields(cs.R1CS{}, test.IgnoreFieldsCsCmp()...)); diff != "" {
 					t.Fatalf("round trip mismatch (-want +got):\n%s", diff)
 				}
 			}
--- a/constraint/bls12-381/r1cs_test.go
+++ b/constraint/bls12-381/r1cs_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/frontend/cs/r1cs"
 	"github.com/consensys/gnark/internal/backend/circuits"
+	"github.com/consensys/gnark/test"
 	"reflect"
 	"testing"
 
@@ -74,15 +75,7 @@ func TestSerialization(t *testing.T) {
 
 				// compare original and reconstructed
 				if diff := cmp.Diff(r1cs1, &reconstructed,
-					cmpopts.IgnoreFields(cs.R1CS{},
-						"System.q",
-						"arithEngine",
-						"CoeffTable.mCoeffs",
-						"System.lbWireLevel",
-						"System.lbHints",
-						"System.SymbolTable",
-						"System.lbOutputs",
-						"System.bitLen")); diff != "" {
+					cmpopts.IgnoreFields(cs.R1CS{}, test.IgnoreFieldsCsCmp()...)); diff != "" {
 					t.Fatalf("round trip mismatch (-want +got):\n%s", diff)
 				}
 			}
--- a/constraint/bls24-315/r1cs_test.go
+++ b/constraint/bls24-315/r1cs_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/frontend/cs/r1cs"
 	"github.com/consensys/gnark/internal/backend/circuits"
+	"github.com/consensys/gnark/test"
 	"reflect"
 	"testing"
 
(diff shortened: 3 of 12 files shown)


ivokub:
There is another approach, which fixes the root cause of the non-determinism:
```diff
diff --git a/constraint/level_builder.go b/constraint/level_builder.go
index c2b1d8c1..cfdb3496 100644
--- a/constraint/level_builder.go
+++ b/constraint/level_builder.go
@@ -10,12 +10,12 @@ package constraint
 // --> l = max(level_of_dependencies(wire)) + 1
 func (system *System) updateLevel(cID int, c Iterable) {
 	system.lbOutputs = system.lbOutputs[:0]
-	system.lbHints = map[*HintMapping]struct{}{}
+	lbHints := map[*HintMapping]struct{}{}
 	level := -1
 	wireIterator := c.WireIterator()
 	for wID := wireIterator(); wID != -1; wID = wireIterator() {
 		// iterate over all wires of the R1C
-		system.processWire(uint32(wID), &level)
+		system.processWire(lbHints, uint32(wID), &level)
 	}
 
 	// level =  max(dependencies) + 1
@@ -38,7 +38,7 @@ func (system *System) updateLevel(cID int, c Iterable) {
 	}
 }
 
-func (system *System) processWire(wireID uint32, maxLevel *int) {
+func (system *System) processWire(lbHints map[*HintMapping]struct{}, wireID uint32, maxLevel *int) {
 	if wireID < uint32(system.GetNbPublicVariables()+system.GetNbSecretVariables()) {
 		return // ignore inputs
 	}
@@ -59,11 +59,11 @@ func (system *System) processWire(wireID uint32, maxLevel *int) {
 		// constraint are the output of the same hint, and input to parent hint are themselves
 		// computed with a hint, we can suffer.
 		// (nominal case: not too many different hints involved for a single constraint)
-		if _, ok := system.lbHints[h]; ok {
+		if _, ok := lbHints[h]; ok {
 			// skip
 			return
 		}
-		system.lbHints[h] = struct{}{}
+		lbHints[h] = struct{}{}
 
 		for _, hwid := range h.Outputs {
 			system.lbOutputs = append(system.lbOutputs, uint32(hwid))
@@ -71,7 +71,7 @@ func (system *System) processWire(wireID uint32, maxLevel *int) {
 		for _, in := range h.Inputs {
 			for _, t := range in {
 				if !t.IsConstant() {
-					system.processWire(t.VID, maxLevel)
+					system.processWire(lbHints, t.VID, maxLevel)
 				}
 			}
 		}
diff --git a/constraint/system.go b/constraint/system.go
index fc7e0eea..e3a78d9e 100644
--- a/constraint/system.go
+++ b/constraint/system.go
@@ -126,9 +126,8 @@ type System struct {
 	bitLen int      `cbor:"-"`
 
 	// level builder
-	lbWireLevel []int                     `cbor:"-"` // at which level we solve a wire. init at -1.
-	lbOutputs   []uint32                  `cbor:"-"` // wire outputs for current constraint.
-	lbHints     map[*HintMapping]struct{} `cbor:"-"` // hints we processed in current round
+	lbWireLevel []int    `cbor:"-"` // at which level we solve a wire. init at -1.
+	lbOutputs   []uint32 `cbor:"-"` // wire outputs for current constraint.
 
 	CommitmentInfo Commitment
 }
@@ -144,7 +143,6 @@ func NewSystem(scalarField *big.Int) System {
 		MHintsDependencies: make(map[solver.HintID]string),
 		q:                  new(big.Int).Set(scalarField),
 		bitLen:             scalarField.BitLen(),
-		lbHints:            map[*HintMapping]struct{}{},
 	}
 }
```

When the solver prepares the dependencies between the levels, it keeps a track which hint as already been called in the same level. Right now the map which stores the information is stored in `System` (and flushed at every level). As the keys are pointers to `HintMapping`, then they are not very stable and reflect.DeepEqual fails.

With this modification we instead pass the cache as an argument and do not store it in `System`. Alternatively, we could flush it after solving, but I think this may be accidentally optimized out in the future ("why do we do pointless op here...").

Pinging @gbotrel, as imo he implemented the level-solver.

gbotrel:
ha didn't realized the problem was that localized into the level builder;
I think simplest is just to move the map / slice reset at the end of update level:
```diff
diff --git a/constraint/level_builder.go b/constraint/level_builder.go
index c2b1d8c1..d87fd57a 100644
--- a/constraint/level_builder.go
+++ b/constraint/level_builder.go
@@ -9,8 +9,6 @@ package constraint
 // We build a graph of dependency; we say that a wire is solved at a level l
 // --> l = max(level_of_dependencies(wire)) + 1
 func (system *System) updateLevel(cID int, c Iterable) {
-	system.lbOutputs = system.lbOutputs[:0]
-	system.lbHints = map[*HintMapping]struct{}{}
 	level := -1
 	wireIterator := c.WireIterator()
 	for wID := wireIterator(); wID != -1; wID = wireIterator() {
@@ -36,6 +34,9 @@ func (system *System) updateLevel(cID int, c Iterable) {
 	} else {
 		system.Levels[level] = append(system.Levels[level], cID)
 	}
+
+	system.lbOutputs = system.lbOutputs[:0]
+	system.lbHints = map[*HintMapping]struct{}{}
 }
 
 func (system *System) processWire(wireID uint32, maxLevel *int) {

```

gbotrel:
(re your concern @ivokub -- we can just add a comment to ensure we keep the re-init of the slices at the end of the function. not a big deal anyway; since the original problem we want to be sure is correctly solved is; if I compile my circuit twice, I want the same deterministic list of constraint so that my setup is still valid)

ivokub:
Yep, your change+comment is definitely less invasive. I'll make a fix.

ivokub:
I created #533. It fixes the cause. I think this PR won't be necessary after that. @Tabaie, what do you think?

ivokub (changes requested):
I think this PR isn't necessary after #533. I recommend closing.

PR #530, opened 2023-03-07
feat: add bounded comparator functions

aybehrouz:
This PR adds some functions, with relatively low circuit complexity, that perform comparison when the difference of two numbers is bounded.

Sometimes the fuzzer fails, and I have no idea why.
I had written these functions, I thought they might come in handy, so I made a PR. It should be reviewed carefully before merging.


ivokub:
Hi @aybehrouz, thanks for the contribution. I had an initial look at the PR. I'm not sure it is worth it. As you write yourself in the comments then the user has to be really really careful to use it correctly (either solver error, undefined behaviour or correct behaviour depending on the configuration and the inputs). And I would avoid giving the users more tools to create unsound proofs.

Lets keep the PR open for now, so someone else on the team can also have a look to see if it makes sense to include.

aybehrouz:
Hi, thanks for taking the time @ivokub , I think the documentation currently is a bit confusing and looks somehow scary 😄 but using the functions is pretty straightforward. 

Assume we have two numbers `a` and `b` and we want to compare them. To use the functions, we just need to know an upper bound for their absolute difference: `|a - b|` i.e. `abs(a - b)`. For example, you may have two 32bit numbers. Obviously their absolute difference can't be higher than `2^32`. You don't even need to use a tight bound here, and you may use something like `2^35`.  However with a tighter bound you'll have less constraints.  

Now what happens if you were wrong about the bound? For example if our numbers were not 32bit? Well first of all that's a programming error. The complex part of the documentation explains this situation. Anyway, in summary, as long as the difference will not cause an overflow in the underlying field, you're gonna be fine.

I've used a new type of API for these functions that use per package configuration. I'm not sure if that is easy to use.

aybehrouz:
Interestingly I found a better implementation for the `Slice` function using this package, although this function is a bit different. With this function the prover fails when `end < start` while in our current implementation a zero array will be returned. 
```go

func Slice(api frontend.API, start, end frontend.Variable, input []frontend.Variable) []frontend.Variable {
	startMask := stepMask(api, len(input), start, 0, 1)
	endMask := stepMask(api, len(input), end, 0, 1)

	out := make([]frontend.Variable, len(input))
	for i := 0; i < len(out); i++ {
		out[i] = api.Mul(api.Sub(startMask[i], endMask[i]), input[i])
	}

	comparator := cmp.NewComparator(api, big.NewInt(int64(len(input))))
	comparator.AssertIsLessEq(start, end)

	return out
}
```

aybehrouz:
```
api.AssertIsLessOrEqual:
Total constraints:  1270
Showing nodes accounting for 1268, 99.84% of 1270 total
Dropped 2 nodes (cum <= 6)
----------------------------------------------------------+-------------
      flat  flat%   sum%        cum   cum%   calls calls% + context 	 	 
----------------------------------------------------------+-------------
                                               254 33.42% |   r1cs.(*builder).Select frontend/cs/r1cs/api.go:472
                                               253 33.29% |   r1cs.(*builder).AssertIsLessOrEqual frontend/cs/r1cs/api_assertions.go:107
                                               253 33.29% |   r1cs.(*builder).Select frontend/cs/r1cs/api.go:467
```
```
api.Cmp:
Total constraints:  3809
Showing nodes accounting for 3806, 99.92% of 3809 total
Dropped 2 nodes (cum <= 19)
----------------------------------------------------------+-------------
      flat  flat%   sum%        cum   cum%   calls calls% + context 	 	 
----------------------------------------------------------+-------------
                                               508 40.03% |   bits.toBinary std/math/bits/conversion_binary.go:79
                                               508 40.03% |   r1cs.(*builder).And frontend/cs/r1cs/api.go:425
                                               253 19.94% |   r1cs.(*builder).Select frontend/cs/r1cs/api.go:443
      1269 33.32% 33.32%       1269 33.32%                | r1cs.(*builder).AssertIsBoolean frontend/cs/r1cs/api_assertions.go:78

```
```
cmp.IsLessOrEqualUnsigned
Total constraints:  770
Showing nodes accounting for 768, 99.74% of 770 total
Dropped 5 nodes (cum <= 3)
----------------------------------------------------------+-------------
      flat  flat%   sum%        cum   cum%   calls calls% + context 	 	 
----------------------------------------------------------+-------------
                                               760 99.87% |   bits.toBinary std/math/bits/conversion_binary.go:79
                                                 1  0.13% |   cmp.BoundedComparator.IsLess std/math/cmp/bounded.go:162
       761 98.83% 98.83%        761 98.83%                | r1cs.(*builder).AssertIsBoolean frontend/cs/r1cs/api_assertions.go:78

```

std/math/cmp/generic.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/std/math/bits"
+	"math/big"
+)
+
+// IsLessUnsigned returns 1 if a < b, and returns 0 if a >= b. The comparison
+// will be unsigned and all field elements will be treated as positive numbers.
+// Therefore, If the underlying field is of prime order P, the elements will be
+// considered to represent integers in range [0, P-1].
+func IsLessUnsigned(api frontend.API, a, b frontend.Variable) frontend.Variable {
```
ivokub, line 13: Variables are field elements and canonical integer representation is in `[0, P-1]`. I think repeating that the methods are unsigned is redundant and potentially misleading. I would rename just to `IsLess` and `IsLessOrEqual` below.

ivokub (commented):
PR is a lot better now. There still are some stylistic suggestions.

What I still would like to avoid is having a notation of signed and unsigned integers. Even though you can assign negative values to the witness, then in practice during solving time gnark mod reduces the value. All arithmetic in-circuit is done on the field elements which do not have a sign.

I feel that this may lead to problems in the future if we say that inputs are signed ints etc. In general I think the PR looks good and I think it would be good to merge after rephrasing.

Btw, package is also missing package documentation and title for browsing.

std/math/cmp/generic.go
```diff
@@ -0,0 +1,91 @@
...
+			return 0
+		}
+	}
+	// Interestingly when one of the two numbers is a constant, using bit by bit
+	// comparison will produce 3 fewer constraints than using the BoundedComparator.
+	if useBoundedCmp && n <= api.Compiler().FieldBitLen()-2 &&
+		isNotConstant(api, a[n-1]) && isNotConstant(api, b[n-1]) {
+		diffBound := new(big.Int).Lsh(big.NewInt(1), uint(n))
+		diffBound.Sub(diffBound, big.NewInt(1))
+		comparator := NewBoundedComparator(api, diffBound, false)
```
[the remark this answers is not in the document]
aybehrouz: That's a nice feature. However it cannot solve the problem of redundant constraints completely. The parameter could be constrained using custom equations (that could be very complex) or simply it may be a public input. 

Therefore, I think the programmer should always have a way to turn off the binary constraints for parameters. If you think without constraining parameters this function is not safe, we can use the same approach as `bits.ToBinary` and use an enum to turn it off and on.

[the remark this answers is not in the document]
ivokub: There is `api.MarkBoolean` which records the variable in the table without adding new constraints. So if the user wants to provide bits which they hav itself decomposed somehow, then they have to call `api.MarkBoolean` on the bits.

If possible I would not add options.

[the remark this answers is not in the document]
aybehrouz: ah all right, I did not know we had that function.

std/math/cmp/generic.go
infinite-pursuits: Hey @aybehrouz I am having issues due to exactly the unsigned comparison issue. Here is me discussing the issue : https://github.com/ConsenSys/gnark/discussions/705 . Can you help me understand how should I deal with these negative integers?

aybehrouz: The proof system uses a finite field of prime order P. (i.e. it has P elements) Since prime fields are isomorphic to modular arithmetic, we can assume that our arithmetic is performed modulus P. The elements of modular arithmetic are equivalence classes and all numbers of the form `x + k*P` belong to the same equivalence class. (all have the same remainder when divided by P) Modular arithmetic is like ordinary integer arithmetic but numbers map into equivalence classes. Therefore, you can use it for any arithmetic as long as you can **uniquely** determine your numbers from these equivalence classes.

For signed arithmetic, the most natural way is to have the same number of positive and negative numbers, so we assume that our numbers are in range [-P/2,P/2]. Then, we will have to make sure that the result of any operation is never less than -P/2 or greater that P/2. (I'm not trying to state boundary cases accurately here) 

gnark internally uses positive numbers for representing equivalence classes, and all number are shown in range [0, P-1]. Therefore, -1 is shown as P-1, -2 as P-2, etc... `api.Cmp` also assumes the same representation, so if you want to check that a number is negative, you should check if it is in range [P/2, P-1]. (boundaries are not accurate)

infinite-pursuits: Thanks a lot @aybehrouz . How do I fix the value of P based on my inputs or is it fixed internally? if the later, then how do I know the value of P to be able to do this?

aybehrouz: You can query the value of P by `api.Compiler().Field()`.

aybehrouz: Also be careful about overflows:
 > Then, we will have to make sure that the result of any operation is never less than -P/2 or greater that P/2.

infinite-pursuits: Thanks a lot. This helped immensely.

infinite-pursuits: How does this conversion thing work with ```api.Div```? In my circuit I am dividing an odd number by 2. So the result is a decimal, resulting in big number and constraint failing because I don't know what to write in ```AssertIsEqual ```.

aybehrouz: Division works by using the multiplicative inverse. If the result of your division is not an integer you'll get weird results. For example if P=11 the multiplicative inverse of 2 would be 6. Therefore, division by 2 works by multiplying into 6. If you want to divide 4 by 2, it would be `4 * 6 mod 11` which is 2. However, if you divide 3 by 2 you'll get `3 * 6 mod 11` which is 7.

I strongly recommend that you consider using a ZK-VM for your task, instead of using gnark. gnark is a low level library for creating compact ZK circuits. You'll need to be familiar with ZK proof systems, and you can easily create insecure circuits. A ZK-VM on the other hand provides a convenient tool for verifying arbitrary computation without the need for digging into the details.

infinite-pursuits: Thanks that makes a lot of sense. I did not know about ZK-VMs. Can I please have some links for some good/common ZKVMs? Why don't people use ZK-VMs often? or do they? I am assuming I would have to write the computations as functions in Golang and then feed the functions to ZK-VMs? Is that how they work?
Also this is for an academic project where I need exact time needed to verify and lengths of proofs. So I would need compact circuits, I guess the ZK-VMs will have redundancy in the circuits.

aybehrouz: > Also this is for an academic project where I need exact time needed to verify and lengths of proofs. So I would need compact circuits, I guess the ZK-VMs will have redundancy in the circuits.

Oh, if that's the case, then you can't use a ZK-VM. You'd have to use a low level library, like gnark, for creating specialized circuits. A VM always uses the same circuit for every computation.

infinite-pursuits: Thanks!

ivokub (approved):
I'm really really sorry, @aybehrouz. I have been busy with Linea and EthCC and completely forgot about the PR.

Thanks for the fixes and your contribution! It is really appreciated!

PR #531, opened 2023-03-07
Nix build

morganthomas:


gbotrel:
@morganthomas hi! I suspect target for this PR was your fork? :-)

morganthomas:
Hi @gbotrel ! This PR is pointed at its intended target but it's just a draft. There are some issues with these changes: some hacks I did to make the tests pass in the Nix build. I am wondering if the project is in principle interested in having Nix build code upstreamed, and if so then I could look at making these changes upstreamable by reducing the hacks.

gbotrel:
not familiar with Nix; why would we want a Nix build?

morganthomas:
Hi, @gbotrel . The benefit Nix offers is reproducible builds, so that without any hassle, one can download some source code and produce a build which is for all intents and purposes the same as the one the developers built from the same source.

PR #533, opened 2023-03-07
fix: stable levelbuilder hint mapping

ivokub:
Fixes cause from #529.

`System.lbHints` wasn't stable over different compiles and our deterministic compilation check failed. Now clean the internal cache where we use pointers.

### DIFF
--- a/constraint/level_builder.go
+++ b/constraint/level_builder.go
@@ -10,9 +10,9 @@ package constraint
 // --> l = max(level_of_dependencies(wire)) + 1
 func (system *System) updateLevel(cID int, c Iterable) {
 	system.lbOutputs = system.lbOutputs[:0]
-	system.lbHints = map[*HintMapping]struct{}{}
 	level := -1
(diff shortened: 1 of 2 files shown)


constraint/level_builder.go
```diff
@@ -10,9 +10,9 @@ package constraint
 // --> l = max(level_of_dependencies(wire)) + 1
 func (system *System) updateLevel(cID int, c Iterable) {
 	system.lbOutputs = system.lbOutputs[:0]
```
gbotrel, line 12: shouldn't we move lbOutputs at the end too ?

ivokub: It seems not. The values are computed deterministically and is stable over different runs. However, lbHints keys are *HintMapping and their values (pointers) depend when we allocate and into where.

ivokub: But for clarity, it makes sense to do it.

gbotrel (approved):

PR #538, opened 2023-03-07
feat: gnark/profile now filter frontend private method fixes #537

### LINKED ISSUE #537: bug: gnark profiler with plonk doesn't filter call stack well
```
09:51:27 INF gnark profiling disabled path=./scs.pprof
Showing nodes accounting for 93802626, 100% of 93802626 total
Dropped 62 nodes (cum <= 469013)
      flat  flat%   sum%        cum   cum%
  93802626   100%   100%   93802626   100%  scs.(*builder).addPlonkConstraint /Users/gbotrel/dev/go/pkg/mod/github.com/consensys/gnark@v0.7.2-0.20230306141737-db76d9e43be5/frontend/cs/scs/builder.go:152
```

is not very useful as a profile :)

gbotrel:


profile/profile_worker.go
```diff
@@ -123,3 +106,27 @@ func collectSample(pc []uintptr) {
 	}
 
 }
+
+func filterSCSPrivateFunc(f string) bool {
+	const scsPrefix = "github.com/consensys/gnark/frontend/cs/scs.(*builder)."
+	if idx := strings.Index(f, scsPrefix); idx != -1 && len(f) > (idx+len(scsPrefix)) {
```
ivokub, line 112: Is there any case where idx isn't 0 or 1? Isn't the question essentially "does f start with scsPrefix". Can there be anything before? Maybe for readibility could then instead use `strings.HasPrefix`?

profile/profile_worker.go
```diff
@@ -123,3 +106,27 @@ func collectSample(pc []uintptr) {
...
+		// filter plonk frontend private APIs from the trace.
+		c := []rune(f)[idx+len(scsPrefix)]
+		if unicode.IsLower(c) {
+			return true
+		}
+	}
+	return false
+}
+
+func filterR1CSPrivateFunc(f string) bool {
```
ivokub, line 122: rename const to `r1csPrefix`? But imo unimportant.

ivokub (approved):
Looks good. Maybe comments make sense.

PR #546, opened 2023-03-08
fix #527 when fuzzer reports bug for inconsistency between bit decomposition with constant and with variable

### LINKED ISSUE #527: Test fails for a trivial circuit
Tests fail for the following simple circuit:
```go

type testCircuit struct {
	N frontend.Variable `gnark:",public"`
}

func (c *testCircuit) Define(api frontend.API) error {
	bits.ToBinary(api, c.N, bits.WithNbDigits(3))
	return nil
}

func TestTest(t *testing.T) {
	assert := test.NewAssert(t)

	assert.ProverSucceeded(&testCircuit{}, &testCircuit{N: 3})
}
```
```
=== RUN   TestTest/fuzz/bn254/groth16
    assert.go:80: errVars [assertIsEqual] 4 == 14914344159867638993849239390888276500196600227666715437247040087537991462988
        bits.toBinary
        	conversion_binary.go:93
        bits.ToBase
        	conversion.go:25
        bits.ToBinary
        	conversion_binary.go:18
        bits_test.(*testCircuit).Define
        	compare_test.go:15
        
    assert.go:80: errConsts <nil>
```

gbotrel:


std/math/bits/conversion_binary.go
```diff
@@ -69,6 +69,24 @@ func toBinary(api frontend.API, v frontend.Variable, opts ...BaseConversionOptio
 		for i := 0; i < len(bits); i++ {
 			bits[i] = c.Bit(i)
 		}
+
```
aybehrouz, line 72: Why do you need to treat constants differently? Can't we just remove that `if c, ok := api.Compiler().ConstantValue(v); ok`?

ivokub: Either this or imo we can also do `if c.Bitlen() > cfg.NbDigits {panic}`?

ivokub (commented):
This change is definitely needed. Imo the inconsistency always lead to problems in test engine when we set all variables to consts vs vars.

I'm not sure about the implementation though, I think could be simpler (either removing separate path for consts or using `Int.Bitlen()` instead)

gbotrel:
```
--- FAIL: TestConstantEqual (3.05s)
[6047](https://github.com/ConsenSys/gnark/actions/runs/4370742004/jobs/7645959621#step:6:6048)
    --- FAIL: TestConstantEqual/Goldilocks/limb=64 (0.02s)
[6048](https://github.com/ConsenSys/gnark/actions/runs/4370742004/jobs/7645959621#step:6:6049)
        --- FAIL: TestConstantEqual/Goldilocks/limb=64/fuzz (0.00s)
[6049](https://github.com/ConsenSys/gnark/actions/runs/4370742004/jobs/7645959621#step:6:6050)
            --- FAIL: TestConstantEqual/Goldilocks/limb=64/fuzz/bn254/groth16 (0.00s)
[6050](https://github.com/ConsenSys/gnark/actions/runs/4370742004/jobs/7645959621#step:6:6051)
            --- FAIL: TestConstantEqual/Goldilocks/limb=64/fuzz/bn254/plonk (0.00s)
```
cc @ivokub  :)

std/math/bits/conversion_binary.go
aybehrouz: Oh, Thank you so much for saying that! 😄

I think if we handle constants properly in the lower level functions, like `api.Mul`, `api.Add` and `api.AssertIsEqual`, or if we use some type of optimization pass that optimizes constants in constraints, then we should not need to handle constants in the higher level functions. 

Handling constants in every function can make code very ugly and many higher level functions with high probability will never be called with a constant. (`ToBinary` itself could be an example for that)

aybehrouz (approved):
Thanks for requesting a review from me.😊

IMO when the constraints are being generated at lower levels, constants could be optimized, and higher level optimizations should not be needed.

std/math/bits/conversion_binary.go
gbotrel: the part that "worries" me is mostly the hint calls. Calling a hint with `big.Int` is costly in the sense that the inputs to the hint will be serialized with the constraint system, and resolved at solving time. Moreover, the constants will be stored as coefficients in the list of coefficients. That's a lot of gymnastic for something that could be handled at compile time.

aybehrouz: Interesting... That's a good point.

PR #547, opened 2023-03-09
feat: reintroduce hints for field emulation

ivokub:
We used to have hints when we hade fake-API. @yelhousni recommends reintroducing as it is helpful for optimising non-native algebra (pairings etc). When I tried to reintroduce the implementation then I realized that actually the old code (already removed few months ago) was kind of broken. We wrapped the hint functions and created anonymous hint functions, which cannot really work.

So, I rewrote it, making defining hints slightly more inconvenient. So instead of writing:
```go
func HintFn(mod *big.Int, inputs []*big.Int, outputs []*big.Int) error {
    // work here
}
```
we do
```go
func HintFn(mod *big.Int, inputs, outputs []*big.Int) error {
	return emulated.UnwrapHint(inputs, outputs,
		func(mod *big.Int, inputs, outputs []*big.Int) error { // NB we shadow initial input, output, mod to avoid accidental overwrite!
                         // here all inputs and outputs are modulo nonnative mod. we decompose them automatically
			// work here
		})
}
```

Keeping draft for now as it seems really convenient for internal hint uses in field emulation (div. inv etc.).

### DIFF
--- a/std/math/emulated/field_hint.go
+++ b/std/math/emulated/field_hint.go
@@ -19,7 +19,10 @@ func (f *Field[T]) wrapHint(nonnativeInputs ...*Element[T]) []frontend.Variable 
 	return res
 }
 
-func UnwrapHint(nativeInputs, nativeOutputs []*big.Int, nonNativeHint solver.Hint) error {
+// UnwrapHint unwraps the native inputs into nonnative inputs. Then it calls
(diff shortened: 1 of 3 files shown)


ivokub:
@yelhousni 🏓

yelhousni (approved):
That works sweet! 👍

PR #549, opened 2023-03-09
feat: lazify the constraints repeatable used; added poseidon for test

ZhAnGeek:
### details:
1. record repeatable constraints structure (during this process, we will record the static constraints **StaticConstraints**, and it will be stored in map by a key defined by user, for looking up the constraint need to be recovered)
2. record repeatable constraints's inputs, and pick only the input constrains to store. (**GeneralLazyInputs**)
3. lazify and remove all the repeatable r1cs structure, with only lazy inputs left. rebuild levels.
4. recover the r1c when solving. the input constraints can be easily extracted from store, for others we have a shift recorded the distance from the static constraints, and add the offset shift to the variables to recover.

### usage:
please check gnark/constraint/lazy/


ivokub:
Thanks @ZhAnGeek, seems very interesting. Do I understand correctly that this PR doesn't reduce the the actual number of constraints, but rather compresses similar constraints (similar being if the user provides the same key) during compile and during solving we lazily recover these constraints?

I think this actually may be of independent interest - we have an upcoming GKR verifier implementation in-circuit and this GKR proofs really benefit from this kind of structure in the circuit. So if we have subcircuit of size `N` being called `k` times then right now the cost is `N*k`. If I recall correctly then with GKR it would be `log(k)*N` (plus some overhead so actually `k` has to be quite large to have real-life improvement).

ZhAnGeek:
> Thanks @ZhAnGeek, seems very interesting. Do I understand correctly that this PR doesn't reduce the the actual number of constraints, but rather compresses similar constraints (similar being if the user provides the same key) during compile and during solving we lazily recover these constraints?
> 
> I think this actually may be of independent interest - we have an upcoming GKR verifier implementation in-circuit and this GKR proofs really benefit from this kind of structure in the circuit. So if we have subcircuit of size `N` being called `k` times then right now the cost is `N*k`. If I recall correctly then with GKR it would be `log(k)*N` (plus some overhead so actually `k` has to be quite large to have real-life improvement).

Thank you, for your question `this PR doesn't reduce the the actual number of constraints,` the answer is yes, the r1cs are still the same when solving and proving, but we don't need them stay in memory any more. This will benefit the memory usage especially for the circuit that heavily use hashes like merkle tree system. And such structures we now only maintain one with only some meta data memory additionally.

For GKR it sounds really cool. And yes I am keeping up with your GKR upgrading, seems like it reduced real constraints for the same structure(like we only need to plug in the verifier for it) ? But I remember users need to write an arithmetic circuit, can it be done autonomously? I cant wait to try :)

gbotrel:
hi -- 
To confirm, can you clarify what happens  when we have a list of constraints like so: 
```
1. (a+b+c) * (d + e) == r1
2. (f + g + h + k) * (d + e) == r2
3. (f + g + h) * (a + b +c) == r3
```

--> is the idea that you are going to store a constraint that look like 
```
plaholder1 * (d + e) == placeholder2
```

and replace dynamically placeholder1 and placeholder2 at runtime? (ie so basically saving storing `d+e` n times if it appears in n constraint on the `R` part of the `L * R == O` constraint?

gbotrel:
+ can you provide a benchmark of the impact of this on some circuits (with and without: for poseidon or mimc maybe?)

ZhAnGeek:
> hi -- To confirm, can you clarify what happens when we have a list of constraints like so:
> 
> ```
> 1. (a+b+c) * (d + e) == r1
> 2. (f + g + h + k) * (d + e) == r2
> 3. (f + g + h) * (a + b +c) == r3
> ```
> 
> --> is the idea that you are going to store a constraint that look like
> 
> ```
> plaholder1 * (d + e) == placeholder2
> ```
> 
> and replace dynamically placeholder1 and placeholder2 at runtime? (ie so basically saving storing `d+e` n times if it appears in n constraint on the `R` part of the `L * R == O` constraint?

I think it is not the case, let's imagine a giant circuit with inputs `i1,i2`, assume that v is the internal variables generated, c is the constant variables used, the circuit which will generate constraints below,
1. (i1 + c1) * (i1 + c1) = v1
2. (i2 + c2) * (i2 + c2) = v2
3. (v1 + c3) * (v1 + c3) = v3
4. i1 * i2 = v3
5. v3 * v3 = v4
...then any other constraints without i1/i2's term into it.
n. ...constraints

if later we got the same circuit, with inputs `i3,i4`, assume the v internal generated to be 1000, then the constraints will be
1. (i3 + c1) * (i3 + c1) = v1001
2. (i4 + c2) * (i4 + c2) = v1002
3. (v1000 + c3) * (v1000 + c3) = v1003
4. i1 * i2 = v1003
5. v1003 * v1003 = v1004
. ...then any other constraints without i3,i4's term into it.
n. ...constraints

as you can see, there will be many internal variables generated, and if in same circuit, the variables' born and order are the same, except that they got a shift like the internal variables always differ 1000.

so we save the structure static constraints for the first time
1. (i1 + c1) * (i1 + c1) = v1
2. (i2 + c2) * (i2 + c2) = v2
3. (v1 + c3) * (v1 + c3) = v3
4. i1 * i2 = v3
...then any other constraints without i1/i2's term into it.
n. ...constraints

and we recover the other by 1. retrieve the input constraints, which contains the input's term. this is exactly the same with normal constraints 2. retrieve contraints and add shift to the variables. in above case, we have shift 1000, so in constraint 5 we can recover (v3 * v3 = v4 in static constraints) to (v1003 * v1003 = v1004)

ZhAnGeek:
this is the experiment data for dumping the r1cs to file (for a living prover, another part pk memory usage will not be affected here), and the mimc testing circuit for testing was like below.

```go=
	for i := 0; i < times; i++ {
		mimc.Write(circuit.PreImage, circuit.PreImage, circuit.PreImage, circuit.PreImage)
		api.AssertIsDifferent(mimc.Sum(), circuit.Hash)
		mimc.Reset()
	}
```

as you can see, for such circuit we save nearly 70% of memory, and both lazified and non-lazified grow linearly with the mimc hashes num, that's because we use input constraints to recover true constraints. If the inputs constraints contained is very little, then the memory of r1cs used actually will be very little.

mimc | 1 | 10 | 100 | 1000 | 10000
-- | -- | -- | -- | -- | --
dump_r1cs_w/o_origin (kb) | 127 | 1,225 | 12,939 | 135,828 | 1,364,706
dump_r1cs_w/o_lazify (kb) | 244 | 1,342 | 13,056 | 135,946 | 1,364,833
dump_r1cs_w/_lazify (kb) | 158 | 479 | 4,055 | 43,072 | 435,292

so I test for poseidon 12 inputs, which will surely the input constraints take less in the whole its own static constraints.

poseidon | 1 | 10 | 100 | 1000 | 10000
-- | -- | -- | -- | -- | --
dump_r1cs_w/o_origin (kb) | 313 | 2,520 | 24,602 | 273,890 | N/A
dump_r1cs_w/o_lazify (kb) | 546 | 2,761 | 24,915 | 275,049 | N/A
dump_r1cs_w/_lazify (kb) | 325 | 450 | 1,745 | 16,962 | N/A

this time it saves nearly 95% of memory used in r1cs.

gbotrel:
perf looks great, will take a look at that this week 👍

gbotrel:
This seems to have also quite some potential to speed up solving time. 
It seems to raise couple of of interesting points though;
1. The constraint/ architecture currently don't allow for an elegant modification like this to be done easily for all the builders. Similarly, it intersects with our thoughts about custom gates (architecture wise). 
2. We ask the user to decorate block of constraints that are "hot" and repeats a lot. In some circuits, users may "over optimize" and it may hurt readability; yet, I suspect if we make it free in constraint/ to have these blocks of "lazy" constraints, many places could benefit. For example emulated arithmetic do a lot of bit decomposition, range checks. ECC operations are also producing constraints with very similar shape. It seems this capability to say "declare to the compiler that this function/block of constraint is instantiated many times with different wires id" should be a bit more baked in.
3. In the current PR, I think we will hit some issues with nested blocks, with Hints and with Logs. 

Will keep the PR open, not asking you to fix any of these points, we need a bit more thinking on that one! 

Do you know if other libraries have similar concepts? (i.e; not store a flat list of constraint).

gbotrel:
mmhh there is another issue; this is a dangerous API, since the constraint that are marked as a group; between `api.RecordConstraintsForLazy` , nothing guarantee that going down this path, everything is "stateless" and purely functional. 
That is, actual constraints may different between invocations, and I don't like relying on the user (rather than the frontend / compiler) to ensure that.

gbotrel:
just a quick note on that one; doing a refactor of the constraint/ package that's inspired by this PR, a first version should land in develop in couple of days 👍

PR #557, opened 2023-03-10
perf: plonk ccs serialization

gbotrel:
- perf: don't add debug info for each bool unless debug tag is set
- feat,perf: remove *HintMapping from System struct, use int id instead


ivokub (approved):
LGTM

PR #561, opened 2023-03-13
refactor: expose all typed backends in gnark/backend (moved from internal/)

gbotrel:


gbotrel:
Probably, the big ugly type switches we have in `backend/groth16/groth16.go` & `plonk.go` should go in an `internal/` package; we use them for test purposes to test across backends, but most users probably don't care about testing their circuit across curves;..

ivokub (approved):
Two no-op duplicate directory make, but otherwise good.

ivokub:
Now when the methods are not internal anymore it would be good if the package documentation would be a bit more thorough, but I would keep it as a future work as part of a bigger documentation overhaul.

Issue #394, opened 2022-11-10
`Compiler.Commit(...)`


Tabaie:
Using Fiat-Shamir inside SNARKs can be prohibitive when the statement is large due to the large number of variables to be hashed. As per Alex, Azam and Olivier's paper, add a `Commit` function that provides a commitment to those variables efficiently, regardless of their visibility.

Suggested change:
```go
type Compiler interface {
  ⋮
  Commit(...Variable) Variable
}
```

Use:
```go
func (c *MyCircuit) Define(api frontend.API) {
  ⋮ // compute variables v1, v2, v3, v4 of any visibility
  transcript := fiatshamir.NewTranscript(api, someHash, "first", "second")
  commitment := api.Compiler().Commit(v1, v2, v3, v4)
  transcript.Bind("first", commitment)
  ⋮
}
```

The `Commit` function can only be called once during `Define`.

ivokub:
I like the simple interface to commit the variables.

I think it makes sense that it is possible to call this method only once. Otherwise there may be strange behaviour (for example, when compiling in parallel and the `Commit` functions are called in non-deterministic orders etc.), although I think this can be solved (we order the variables by their order in the witness, for example). But in the context of composability, it may be difficult to use. So lets say there are different gadgets/sub-circuits which want to commit to the elements, but they cannot (because it may already be called or someone else is expecting to call it). One option would to have some kind of accumulator etc. and the accumulated variables would later be committed asynchronously, but then it would be difficult to obtain the commitment itself 🤔.

And, I guess the method should also return errors. Imo there may be several cases where an error can happen -- we build vanilla-R1CS/PLONK without commitments (where in which case we should manually compute the commitment, or have some kind of other fallback); the inputs are somehow invalid; method has already been called etc.

At some point I was thinking about having some `Defer`/`Finalise` method for the compiler which allow to register functions which are to be called when circuits `Define` method has returned. To extend this idea for the commitments, we could have something like:
```
type Compiler interface {
    Commit(in... Variable) error // adds in to be commited
    PostCommitment(finaliser func(api API, commitment Variable) error) error
    ...
}
```
Now, after the `Define` method of the circuit returns, we compute the commitment and call all `finaliser` functions (`PostCommitment` can be called multiple times) with the commitment and API, such that we can add further constraints/gates. If the constraint system doesn't support commitments, then `PostCommitment` returns an error which allows the caller to implement fallbacks if needed. In the finaliser method it is now invalid to call `Commit`.

[...]

PR #562, opened 2023-03-14
feat: add gadget for enabling multiple commitments in-circuit

ivokub:
In the end I took the callback approach. It seemed the most cleanest. I understand that the callback-based approach is too bit JavaScript-y, but it was quite straightforward to implement and I'm quite confident it is correct.

@Tabaie - how does this look to you particularly?

### DIFF
--- a/frontend/compile.go
+++ b/frontend/compile.go
@@ -131,8 +131,8 @@ func parseCircuit(builder Builder, circuit Circuit) (err error) {
 }
 
 func callDeferred(builder Builder) error {
-	for i, cb := range circuitdefer.GetAll[func(API) error](builder) {
-		if err := cb(builder); err != nil {
+	for i := 0; i < len(circuitdefer.GetAll[func(API) error](builder)); i++ {
+		if err := circuitdefer.GetAll[func(API) error](builder)[i](builder); err != nil {
 			return fmt.Errorf("defer fn %d: %w", i, err)
 		}
 	}
--- a/internal/circuitdefer/defer.go
+++ b/internal/circuitdefer/defer.go
@@ -15,9 +15,7 @@ func Put[T any](builder any, cb T) {
 	}
 	val := kv.GetKeyValue(deferKey{})
 	var deferred []T
-	if val == nil {
-		deferred = []T{cb}
-	} else {
+	if val != nil {
 		var ok bool
 		deferred, ok = val.([]T)
 		if !ok {
--- a/std/commitments/native/doc_test.go
+++ b/std/commitments/native/doc_test.go
@@ -1,0 +1,104 @@
+package native_test
+
+import (
+	"fmt"
+
+	"github.com/consensys/gnark-crypto/ecc"
+	"github.com/consensys/gnark/backend/groth16"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/frontend/cs/r1cs"
+	"github.com/consensys/gnark/std/commitments/native"
+)
+
+// MultipleCommitmentCircuit is an example circuit showing usage of multiple
+// independent commitments in-circuit.
+type MultipleCommitmentsCircuit struct {
+	Secrets [4]frontend.Variable
+}
+
+func (c *MultipleCommitmentsCircuit) Define(api frontend.API) error {
+	// first callback receives first unique commitment derived from the root commitment
+	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
+		// compute (X-s[0]) * (X-s[1]) for a random X
+		res := api.Mul(api.Sub(commitment, c.Secrets[0]), api.Sub(commitment, c.Secrets[1]))
+		api.AssertIsDifferent(res, 0)
+		return nil
+	}, c.Secrets[:2]...)
+
+	// second callback receives second unique commitment derived from the root commitment
+	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
+		// compute (X-s[2]) * (X-s[3]) for a random X
+		res := api.Mul(api.Sub(commitment, c.Secrets[2]), api.Sub(commitment, c.Secrets[3]))
+		api.AssertIsDifferent(res, 0)
+		return nil
+	}, c.Secrets[2:4]...)
+
+	// we do not have to pass any variables in if other calls to [WithCommitment] have
+	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
+		// compute (X-s[0]) for a random X
+		api.AssertIsDifferent(api.Sub(commitment, c.Secrets[0]), 0)
+		return nil
+	})
+
+	// we can share variables between the callbacks
+	var shared, stored frontend.Variable
+	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
+		shared = api.Add(c.Secrets[0], commitment)
+		stored = commitment
+		return nil
+	})
+	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
+		api.AssertIsEqual(api.Sub(shared, stored), c.Secrets[0])
+		return nil
+	})
+	return nil
+}
+
+// Full written on how to use multiple commitments in a circuit.
+func ExampleWithCommitment() {
+	circuit := MultipleCommitmentsCircuit{}
+	assignment := MultipleCommitmentsCircuit{Secrets: [4]frontend.Variable{1, 2, 3, 4}}
+	ccs, err := frontend.Compile(ecc.BN254.ScalarField(), r1cs.NewBuilder, &circuit)
+	if err != nil {
+		panic(err)
+	} else {
+		fmt.Println("compiled")
+	}
+	pk, vk, err := groth16.Setup(ccs)
+	if err != nil {
+		panic(err)
+	} else {
+		fmt.Println("setup done")
+	}
+	secretWitness, err := frontend.NewWitness(&assignment, ecc.BN254.ScalarField())
+	if err != nil {
+		panic(err)
+	} else {
+		fmt.Println("secret witness")
+	}
+	publicWitness, err := secretWitness.Public()
+	if err != nil {
+		panic(err)
+	} else {
+		fmt.Println("public witness")
+	}
+	proof, err := groth16.Prove(ccs, pk, secretWitness)
+	if err != nil {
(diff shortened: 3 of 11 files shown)


ivokub:
Just a comment how this works. Previously we computed the commitment as
```
C = Commit(x1, x2, .., xn)
```
where variables `x1, ..., xn` were commited to and the commitment is `C`. When trying to compute multiple commitments we get a panic
```
C = Commit(x1, x2, ..., xn)
C2 = Commit(y1, y2, ..., yn) // panic, we can only compute a single commitment per circuit
```

With this PR, we propose a framework to collect all variables to commit and the functions where the commitments will be used. Essentially, we do
```
root = Commit(x1, x2, ..., xn, y1, y2, ..., yn)
C1 = Hash(1, root)
C2 = Hash(2, root)
// now can use C1 and C2 as independent commitment in-circuit
```

For the hash function we use MiMC hash. The scheme is binding because due to preimage resistance of MiMC, we are prevented from computing suitable `root`. The commitments are independent because MiMC is collision resistant.

Tabaie (approved):
LGTM! However I do think in the long run there's no escaping the need to enable multiple `api.Commit` calls. The only way I can imagine it now would involve more and even dirtier hint hacks than we have now.

ivokub:
> LGTM! However I do think in the long run there's no escaping the need to enable multiple `api.Commit` calls. The only way I can imagine it now would involve more and even dirtier hint hacks than we have now.

Great, thanks! I'll wait with merging for a bit, I think @gbotrel also had ideas about the interface. Don't want to have something unidiomatic sticking around for ages.

gbotrel:
mmhh overall I like that it's a simple PR; but, the name "native" confused me at first. Also, when we have introduce the risk to have something like 

```go 
c.X = api.Add(...)
native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
 		// compute (X-s[0]) * (X-s[1]) for a random X
 		c.X = api.Mul(c.X, api.Sub(commitment, c.Secrets[1]))
 		api.AssertIsDifferent(c.X, 0)
 		return nil
 	}, c.Secrets[:2]...)
 Y = api.Mul(c.X, ...)
```

(ie we put forward an "async" programming model while the rest of the API is not)

gbotrel:
```go
// we can share variables between the callbacks
	var shared, stored frontend.Variable
	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error { // cb4
		shared = api.Add(c.Secrets[0], commitment)
		stored = commitment
		return nil
	})
	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error { // cb5
		api.AssertIsEqual(api.Sub(shared, stored), c.Secrets[0])
		return nil
	})
```

huh

gbotrel:
question; since the `api.Commit` value is the result of a hint. couldn't we make that hint stateful and handle the logic in the consecutive calls to the hint? (which is injected in a single solving instance of the constraint system anyway?)

gbotrel:
@Tabaie I'm probably missing a step but technically;
```
v1 := api.Commit(s1, s2)
v2 := api.Commit(s3,s4)
``` 
how are v1 and v2 computed behind the scenes so we have a single commitment in the proving key ?

Tabaie:
In this PR we don't actually make multiple calls to `api.Commit`. The gadget registers callbacks and aggregates all committed values into a single call to `api.Commit`.
Was that your question or did I misunderstand?

gbotrel:
yeah my question is; what if we don't do this PR with async callbacks, and do multiple api.Commit; what does it take behind the scene? Can't the same "defer" logic be handled at the hint level by being "stateful" and maintaining a single commitment in the proving key? 

to put it bluntly, I don't understand why it's complicated to modify the hint that handles api.Commit to do multi commit, because I don't understand what it is "morally" supposed to do if we were to have multiple api.Commit calls

ivokub:
> ```go
> // we can share variables between the callbacks
> 	var shared, stored frontend.Variable
> 	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error { // cb4
> 		shared = api.Add(c.Secrets[0], commitment)
> 		stored = commitment
> 		return nil
> 	})
> 	native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error { // cb5
> 		api.AssertIsEqual(api.Sub(shared, stored), c.Secrets[0])
> 		return nil
> 	})
> ```
> 
> huh

Essentially I wanted to show that it is possible to share variables between the callbacks, if done carefully (understanding that the callbacks are executed FIFO order). But not that I recommend doing it.

ivokub:
> mmhh overall I like that it's a simple PR; but, the name "native" confused me at first. Also, when we have introduce the risk to have something like
> 
> ```go
> c.X = api.Add(...)
> native.WithCommitment(api, func(api frontend.API, commitment frontend.Variable) error {
>  		// compute (X-s[0]) * (X-s[1]) for a random X
>  		c.X = api.Mul(c.X, api.Sub(commitment, c.Secrets[1]))
>  		api.AssertIsDifferent(c.X, 0)
>  		return nil
>  	}, c.Secrets[:2]...)
>  Y = api.Mul(c.X, ...)
> ```
> 
> (ie we put forward an "async" programming model while the rest of the API is not)

Agree, `native` is not good name for the package. But when trying to fit it in `std/commitment`, then seemed suitable. But actually I think something like `std/randomoracle` is even more suitable, because this is imo actually what this package provides.

ivokub:
> yeah my question is; what if we don't do this PR with async callbacks, and do multiple api.Commit; what does it take behind the scene? Can't the same "defer" logic be handled at the hint level by being "stateful" and maintaining a single commitment in the proving key?
> 
> to put it bluntly, I don't understand why it's complicated to modify the hint that handles api.Commit to do multi commit, because I don't understand what it is "morally" supposed to do if we were to have multiple api.Commit calls

Technically, it can be done but imo it has several consequences (see also discussion in #394):
* we need to have separate commitment key for every call to `Commit`. And the verifier needs to verify every commitment separately. So the verification wouldn't be three pairings anymore, but 3+nbCommitments.
* I'm also afraid it will cause soundness problems. Essentially the commitment is a Pedersen hash of the inputs for a basis which depends on the proving key. An adversary would be able to predict the value of second commitment from the first commitment if the inputs to the commitments are similar. See for example grjte writeup which imo captures the problem well: https://hackmd.io/@grjte/zk-hack-5-strong-adaptivity. Now, if we do it like in this PR (only one Pedersen hash as a commitment and then MiMC for separation), then we do not have the issue as all commitments are independent from each other (thanks to collision-resistance of MiMC).

gbotrel:
mmhh got you and couldn't we do a non-async programming model ? ie have 
```
v1 := api.Commit(s1, s2)
v2 := api.Commit(s3,s4)
```
and let the frontend package do the MiMC for separation between calls or something?

ivokub:
> mmhh got you and couldn't we do a non-async programming model ? ie have
> 
> ```
> v1 := api.Commit(s1, s2)
> v2 := api.Commit(s3,s4)
> ```
> 
> and let the frontend package do the MiMC for separation between calls or something?

I think theoretically we can, but I think implementation-wise it would be a bit fragile. What I'm thinking, is that we can do something like this:
```go
var collected []frontend.Variable
var retvars []frontend.Variable

func (builder *r1cs) AsyncCommit(vars ...frontend.Variable) {
    collected = append(collected, vars...)
    retvar := builder.newInternalVariable()
    return retvar
}

func (builder *r1cs) RunAsyncedCommitments() {
    rootcommitment := builder.Commit(collected...)
    for i := range retvars {
        ctxcommit := mimc.Sum(i+1, rootcommitment)
        builder.addConstraint(1 * retvars[i] = ctxcommit)
    }
}

func (c *Circuit) Define(api frontend.API) error {
    v1 := api.AsyncCommit(s1, s2)
    v2 := api.AsyncCommit(s3,s4)
    // do stuff with v1 v2 here
    return nil
    // now we run deferred api.RunAsyncedCommitments() to generate the constraints.
    // the solver then first computes the commitment, ctxcommitments and then solve everything which depends on v1 and v2
}
```

But it isn't that easy - in `AsyncCommit` we also have to check that we do not have any recursive cycles (we asynccommit to something which is a linear combination of a commitment). Maybe if it is obvious?

Second thing is that we have to embed multi-commitment in the builders. Which isn't a big problem actually.


(now looking at the example it seems that the second approach is actually a lot better if we can get it right)

gbotrel:
ok you both convinced me that this is a necessary evil (and simple solution), that hopefully won't get too invasive 👍

gbotrel:
(I like syntax of second example better I think, but I'll let you judge here -- if doing first version unblocks things let's go w ith that, mark it experimental and re-iterate later down the road)

Tabaie:
@ivokub I'm a bit torn between requiring an `api` input to the callback or not. If we're mostly expecting lambdas here they can just use the `api` variable already available. I'm not sure though (that the callback would always be lambdas.)

ivokub:
> @ivokub I'm a bit torn between requiring an `api` argument to the callback or not. If we're mostly expecting lambdas here they can just use the `api` variable already available and have a simpler interface. I'm not sure though (that the callback would always be lambdas.)

Hmm, I think theoretically we can without api in callbacks, but the passed implementations may actually be different. For example we could pass non-commiting version of the builder to the callback (as committing in the callback is illegal anyway). This ensures that we cannot type assert api to `frontend.Committer`.

But I'm not directly opposed to it. I'll try to figure out if the non-async version doesn't work out.

ivokub:
@gbotrel, ping.

I tried out the synchronous approach in [1ca9426](https://github.com/ConsenSys/gnark/pull/562/commits/1ca94267937270af1018f078383326c2f3156ce2). It seems to work (at least the test doesn't fail). Maybe you can have a quick look if this makes sense? If so, then I will try out in bigger circuits and will polish into a nice PR.

There is still some work to do:
* dunno if wire levels should use slice of ints instead of map. It was just easier to work it that way.
* there is one suboptimal loop in `AssignToLazyVariable`. We could track the value instead of big loop.
* Implement for PLONK
* Think about the commitment separation (right now in PR use squaring, but maybe need hashing. Or maybe have something parametric?)
* maybe something else...

I'm not 100% confident though that is is correct :/

ivokub:
Removed the sync variant from the PR. It seems to work but waiting for other optimisations which handle the wire solving dependency graph.

Also moved the package to internal to prevent using externally.

PR #563, opened 2023-03-14
refactor: based on #515 generify groth16 MPC setup for all curves, flatten packages+ refactor

gbotrel:
Based on #515 (cherry picked the commits from @HSG88 ); 

Did mostly stylistic edits; flattened the package structure, avoided couple of allocs, and generify for all curves.

To be merged after #561 ;  MPC setup is now `gnark/backend/groth16/bn254/mpcsetup` .

A future PR (cc @HSG88 ) would probably be nice to ensure APIs don't panic but return errors + struct and public methods are documented (you can test output with `godoc`) . 

On the perf note, the individual scalar mul on affine coordinates are a clear bottleneck, maybe we should use jacobian? (+ optimize the memory usage of the `mulGLV` path cc @yelhousni  in `gnark-crypto`) . 

- groth16 MPC setup
- fix Z in preparePhase2
- parallelize coefficients multiplication
- parallelize scalar multiplication
- refactor: expose all typed backends in gnark/backend (moved from internal/)
- refactor: flatten mpc structure, idomify APIs
- test: ensure phase2 serialization is tested
- test: added reference benchmark
- refactor: setup -> mpcsetup
- refactor: move utils in mpcsetup; limit api surface
- refactor: minor code cleaning
- build: generify mpcsetup for all curves


internal/generator/backend/template/zkpschemes/groth16/mpcsetup/lagrange.go.tmpl
```diff
@@ -0,0 +1,201 @@
...
+
+	{{- template "import_fr" . }}
+	{{- template "import_curve" . }}
+	{{- template "import_fft" . }}
+	"github.com/consensys/gnark/internal/utils"
+)
+
+
+
+func LagrangeCoeffsG1(powers []curve.G1Affine, size int) []curve.G1Affine {
```
ivokub, line 16: I think it can be made private. Used only within package and imo makes looking at the package methods otherwise difficult.

internal/generator/backend/template/zkpschemes/groth16/mpcsetup/lagrange.go.tmpl
```diff
@@ -0,0 +1,201 @@
...
+
+	utils.Parallelize(size, func(start, end int) {
+		for i := start; i < end; i++ {
+			coeffs[i].ScalarMultiplication(&coeffs[i], &invBigint)
+		}
+	})
+	return coeffs
+}
+
+func LagrangeCoeffsG2(powers []curve.G2Affine, size int) []curve.G2Affine {
```
ivokub, line 37: Can be made private.

ivokub (changes requested):
I think it looks good. Didn't go through the math exactly, hoping that @ThomasPiellard had a look.

Considering that the usage is more complex, I would recommend adding documentation example and a bit of package documentation. If sounds good, then can also do it myself.

Maybe would be good to add type-switched methods into `backend/groth16`?

PR #579, opened 2023-03-16
fix: subtraction overflow computation bug

ivokub:
When we subtract `a-b` in field emulation then we instead do
```
d + a - b
```

The value `d` is padding and a multiple of emulated modulus. It is constructed in a way that `d+a-b` would never underflow the field (i.e. it is one bit longer than `b` and high bit always set). Previously I had set the estimated overflow for the subtraction result to be `max(a.overflow, b.overflow+2)`, but this doesn't take into account that can overflow `a`. So correct would be `max(a.overflow+1, b.overflow+2)`.

Now, as the overflow of the result is always monotonically larger than for the inputs (previously could have been overflow of a), then we ran into an infinite recursion. It happened because in modular reduction we compute difference of `a-b` using automatically reducing version of subtraction. To prevent this from happening, we decrease the maximal allowed overflow by one and use (newly implemented) non-reducing version of subtraction inside modular reduction.


gbotrel (approved):
makes sense, however, I really want to get to this SMT solver thing, would help to identify these edge cases automatically.

ivokub:
> makes sense, however, I really want to get to this SMT solver thing, would help to identify these edge cases automatically.

Yep, it would have been perfect here.

PR #583, opened 2023-03-16
feat: range checks using log derivative, fixes #581

### LINKED ISSUE #581: Use logarithmic derivatives instead of exponentiation for range checks
Currently for verifying in circuit that (f_i)_i is a subset of a table (s_i)_i, we record the f_i, use a hint to compute the multiplicity e_i of the f_i and we check using a random challenge x that Pi_i (x-s_i)^e_i = Pi_i (x-f_i). The e_i are themselves internal variables and this computation leads to a lot of exponentiations.

We should use the logarithmic derivatives (taken from Caulk) to rather check that Sum_i e_i / (x-s_i) = Sum_i 1 / (x-f_i). Those are rational functions, but as long as x is not in the set of poles (that is it does not belong to {f_1, .., f_n} ) Schwartz Zippel lemma (univariate case) works exactly as for polynomials (in fact for polynomials we just avoid the set of poles which is the infinity when sampling a random challenge).

This would save all the exponentiations by the e_i.

Here is the [ref](https://eprint.iacr.org/2022/1530.pdf) for the logarithmic derivative

### LINKED ISSUE #582: Challenge computed prior to sensitive data (range checker)
To check that (f_i)_i is a subset of (s_i)_i, we verify in circuit that Pi_i (x-f_i) = Pi_i (x-s_i)^e_i where the e_i are the purported multiplicities of the f_i. The e_i are computed using a hint and the challenge x is derived using Fiat Shamir.

The issue is that x is derived using (f_i) only, prior to computing the (e_i). If  x is such that one of the (x - s_i) generates F_r* we know that there exist exponents e_i such that Pi_i (x-f_i) = Pi_i (x-s_i)^e_i no matter the f_i (even if the e_i is likely to be big). But we can imagine that it's possible for a prover to craft a set of e_i that verifies Pi_i (x-f_i)=Pi_i (x-s_i)^e_i, where the e_i are in the correct range, and (f_i) not a subset of (s_i).

ThomasPiellard:
Fixes #581 
Fixes #582 

Bench `TestCheck` in `std/rangecheck/` with 1000 values:
Before: 15647 constraints
After: 12129 constraints

I expect the gain to be way more noticeable as the number of values to check grows.

So far the compilation is extremely slow -> I'll try to find what is wrong


ivokub:
Just FYI - there is also this method https://github.com/ConsenSys/gnark/blob/develop/std/rangecheck/rangecheck_commit.go#L191-L238. It tries to estimate the number of constraints for different table sizes and chooses parameters which minimise it. Right now we look at the sizes between 2^2 ... 2^17. Using product approach, for biggest circuits (ECDSA, pairing) the optimal size was approx 2^11 ... 2^13 (depending if Groth16 or PLONK), but I would think that with this one the bigger is better. I think it should be easy to estimate (for Groth16 `tablesize + len(decomposed)` and for PLONK `3*(tablesize + len(decomposed))` I think).

ThomasPiellard:
Yes I figured there might be some parameters to tweak from the implementation but haven't touch them. Now the compilation is extremely slow though :/

ivokub:
And to speed up compilation, maybe `WithCompressThreshold` helps (https://pkg.go.dev/github.com/consensys/gnark@v0.8.0/frontend#WithCompressThreshold)? You are going to have huge linear expressions in Groth16. When I needed it, then I used value 500 etc, but I think now with the most recent optimisations it could be bigger.

ThomasPiellard:
yes you are right it must be that. For some reason I thought this problem was solved automatically in a previous pr

ivokub:
I tested it out and seems to be the case of large linear expressions. When checking 100k 64bit vars and setting compress threshold to 1000 then it compiles in 8 seconds (Groth16). Without this argument it was time-wise the same for 1k vars.

ivokub:
Another thing - what do you think about using `api.DivUnchecked` when computing the table with multiplies? It should be safe (with high probability) when we include e_i in the commitment.

ivokub:
Update PR a bit:
* exact computation of the number of constraints for better prediction of the table size
* included exponents in the commitment
* default test with threshold compression
* use unchecked division
* update package documentation

I'll see what the CI does with it. I think there are compile tests which depend on field emulation and thus on range checker. But I think we should have some kind of default for the compression threshold (a few hundred) to make compilation fast. Otherwise even though we have the option I'm afraid we're going to get a lot of issues due to slow compile.

ThomasPiellard:
Yes I agree with using divUnchecked. It's impossible that it happens accidentally, if so someone must have broken the hash function to find the pre-image of one of the {s_i}. I don't know if we should keep both implementations or just this one. I expect the gain using this one to be huge if the redundancies (so the e_i) are big

ivokub:
> Yes I agree with using divUnchecked. It's impossible that it happens accidentally, if so someone must have broken the hash function to find the pre-image of one of the {s_i}. I don't know if we should keep both implementations or just this one. I expect the gain using this one to be huge if the redundancies (so the e_i) are big

I think it makes sense to keep only this one. `e_i` is `log(number of checked elements)` for every `e_i` (as all values can fall into a single bucket) and for any practical circuit its going to be large (1K+++). In essence this would give significant saving.

Lets wait for the CI to finish but I think I'm done on my side. If CI doesn't bother too much then I'm good to merge.

I created a new issue #590 because I think that with this implementation it will start to really hurt otherwise.

ivokub (approved):
LGTM. Good idea for optimisation!

ivokub:
CI fails. One failure is due to not updating the statistics and easy to fix.

The other one is with unreplaced commitment hint function. If we do not replace it (e.g. if we run only the solver which happens in fuzzing) then the current implementation returns 0 as commitment. But this fails DivUnchecked. The fix is also easy, but will do in separate PR.

ivokub:
@ThomasPiellard - I'm satisfied. I think we can merge as is. I resorted to doing the unreplaced commitment hint in this PR.

But we also need #599, so it would be good to review and merge that PR so that circuits with this PR compile quickly :)

PR #586, opened 2023-03-17
406 bsb22 commitments plonk

Tabaie:
Introducing BSB22-based `api.Commit` for Plonk. The commitment is KZG, using the same SRS as Plonk. Like in the Groth16 version, it is hashed onto the fr field using Sha256 and then usable in Fiat-Shamir challenges.

Consistency with the Plonk witness is enforced by replacing the constraint equation $q_L.f_L+q_R.f_R+q_O.f_O+q_M.f_L.f_R+(q_C+PI)=0$ with $q_L.f_L+q_R.f_R+q_O.f_O+q_M.f_L.f_R+(q_C+PI_1+q'_C.PI_2)=0$ as if this feature is a custom gate. $PI_2$ is the commitment shipped to the verifier and $q'_C$ is a sparse preprocessed binary polynomial (i.e. with values $0,1$ on the subgroup.)
For any committed value, say $w_m$ or the $m$'th value in the witness vector $w$, we add a constraint of the form $q_L(j)=-1, f_L(j)=w_m, PI_2(j)=w_m$ thus binding the value in the commitment to the one used in normal constraints.
The hashed-to-field commitment $C$ is injected into the circuit as if it were a public value. That is, for some $j$ we have a constraint where $q_L(j)=-1$ and other selectors are $0$. When computing $PI(\zeta)$, the verifier simply adds $C.L_j(\zeta)$ to the sum.


frontend/cs/scs/api.go
```diff
@@ -559,6 +561,34 @@ func (builder *builder) Compiler() frontend.Compiler {
...
+func scsBsb22CommitmentHintPlaceholder(*big.Int, []*big.Int, []*big.Int) error {
+	return errors.New("placeholder - should never be called")
+}
+
+func (builder *builder) Commit(v ...frontend.Variable) (frontend.Variable, error) {
+
+	committed := make([]int, len(v))
+
+	for i, vI := range v { // TODO @Tabaie Perf; If public, just hash it
+		vINeg := builder.Neg(vI).(expr.Term)
```
ThomasPiellard, line 573: Perhaps we should handle the error here if the variable is constant (or anything else than a Term) ? Otherwise we get a message like this: `parse circuit: interface conversion: frontend.Variable is *big.Int, not expr.Term`, not sure if it's very explicit.

frontend/cs/scs/api.go
```diff
@@ -559,6 +561,34 @@ func (builder *builder) Compiler() frontend.Compiler {
...
 
+func scsBsb22CommitmentHintPlaceholder(*big.Int, []*big.Int, []*big.Int) error {
+	return errors.New("placeholder - should never be called")
+}
+
+func (builder *builder) Commit(v ...frontend.Variable) (frontend.Variable, error) {
+
+	committed := make([]int, len(v))
+
+	for i, vI := range v { // TODO @Tabaie Perf; If public, just hash it
```
ThomasPiellard, line 572: should handle the public inputs here
```
tvI := vI.(expr.Term)
if tvI.VID < builder.cs.GetNbPublicVariables() {
	continue
}
```

ThomasPiellard:
LGTM, perhaps for a later PR we should definitely remove the constraints linked to the commitments of public inputs, as these constraints are duplicated at the beginning of the circuit when it is compiled, could blow up the number of constraints if there are a lot of public inputs...

ThomasPiellard (approved):

ThomasPiellard (approved):

Tabaie:
> LGTM, perhaps for a later PR we should definitely remove the constraints linked to the commitments of public inputs, as these constraints are duplicated at the beginning of the circuit when it is compiled, could blow up the number of constraints if there are a lot of public inputs...

My original plan was to let both prover and verifier conventionally hash the public committed values, but after today's standup discussion seems like that's off the table and we'll have to stick to this method to make the verifier's life easier.

frontend/cs/scs/api.go
```diff
@@ -559,6 +560,30 @@ func (builder *builder) Compiler() frontend.Compiler {
...
+		committed[i] = builder.cs.GetNbConstraints()
+		builder.addPlonkConstraint(sparseR1C{xa: vINeg.VID, qL: vINeg.Coeff, commitment: constraint.COMMITTED})
+	}
+	outs, err := builder.NewHint(cs.Bsb22CommitmentComputePlaceholder, 1, v...)
+	if err != nil {
+		return nil, err
+	}
+	commitmentVar := builder.Neg(outs[0]).(expr.Term)
+	commitmentConstraintIndex := builder.cs.GetNbConstraints()
+	builder.addPlonkConstraint(sparseR1C{xa: commitmentVar.VID, qL: commitmentVar.Coeff, commitment: constraint.COMMITMENT}) // value will be injected later
```
gbotrel, line 578: what is this constraint exactly? 
-xa + __ == 0

?

frontend/cs/scs/api.go
Tabaie: Addressed in new commit 👍

PR #593, opened 2023-03-21
Perf/plonkprover

gbotrel:
To merge after #586 . 

Improve perf of plonk prover on 96core target machine. No major algorithmic changes, mostly less mem allocs / moves, more parallelism in the prover.


ThomasPiellard:
I just modified at little thing in the setup, one of the polynomials to handle is just X, so we can save some operations

ThomasPiellard (approved):

PR #599, opened 2023-03-24
feat: set default compression threshold

### LINKED ISSUE #590: feat: add default compress threshold for R1CS compile
References #583.

With the log-derivative variant of the range check gadget we get very large linear expressions which make compile times quite slow. We should set some reasonable default. I think a few hundred should suffice.

This makes estimating the constraint counts a bit difficult, but I think it is a necessary evil.

ivokub:
I ended up with 300. After that even with large circuits the constraint number reduction is insignificant and before that the compile time overhead is small (but get approx 4% less constraints for pairing circuit).

Resolves #590.


ivokub:
It is good to have this PR with #583, otherwise circuits with a lot of range checks compile very slowly.

ThomasPiellard (approved):

PR #610, opened 2023-03-30
secp256r1: ecdsa

SherLzp:
Emulated secp256r1 based ECDSA.
Constraints:
379982


ivokub:
Thanks for the contribution! I had overlooked this PR and implemented the same, but with also P384.

I merged this PR to include the tests and extended to P-384.

PR #625, opened 2023-04-06
perf: use `api.Select` for 2 to 1 mux

aybehrouz:
A 2 to 1 multiplexer is a special case and has an efficient solution which can not be generalized to other cases:
```
out == sel * i_1 + (1 - sel) * i_0
sel * (sel - 1) == 0
```
Therefore, we should use `api.Select` for constructing 2 to 1 multiplexers as a special case.

### DIFF
--- a/std/selector/multiplexer.go
+++ b/std/selector/multiplexer.go
@@ -14,6 +14,7 @@ package selector
 
 import (
 	"fmt"
+	"github.com/consensys/gnark/std/math/bits"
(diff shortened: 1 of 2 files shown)


yelhousni (approved):
LGTM

aybehrouz:
> @aybehrouz and how does the 4 to 1 MUX compare to `api.Lookup2` ?

You're right. I should add that as a special case too. 😅

PR #631, opened 2023-04-07
Feat: unified ECADD

### LINKED ISSUE #630: `ECADD` precompile: input points can be equal
the `ECADD` precompile uses the `Add()` method which makes the assumption that the input points are different and nonzero. We use `Add()` in places where we are guaranteed of this assumption (`ScalarMul()` and `ScalarMulBase()`) but the precompile does NOT make this assumption.

yelhousni:
Fixes #630.

→ ~~*Suggested solution*:~~

~~Select `λ` to be either `(y1-p.y2)/(x1-x2)` or `3x1²/2y1` based on `x1==x1 && y1==y2`. The problem is that division by `x1-x2` will trigger an error as it uses big.Int `ModInverse` https://github.com/ConsenSys/gnark/blob/c02da5f61944d13c557d2b390ca56d5a26c597a1/std/math/emulated/hints.go#L251~~
~~So we implement a special method `DivSpecial()` that returns 0 when denominator is 0. We use it to compute `(y1-p.y2)/(x1-x2)` and proceed further, and then select `3x1²/2y1` if applicable.~~

Actually we can use [Brier-Joye](https://link.springer.com/content/pdf/10.1007/3-540-45664-3_24.pdf) unified addition formulas, which compute a slope that works for both addition and doubling in affine coordinates — and also save half the constraints of the previous solution.

### DIFF
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -144,8 +144,8 @@ func (c *Curve[B, S]) AddUnified(p, q *AffinePoint[B]) *AffinePoint[B] {
 
 	// λ = ((p.x+q.x)² - p.x*q.x + a)/(p.y + q.y)
 	pxqx := c.baseApi.MulMod(&p.X, &q.X)
-	num := c.baseApi.Add(&p.X, &q.X)
-	num = c.baseApi.MulMod(num, num)
+	pxplusqx := c.baseApi.Add(&p.X, &q.X)
+	num := c.baseApi.MulMod(pxplusqx, pxplusqx)
 	num = c.baseApi.Sub(num, pxqx)
 	if c.addA {
 		num = c.baseApi.Add(num, &c.a)
@@ -158,8 +158,7 @@ func (c *Curve[B, S]) AddUnified(p, q *AffinePoint[B]) *AffinePoint[B] {
 
 	// x = λ^2 - p.x - q.x
 	xr := c.baseApi.MulMod(λ, λ)
-	xr = c.baseApi.Sub(xr, &p.X)
-	xr = c.baseApi.Sub(xr, &q.X)
+	xr = c.baseApi.Sub(xr, pxplusqx)
 
 	// y = λ(p.x - xr) - p.y
 	yr := c.baseApi.Sub(&p.X, xr)
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -96,13 +96,49 @@ func (c *Curve[B, S]) AssertIsEqual(p, q *AffinePoint[B]) {
 }
 
 // Add adds p and q and returns it. It doesn't modify p nor q.
+//
+// ⚠️  p must be different than q and both nonzero.
+//
 // It uses incomplete formulas in affine coordinates.
-// The points p and q should be different and nonzero (neutral element).
 func (c *Curve[B, S]) Add(p, q *AffinePoint[B]) *AffinePoint[B] {
+	return c.add(p, q, false)
+}
+
+// AddSafe adds p and q and returns it. It doesn't modify p nor q.
+//
+// ✅ p can be equal to q, but none nonzero.
+//
+// It uses incomplete formulas in affine coordinates.
+func (c *Curve[B, S]) AddSafe(p, q *AffinePoint[B]) *AffinePoint[B] {
+	return c.add(p, q, true)
+}
+
+// add adds p and q and returns it. It doesn't modify p nor q.
+// It uses incomplete formulas in affine coordinates.
+func (c *Curve[B, S]) add(p, q *AffinePoint[B], safe bool) *AffinePoint[B] {
+
 	// compute λ = (q.y-p.y)/(q.x-p.x)
 	qypy := c.baseApi.Sub(&q.Y, &p.Y)
 	qxpx := c.baseApi.Sub(&q.X, &p.X)
-	λ := c.baseApi.Div(qypy, qxpx)
+
+	// if qxpx == 0, set λ to 0
+	λ := c.baseApi.DivSpecial(qypy, qxpx)
+
+	if safe {
+		// compute _λ = (3p.x²+a)/2*p.y
+		xx3a := c.baseApi.MulMod(&p.X, &p.X)
+		xx3a = c.baseApi.MulConst(xx3a, big.NewInt(3))
+		if c.addA {
+			xx3a = c.baseApi.Add(xx3a, &c.a)
+		}
+		y2 := c.baseApi.MulConst(&p.Y, big.NewInt(2))
+		_λ := c.baseApi.Div(xx3a, y2)
+
+		selector := c.api.And(
+			c.baseApi.IsZero(qxpx), c.baseApi.IsZero(qypy),
+		)
+		λ = c.baseApi.Select(selector, _λ, λ)
+	}
 
 	// xr = λ²-p.x-q.x
 	λλ := c.baseApi.MulMod(λ, λ)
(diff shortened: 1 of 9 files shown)


yelhousni:
We use `zero=(0,0)` as the affine "representation" of the infinity point.
- [x] `UnifiedAdd(P, zero)` should output `P`
- [x] `UnifiedAdd(zero, P)` should output `P`
- [x] `UnifiedAdd(zero, zero)` should output `zero`
- [x] `UnifiedAdd(P, -P)` should output `zero`
- [x] `ScalarMul(0, P)` should output `zero`
- [x] `ScalarMul(s, zero)` should output `zero`
- [x] `ScalarMulBase(0)` should output `zero`
-----
- These methods are used in the EVM precompiles, so they should follow the `(0,0)` convention. 
- For `double()`, `triple()`, `add()` and `doubleAndAdd()` (now private) we don't handle the edge cases (used when these should not happen).
- For `Pair()` the zkEVM arithmetization would filter the `(0,0)` points @OlivierBBB.

yelhousni:
fixes #634

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -120,9 +122,69 @@ func (c *Curve[B, S]) Add(p, q *AffinePoint[B]) *AffinePoint[B] {
...
+	}
+	denum := c.baseApi.Add(&p.Y, &q.Y)
+	// if p.y + q.y = 0, assign dummy 1 to denum and continue
+	selector3 := c.baseApi.IsZero(denum)
+	denum = c.baseApi.Select(selector3, c.baseApi.One(), denum)
+	λ := c.baseApi.Div(num, denum)
+
+	// x = λ^2 - p.x - q.x
+	xr := c.baseApi.MulMod(λ, λ)
+	xr = c.baseApi.Sub(xr, &p.X)
```
ivokub, line 158: We have already computed `p.X+q.X` previously. Could reuse it instead? Would save one sub in plonkish.

ivokub:
Suggested edit:
```diff
diff --git a/std/algebra/emulated/sw_emulated/point.go b/std/algebra/emulated/sw_emulated/point.go
index 69e32594..0ef6cd01 100644
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -126,12 +126,12 @@ func (c *Curve[B, S]) add(p, q *AffinePoint[B]) *AffinePoint[B] {
 //
 // ✅ p can be equal to q, and either or both can be (0,0).
 // (0,0) is not on the curve but we conventionally take it as the
-// neutral/infinity point as per the EVM [EYP].
+// neutral/infinity point as per the [EVM].
 //
-// It uses the unified formulas of Brier and Joye [BriJoy02] (Corollary 1).
+// It uses the unified formulas of Brier and Joye ([[BriJoy02]] (Corollary 1)).
 //
 // [BriJoy02]: https://link.springer.com/content/pdf/10.1007/3-540-45664-3_24.pdf
-// [EYP]: https://ethereum.github.io/yellowpaper/paper.pdf
+// [EVM]: https://ethereum.github.io/yellowpaper/paper.pdf
 func (c *Curve[B, S]) AddUnified(p, q *AffinePoint[B]) *AffinePoint[B] {
 
 	// selector1 = 1 when p is (0,0) and 0 otherwise
@@ -338,7 +338,7 @@ func (c *Curve[B, S]) Lookup2(b0, b1 frontend.Variable, i0, i1, i2, i3 *AffinePo
 //
 // ✅ p can can be (0,0) and s can be 0.
 // (0,0) is not on the curve but we conventionally take it as the
-// neutral/infinity point as per the EVM [EYP].
+// neutral/infinity point as per the [EVM].
 //
 // It computes the standard little-endian variable-base double-and-add algorithm
 // [HMV04] (Algorithm 3.26).
@@ -352,7 +352,7 @@ func (c *Curve[B, S]) Lookup2(b0, b1 frontend.Variable, i0, i1, i2, i3 *AffinePo
 //
 // [ELM03]: https://arxiv.org/pdf/math/0208038.pdf
 // [HMV04]: https://link.springer.com/book/10.1007/b97644
-// [EYP]: https://ethereum.github.io/yellowpaper/paper.pdf
+// [EVM]: https://ethereum.github.io/yellowpaper/paper.pdf
 func (c *Curve[B, S]) ScalarMul(p *AffinePoint[B], s *emulated.Element[S]) *AffinePoint[B] {
 
 	// if p=(0,0) we assign a dummy (0,1) to p and continue
@@ -402,7 +402,7 @@ func (c *Curve[B, S]) ScalarMul(p *AffinePoint[B], s *emulated.Element[S]) *Affi
 //
 // ✅ When s=0, it retruns (0,0).
 // (0,0) is not on the curve but we conventionally take it as the
-// neutral/infinity point as per the EVM [EYP].
+// neutral/infinity point as per the [EVM].
 //
 // It computes the standard little-endian fixed-base double-and-add algorithm
 // [HMV04] (Algorithm 3.26).
@@ -413,7 +413,7 @@ func (c *Curve[B, S]) ScalarMul(p *AffinePoint[B], s *emulated.Element[S]) *Affi
 // [3]g, [5]g and [7]g points.
 //
 // [HMV04]: https://link.springer.com/book/10.1007/b97644
-// [EYP]: https://ethereum.github.io/yellowpaper/paper.pdf
+// [EVM]: https://ethereum.github.io/yellowpaper/paper.pdf
 func (c *Curve[B, S]) ScalarMulBase(s *emulated.Element[S]) *AffinePoint[B] {
 	g := c.Generator()
 	gm := c.GeneratorMultiples()

```

ivokub:
Suggested edit:
```diff
diff --git a/std/algebra/emulated/sw_emulated/point.go b/std/algebra/emulated/sw_emulated/point.go
index 69e32594..78b05b99 100644
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -400,7 +400,7 @@ func (c *Curve[B, S]) ScalarMul(p *AffinePoint[B], s *emulated.Element[S]) *Affi
 // ScalarMulBase computes s * g and returns it, where g is the fixed generator.
 // It doesn't modify s.
 //
-// ✅ When s=0, it retruns (0,0).
+// ✅ When s=0, it returns (0,0).
 // (0,0) is not on the curve but we conventionally take it as the
 // neutral/infinity point as per the EVM [EYP].
 //

```

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -120,9 +122,69 @@ func (c *Curve[B, S]) Add(p, q *AffinePoint[B]) *AffinePoint[B] {
...
+//
+// ✅ p can be equal to q, and either or both can be (0,0).
+// (0,0) is not on the curve but we conventionally take it as the
+// neutral/infinity point as per the EVM [EYP].
+//
+// It uses the unified formulas of Brier and Joye [BriJoy02] (Corollary 1).
+//
+// [BriJoy02]: https://link.springer.com/content/pdf/10.1007/3-540-45664-3_24.pdf
+// [EYP]: https://ethereum.github.io/yellowpaper/paper.pdf
+func (c *Curve[B, S]) AddUnified(p, q *AffinePoint[B]) *AffinePoint[B] {
```
ivokub, line 135: What do you think about renaming it to `Add` instead? We don't have public `AddIncomplete` anyway and it would be better anyway if the users would just use this safe method.

We could insead rename `add` to `addIncomplete`.

ivokub (approved):
* A few stylistic change suggest: for paper use "academic"-style links and for EVM just use it.
* An idea about renaming the methods.
* One potential optimisation.

Other than that I think it looks good.

std/algebra/emulated/sw_emulated/point.go
yelhousni: hmm usually when we say `Add` in Weierstrass EC it means implicitly that the formulas are not unified nor complete so one would expect a public method `Double`. I was rather thinking of re-naming it to `AddComplete` because we deal with the `(0,0)` case but it's not a "natural" completeness — so keeping `AddUnified` makes sense IMO.

yelhousni: I think in this case adding 
```go
func (c *Curve[B, S]) Add(p *AffinePoint[B]) *AffinePoint[B] {
    return c.AddUnified(p, p)
}

func (c *Curve[B, S]) Double(p *AffinePoint[B]) *AffinePoint[B] {
    return c.AddUnified(p, p)
}
```
Or just explaining in the package documentation that we use the unified formula which capture both point adds and doubling.  But I'm more in favour of the second option TBH.

ivokub: I think the second option is good.

ivokub:
Applied the changes. If looks good then go for merge.

PR #636, opened 2023-04-09
perf: add a generalized version of binary selection

aybehrouz:
This commit adds `BinaryMux` which is a generalization of `api.Select` and `api.Lookup2` for inputs of arbitrary length.


aybehrouz:
I think that the number of R1CS constraints for an input of length `n` (n power of 2) should be `log(n)+1` but `ccs.GetNbConstraints()` shows a higher value.

aybehrouz:
Undefined behavior happens only when `n` is not a power of 2, and it could be prevented by doing a comparison with a constant. The comparison can be done with `log(n)` constraints. `n + log(n)` will outperform the solution that needs adding wires with zero values to make n a power of 2.

ivokub:
Thanks for the contribution. I'll try to get to reviewing it this week.

std/math/bits/conversion_binary.go
```diff
@@ -63,6 +63,13 @@ func toBinary(api frontend.API, v frontend.Variable, opts ...BaseConversionOptio
 		}
 	}
 
+	// when cfg.NbDigits == 1, v itself has to be a binary digit. This if clause
```
ivokub, line 66: I think this optimisation is interesting on its own.

ivokub (commented):
I'm not sure it would be good addition. The functionality is similar to ordinary mux, but it takes in the binary decomposition of a value. As you have mentioned, then this may lead to undefined behaviour in cases where the inputs is not a power of two or if the inputs are not binary.

Considering that `std/` packages are the main tools a developer would be using when implementing the circuits, then I would rather keep it simple and as safe as possible without having some implicit assumptions. Even if the edge-cases are documented in function doc then in a lot of cases people do not read them properly. This leads to unsound circuits and potentially monetary losses.

But the optimisation for binary decomposition for `NbDIgits(1)` case is useful. I do recommend creating a separate PR for it.

aybehrouz:
@ivokub  
The current implementation of mux creates `3n` constraints. (Though it should generate `2n` and I don't know why it creates `3n`) In this implementation the number of generated constraints is only `n-1`. Since converting to binary only needs `log(n)` constraints, we can use this implementation to create selector with `n + log(n)` constraints for non-binary selectors too.

There is no edge cases. This is just a generalization of `api.Lookup2`. Trivially, a multiplexer with `n` bit selectors needs `2^n` inputs. You can just add a simple `if-clause` and `panic` when the input is not a power of 2 to make it safe. 

Currently, inside the general Mux, I'm using this implementation only for `n` that is a power of 2. By doing a comparison, we can use this implementation for all cases, without having undefined behavior, and create a Mux with `n + 2*log(n)` constraints.

aybehrouz:
And I should add that I personally do not recommend adding binary constraints for selector bits. That reduces the flexibility of the function. If a user has selectors that are already constrained he would not have any ways for preventing duplicated constraints.

aybehrouz:
For one bit selector, this function exactly generates the same constraints as `api.Select`. For 2 bits it creates exactly the same constraints as `api.Lookup2`. However it can handle any number of selector bits.

std/selector/multiplexer.go
```diff
@@ -105,6 +98,39 @@ func generateSelector(api frontend.API, wantMux bool, sel frontend.Variable,
...
+//	inputs[selBits[0]+selBits[1]*(1<<1)+selBits[2]*(1<<2)+...]
+//
+// len(inputs) does not need to be necessarily a power of 2. However, in that
+// case when the index represented by selBits is greater than or equal with
+// len(inputs), the output will be undefined.
+//
+// BinaryMux does NOT define any boolean constraints for selBits. Outside this
+// function, all the elements of selBits must be constrained to be binary
+// digits.
+func BinaryMux(api frontend.API, selBits, inputs []frontend.Variable) frontend.Variable {
```
aybehrouz, line 115: We can change the code to this:
```go
func BinaryMux(api frontend.API, selBits, inputs []frontend.Variable) frontend.Variable {
	if len(inputs) != 1<<len(selBits) {
		panic("invalid input length for BinaryMux")
	}
	return binaryMuxRecursive(api, selBits, inputs)
}
```

std/selector/multiplexer.go
```diff
@@ -105,6 +98,39 @@ func generateSelector(api frontend.API, wantMux bool, sel frontend.Variable,
...
+	// n does not need to be a power of 2.
+	if len(selBits) == 0 {
+		return inputs[0]
+	}
+
+	nextSelBits := selBits[:len(selBits)-1]
+	sel := selBits[len(selBits)-1]
+	pivot := 1 << len(nextSelBits)
+	if pivot >= len(inputs) {
+		return BinaryMux(api, nextSelBits, inputs)
```
aybehrouz, line 126: If we change this line to this
```go
return api.Mul(api.Sub(1, sel), BinaryMux(api, nextSelBits, inputs))
```
We will not have undefined behavior anymore, and the return value will be zero when the selector is outside. However it is not as efficient as using a comparison.

std/selector/multiplexer.go
ivokub: Agree

std/selector/multiplexer.go
ivokub: I think the previous approach is better (asserting input len is power of two)

ivokub:
Thanks for the clarification, makes a lot more sense now.

In that case I would drop the current implementation of `Mux` and just use BinaryMux within after decomposing the input. Then there is a single implementation and optimised path when the input is already binary decomposed.

You can still do `AssertIsBoolean` on the bits inside `BinaryMux`. gnark keeps track of all variables for which the function `MarkBoolean` has been called on does not create a binary constraint if already has been. So it should be no-op if the user provides bits which have been already binary constrained.

ivokub:
About the `Mux` function being sub-optimal -- imo in the PR which introduced it I thought about it quickly and seemed to be optimal. But now if `BinaryMux` is better, then there seems to be a contradiction.

aybehrouz:
> In that case I would drop the current implementation of Mux and 

For cases that n is not a power of 2, We'll need an efficient function for comparing two numbers which gnark currently doesn't have. If PR #530 gets merged, we can use this implementation for all cases.

ivokub:
> > In that case I would drop the current implementation of Mux and
> 
> For cases that n is not a power of 2, We'll need an efficient function for comparing two numbers which gnark currently doesn't have. If PR #530 gets merged, we can use this implementation for all cases.

Oh, sure, you're right. We cannot pad the input with dummy values as we don't know the selector value in `Mux`. I didn't think about it.
I'll have another look at this and #530 hopefully tomorrow.

ivokub (approved):
Thanks for the contribution and sorry for delay on my side.

After going back and forth on this PR I think it makes sense and is good addition to gnark. Regarding the PR - I just reverted the general `AssertBits` helper in `math/bits`. I saw that it was used only once and wouldn't write out as a helper just yet.

aybehrouz:
Thanks.

I added `bits.AssertBits` because in PR #530 I needed the assertion in several places. However, it gets added to the public API of the `bits` package, so if you think it is not very useful, we can remove it here, and in the other PR I'll use a private function instead. 

I spotted and fixed a few issues in the comments. Also I made a minor refactoring.

aybehrouz:
I can't figure out what's wrong with the staticcheck

ivokub:
> I can't figure out what's wrong with the staticcheck

Rerunning jobs. There were also some other odd failures last week.

std/selector/multiplexer.go
```diff
@@ -32,76 +30,112 @@ func GetHints() []solver.Hint {
...
+// exactly one of its inputs based on sel. The index of inputs starts from zero.
 //
-// sel needs to be between 0 and n - 1 (inclusive), where n is the number of inputs, otherwise the proof will fail.
+// sel needs to be between 0 and n - 1 (inclusive), where n is the number of
+// inputs, otherwise the proof will fail.
 func Mux(api frontend.API, sel frontend.Variable, inputs ...frontend.Variable) frontend.Variable {
-	return generateSelector(api, true, sel, nil, inputs)
+	// we use BinaryMux when len(inputs) is a power of 2 and its math.Log2 is an
+	// integer.
+	if log := math.Log2(float64(len(inputs))); log == math.Trunc(log) {
```
ivokub, line 59: Could also do `bits.Onescount(len(inputs)) == 1`, see https://pkg.go.dev/math/bits#OnesCount. I would always prefer to use floats due to imprecision.

ivokub (approved):
Looks good. One comment about how to check if something is power of two, but its minor. If the tests succeeds then from my side good to go.

std/selector/multiplexer.go
aybehrouz: Yes, that's a good idea.

std/selector/multiplexer.go
```diff
@@ -43,7 +42,7 @@ func Map(api frontend.API, queryKey frontend.Variable,
 	// we don't need this check, but we added it to produce more informative errors
 	// and disallow len(keys) < len(values) which is supported by generateSelector.
 	if len(keys) != len(values) {
-		panic("The number of keys and values must be equal")
+		log.Panicf("The number of keys and values must be equal (%d != %d)", len(keys), len(values))
```
ivokub, line 45: Please revert. Imo log.Panicf writes to standard logger. gnark already has its "standard" logger and this makes configuring the output difficult. See https://github.com/ConsenSys/gnark/blob/master/logger/logger.go

std/selector/mux.go
```diff
@@ -14,7 +15,7 @@ import (
 // len(inputs) must be 2^len(selBits).
 func BinaryMux(api frontend.API, selBits, inputs []frontend.Variable) frontend.Variable {
 	if len(inputs) != 1<<len(selBits) {
-		panic("invalid input length for BinaryMux")
+		log.Panicf("invalid input length for BinaryMux (%d != 2^%d)", len(inputs), len(selBits))
```
ivokub, line 18: Same, use `panic`. If you want to format, then can use `panic(fmt.Sprintf(...))`

ivokub (changes requested):

std/selector/multiplexer.go
aybehrouz: Oh, sorry I didn't know that.

ivokub (approved):

ivokub:
Good to merge! Thanks

PR #637, opened 2023-04-11
feat: add generic binary comparison functions

aybehrouz:
This PR adds functions for comparing any two numbers given their binary representations. This implementation defines `2*n` R1CS constraints. 

Both `api.AssertIsLessOrEqual` and `api.Cmp` functions can be improved using this implementation. However it seems that binary return value is more useful than thet ternary return value of `api.Cmp`.

In general for comparing numbers `BoundedComparator` methods are more efficient (they generate `n` constraints), but these functions do not have the limitations of `BoundedComparator` methods.

### DIFF
--- a/std/cmp/binary.go
+++ b/std/cmp/binary.go
@@ -1,0 +1,36 @@
+package cmp
+
+import (
+	"github.com/consensys/gnark/frontend"
+)
+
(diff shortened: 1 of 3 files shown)


aybehrouz:
Included in PR #530

PR #654, opened 2023-04-20
perf: reduce mem allocs in scs frontend

gbotrel:
Example for a 1000 MiMC:
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkCompileMiMC-10     226517208     170724021     -24.63%

benchmark                   old allocs     new allocs     delta
BenchmarkCompileMiMC-10     4101305        2656976        -35.22%

benchmark                   old bytes     new bytes     delta
BenchmarkCompileMiMC-10     363802628     181183808     -50.20%
```


ivokub (approved):
LGTM

PR #655, opened 2023-04-21
fix: assert oddity of y coordinate from v instead of high bit

ivokub:
See https://github.com/ConsenSys/gnark-crypto/pull/385 for discussion.

I update go dependencies to match the commit in gnark-crypto which computes v correctly. However it isn't latest on develop as it breaks gnark.


yelhousni (approved):
LGTM

PR #659, opened 2023-04-22
perf: add a function for handling hints more efficiently

### LINKED ISSUE #653: The output of a hint with constant inputs should be a constant
When a hint is called on constant inputs the return value should be considered to be constant. This can greatly help in detecting constant wires and performing optimizations. Currently it seems that the output of a hint is always flagged as a non constant:
```go
bits, _ := api.Compiler().NewHint(NBits, 5, 17)
_, isConstant := api.Compiler().ConstantValue(bits[0])
print(isConstant)
```
output:
```
false
```

aybehrouz:
This PR provides a possible solution for issue #653.

### DIFF
--- a/constraint/bls12-377/coeff.go
+++ b/constraint/bls12-377/coeff.go
@@ -72,8 +72,8 @@ func (ct *CoeffTable) AddCoeff(coeff constraint.Element) uint32 {
 	return cID
 }
 
-func (ct *CoeffTable) MakeTerm(coeff *constraint.Element, variableID int) constraint.Term {
-	cID := ct.AddCoeff(*coeff)
+func (ct *CoeffTable) MakeTerm(coeff constraint.Element, variableID int) constraint.Term {
+	cID := ct.AddCoeff(coeff)
 	return constraint.Term{VID: uint32(variableID), CID: cID}
 }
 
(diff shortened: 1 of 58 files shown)


aybehrouz:
Tests are failing here, but on my local machine they look fine. By using `NewStateLessHint` function when `ToBinary` is called on a constant, it only creates one constraint, while I've not handled the constant path in the function at all.

aybehrouz:
@ivokub @gbotrel @Tabaie 

Hi,
What do you think about this PR, guys? It is the implementation of this idea: https://github.com/ConsenSys/gnark/issues/653#issuecomment-1517919584

gbotrel:
hey sorry we had a lot going on forgot about that one, thanks for the nudge.

* I prefer to see hints (all of them) as stateless functions. but we have several examples where circuit-developers used them as ... a way to introduce new wires and/or put a "hook" in the solver.
 --> so far the contract was "using a solver hint, you are introducing new wires and putting a placeholder for the solver to resolve these wires at solve time".  
 
 * not sure we do need the "hint inputs are constants, let's optimized it out" a lot. If it's isolated in one or two places (like ToBinary) might be simpler to target the issue there for now. it is more conservative but maybe a couple of extra constraints on some edge cases is better than introducing subtle behavior / doubling APIs for hints (note that for the stateless one; the compiler has no way to actually check that the function is indeed stateless). 
 
 will sync with @ivokub & @Tabaie 👍

aybehrouz:
I see. So, If you don't like the idea of a generic solution, For now I'd recommend handling constant inputs in `ToBinary` explicitly. There are some practical examples that shows `ToBinary` may be called with a constant input, while the programmer cannot notice it. (the call is done inside another function) The current Implementation of `ToBinary` creates around 250 constraints for a constant input, while no constraints are needed.

PR #661, opened 2023-04-26
perf(ecdsa): JoinScalarMulBase avoids 0 edge-cases

yelhousni:
In ECDSA verification (or ECRecover), we compute `[r/s]pubKey + [h/s]g` where `(r,s)` is the signature, `h` the hash of the message and `g` the canonical generator of the subgroup. Previously, we took into consideration the edge cases where either or both `r` and `h` can be 0 and also `pubKey=(0,0)`, but the [EYP](https://ethereum.github.io/yellowpaper/paper.pdf) rules out these:

- `s ≠ 0` and `r ≠ 0`

image

- `pubKey ≠ (0,0)` (because private key `≠ 0`)

image

The PR introduces the `JointScalarMulBase` method which performs `[s1]p1 + [s2]g` assuming that `s1` and `s2` are nonzero and `p1` not the infinity `(0,0)`. This saves ~17k constraints in ECDSA verifier gadget.

### DIFF
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -461,3 +461,73 @@ func (c *Curve[B, S]) ScalarMulBase(s *emulated.Element[S]) *AffinePoint[B] {
 
 	return res
 }
+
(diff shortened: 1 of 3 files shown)


ivokub (approved):
Looks good!

PR #668, opened 2023-04-30
Feat/plonk multicommit

Tabaie:
This PR enables a basic version of multicommits in Plonk. Its main use-case are in subprotocols with large prover messages, where MiMC hashing would not be practical:
```go
c0 := api.Commit(statementVars)
...    // compute prover response vector p0 based on c0
p0 = append(p0, c0)
c1 := api.Commit(p0...)
...
p1 = append(p1, c1)
c2 := api.Commit(p1)
```
As seen here, the previous challenge must be fed into subsequent `api.Commit` calls. Otherwise the commitments are normally considered independent from each other. In case there were independent protocols running in parallel, it is on the user to batch their commitments together for increased efficiency. It may be the objective of future efforts to perform this batching automatically.

The implementation is a naive extension of #586. For every commitment, a selector polynomial $q'_{C_i}$ is added to the setup (pre-computations, really) and a commitment $C_i$ to the proof. If `api.Commit` is called a super-logarithmic number of times, this will compromise the succinctness of Plonk so should such use-cases arise, more work on the theoretical side may be necessary to somehow batch verify the commitments without having to ship them all directly to the verifier.

A nice side effect of this PR is that now in a circuit with no commitments, everything actually (and not only in effect) defaults back to vanilla Plonk.


ThomasPiellard:
lgtm, but just for clarification, in your sample you keep updating the list of the previous committed value, but it's to only to illustrate how to derive successive challenges that depend on the previous state right ? If one wants to compute independent commitments it's possible to just call Commit() several times on different sets of wires right ?

backend/plonk/bls24-315/prove.go
```diff
@@ -55,7 +55,7 @@ type Proof struct {
 	H [3]kzg.Digest
 
 	// PI2, the BSB22 commitment
-	PI2 kzg.Digest
+	PI2 []kzg.Digest
```
ThomasPiellard, line 58: Should we find a better name for those commitments? What does the '2' means, is it to tell that a new public input is derived per commitment ?

Tabaie: Yes I'm in favor of a better name! PI2 came from how in the original paper the vector/polynomial holding the public values was called PI.

ThomasPiellard (approved):

PR #673, opened 2023-05-03
Adds ground work for custom blueprints. Fixes #660

### LINKED ISSUE #660: perf/bug: `logderivlookup.Table` memory blowup
Each lookup rely on calling a hint with nbInputs = len(table) + lookup params. 

So for a "large table", each lookup adds duplicate hint inputs in the constraint system, for no good reason.

Good candidate to use a "custom blueprint" and store the table only once 👍

gbotrel:
Fixes #660  .

High level idea of this PR; replace the lookup hint, with a "custom blueprint". The blueprint is functionally similar to a hint, in the sense that it does not add constraint, and indicates to the solver how to solve some wires at run time. 

However, the blueprint can have a state; in particular here, it encodes the lookup entries only once, such that additional query calls on the table do not incur a significant memory overhead. 

There are couple of perf adjustment to do (like caching the evaluation of the table in the solver and decrease overall memory allocs) but not urgent / will be in separate PR.

* The `test/Engine` also now implements `constraint.Solver` (partially) to support custom blueprints.
* This PR also kills the `WireIterator` and adds a simpler `WireWalker`; blueprint calls a callback with each wire that appears in an instruction.

## Benchmarks

 In particular, against `develop` for a circuit with 1000 lookups and a table with 130k entries, the size of the serialized constraint systems:

```
plonkish: -88.66%
r1cs: -93.41%
```

```
benchmark                                old ns/op      new ns/op      delta
BenchmarkCircuitTest/compile_scs-10      2936793375     803487125      -72.64%
BenchmarkCircuitTest/solve_scs-10        2854547292     665826250      -76.67%
BenchmarkCircuitTest/compile_r1cs-10     4438137000     1018271167     -77.06%
BenchmarkCircuitTest/solve_r1cs-10       3463089250     436040819      -87.41%

benchmark                                old allocs     new allocs     delta
BenchmarkCircuitTest/compile_scs-10      6787375        7583678        +11.73%
BenchmarkCircuitTest/solve_scs-10        2457896        1191280        -51.53%
BenchmarkCircuitTest/compile_r1cs-10     139058391      8515117        -93.88%
BenchmarkCircuitTest/solve_r1cs-10       4113213        1191315        -71.04%

benchmark                                old bytes       new bytes      delta
BenchmarkCircuitTest/compile_scs-10      10727280448     863376192      -91.95%
BenchmarkCircuitTest/solve_scs-10        1360056376      6508774600     +378.57%
BenchmarkCircuitTest/compile_r1cs-10     16213037632     2016655336     -87.56%
BenchmarkCircuitTest/solve_r1cs-10       1432770664      6445375594     +349.85%
```

### DIFF
--- a/constraint/bls12-377/coeff.go
+++ b/constraint/bls12-377/coeff.go
@@ -182,3 +182,11 @@ func (engine *field) String(a constraint.Element) string {
 	e := (*fr.Element)(a[:])
 	return e.String()
 }
+
+func (engine *field) Uint64(a constraint.Element) (uint64, bool) {
+	e := (*fr.Element)(a[:])
+	if !e.IsUint64() {
+		return 0, false
+	}
+	return e.Uint64(), true
+}
--- a/constraint/bls12-377/solver.go
+++ b/constraint/bls12-377/solver.go
@@ -376,7 +376,7 @@ func (s *solver) Read(calldata []uint32) (constraint.Element, int) {
 
 // processInstruction decodes the instruction and execute blueprint-defined logic.
 // an instruction can encode a hint, a custom constraint or a generic constraint.
-func (solver *solver) processInstruction(inst constraint.Instruction, scratch *scratch) error {
+func (solver *solver) processInstruction(inst constraint.PackedInstruction, scratch *scratch) error {
 	// fetch the blueprint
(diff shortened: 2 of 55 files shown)


ivokub (approved):
Some small stylistic suggestions, but otherwise looks good.

PR #676, opened 2023-05-05
fix race condition when compiling circuits in parallel

gbotrel:


Tabaie (approved):

ivokub (approved):
I'm not sure if atomic operations in test engine are necessary, but definitely doesn't hurt. Otherwise looks good.

gbotrel:
> I'm not sure if atomic operations in test engine are necessary, but definitely doesn't hurt. Otherwise looks good.

`go test -race` said it was useful :p

PR #696, opened 2023-05-18
refactor: emulation parameters

ivokub:
I was implementing ECDSA verification over P-384 and meanwhile saw a few things to improve:
* for Goldilocks field use the modulus from gnark-crypto instead of hardcoding. Makes easier to review when there is only a single definition.
* same for secp256k1 base and scalar fields. Previously weren't available in gnark-crypto but now is, so makes more sense to refer.
* moved emulation parameters into new package to decrease the number of implemented types and methods there. This hopefully makes it easier for the users to find the methods they are actually interested in. I enforce type aliases for existing parameters not to break any existing code.
* structured the type descriptions for types which implement FieldParams. Now everything is bullet pointed and modulus is in separate code-block for easier copying.
* added types which define 4 and 6 limbs. Then use type embedding for actual parameter implementations to minimise code overlap and possible errors.
* Added P-256 and P-384 parameters. This is preparatory work to add P-256 and P-384 parameters in emulated short weierstrass arithmetic for allowing verifying the P-256/P-384 ECDSA signatures in-chain.

### DIFF
--- a/std/math/emulated/params.go
+++ b/std/math/emulated/params.go
@@ -15,14 +15,6 @@ type FieldParams interface {
 	Modulus() *big.Int // returns modulus. Do not modify.
 }
 
-var (
-	qSecp256k1, rSecp256k1 *big.Int
-)
-func init() {
-	qSecp256k1, _ = new(big.Int).SetString("fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f", 16)
-	rSecp256k1, _ = new(big.Int).SetString("fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141", 16)
-}
-
 // Goldilocks provide type parametrization for emulated field on 1 limb of width 64bits
 // for modulus 0xffffffff00000001
 type Goldilocks struct{}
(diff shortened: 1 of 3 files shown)


std/math/emulated/params.go
```diff
@@ -3,108 +3,32 @@ package emulated
...
-// 64bits for modulus
-// 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001.
-// This is the scalar field of the BLS12-381 curve.
-type BLS12381Fr struct{}
-
-func (fp BLS12381Fr) NbLimbs() uint     { return 4 }
-func (fp BLS12381Fr) BitsPerLimb() uint { return 64 }
-func (fp BLS12381Fr) IsPrime() bool     { return true }
-func (fp BLS12381Fr) Modulus() *big.Int { return ecc.BLS12_381.ScalarField() }
+type Goldilocks = emparams.Goldilocks
```
yelhousni, line 27: Shouldn't be better to remove these types and call `emparams` directly when needed? That's fine for now to keep backwards compatibility but if we add new params and list them only in `emparams` and not here it would look weird IMO.

yelhousni (approved):
That looks neat 👍

std/math/emulated/params.go
ivokub: So, the main reason I separated the field parameters is to keep the `emulated` package documentation contained. Otherwise it looks like:
image

But now, when doing type aliases it is much better:
image

In that sense, I think it is OK if we add type aliases for all types defined `emparams`. I agree that better option is to remove this aliases here, but I'm just afraid that if we scrap the type aliases in `emulated` package, then a lot of external code will be broken and I would avoid that.

PR #697, opened 2023-05-18
feat: implement NIST P-256 and P-384 curves

ivokub:
These curves are fairly popular in practical applications (certificates, TLS, ID cards, signatures) and imo its better to have included so that users would implement correctly.

As these curves are supported by Go standard library, then do not need to add anything in gnark-crypto and can perform precomputation+tests using standard library only. So no more imports.

Is based on top of https://github.com/ConsenSys/gnark/pull/696, so it is better to have that reviewed/merged first.

Thanks to @SherLzp for initial contribution.

### DIFF
--- a/std/math/emulated/params.go
+++ b/std/math/emulated/params.go
@@ -15,14 +15,6 @@ type FieldParams interface {
 	Modulus() *big.Int // returns modulus. Do not modify.
 }
 
-var (
-	qSecp256k1, rSecp256k1 *big.Int
-)
-func init() {
-	qSecp256k1, _ = new(big.Int).SetString("fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f", 16)
-	rSecp256k1, _ = new(big.Int).SetString("fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141", 16)
-}
-
 // Goldilocks provide type parametrization for emulated field on 1 limb of width 64bits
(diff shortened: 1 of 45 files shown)


ivokub:
Oh, I didn't realise this is already done in https://github.com/ConsenSys/gnark/pull/610. Will see if I can attribute the contributor by merging the implementations.

std/math/emulated/params.go
```diff
@@ -3,108 +3,32 @@ package emulated
...
-// 64bits for modulus
-// 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001.
-// This is the scalar field of the BLS12-381 curve.
-type BLS12381Fr struct{}
-
-func (fp BLS12381Fr) NbLimbs() uint     { return 4 }
-func (fp BLS12381Fr) BitsPerLimb() uint { return 64 }
-func (fp BLS12381Fr) IsPrime() bool     { return true }
-func (fp BLS12381Fr) Modulus() *big.Int { return ecc.BLS12_381.ScalarField() }
+type Goldilocks = emparams.Goldilocks
```
yelhousni, line 27: hmm so either enforce P256 and P384 types here or remove others but I don't have a strong opinion

yelhousni (approved):
Definitely a great contribution to gnark std for users! LGTM 👍

std/math/emulated/params.go
ivokub: Included alias here.

PR #702, opened 2023-05-24
Groth16 Multicommits

Tabaie:
Equivalent of https://github.com/ConsenSys/gnark/pull/668 for Groth16.


backend/groth16/bls12-377/prove.go
```diff
@@ -37,9 +38,10 @@ import (
 // with a valid statement and a VerifyingKey
 // Notation follows Figure 4. in DIZK paper https://eprint.iacr.org/2018/691.pdf
 type Proof struct {
-	Ar, Krs                   curve.G1Affine
-	Bs                        curve.G2Affine
-	Commitment, CommitmentPok curve.G1Affine
+	Ar, Krs       curve.G1Affine
+	Bs            curve.G2Affine
+	CommitmentPok curve.G1Affine
```
gbotrel, line 43: can you add a comment here to point to the right resource or TLDR; what is CommitmentPok and Commitments?

backend/groth16/bls12-377/prove.go
```diff
@@ -61,35 +63,32 @@ func Prove(r1cs *cs.R1CS, pk *ProvingKey, fullWitness witness.Witness, opts ...b
...
-			if len(in) != r1cs.CommitmentInfo[0].NbCommitted() { // TODO: Remove
-				return fmt.Errorf("unexpected number of committed variables")
-			}
-			values := make([]fr.Element, r1cs.CommitmentInfo[0].NbPrivateCommitted)
-			nbPublicCommitted := len(in) - len(values)
-			inPrivate := in[nbPublicCommitted:]
-			for i, inI := range inPrivate {
-				values[i].SetBigInt(inI)
-			}
+	commitmentSubIndexes := r1cs.CommitmentInfo.CommitmentIndexesInCommittedLists()
```
gbotrel, line 70: maybe `CommitmentIndexesInCommittedLists` -> `Indexes` ?

backend/groth16/bls12-377/prove.go
```diff
@@ -61,35 +63,32 @@ func Prove(r1cs *cs.R1CS, pk *ProvingKey, fullWitness witness.Witness, opts ...b
...
+	privateCommittedValues := make([][]fr.Element, len(r1cs.CommitmentInfo))
+	for i := range r1cs.CommitmentInfo {
+		solverOpts = append(solverOpts, solver.OverrideHint(r1cs.CommitmentInfo[i].HintID, func(i int) solver.Hint {
+			return func(_ *big.Int, in []*big.Int, out []*big.Int) error {
+				privateCommittedValues[i] = make([]fr.Element, r1cs.CommitmentInfo[i].NbPrivateCommitted-len(commitmentSubIndexes[i]))
+				hashed, committed := internal.DivideByThresholdOrList(r1cs.CommitmentInfo[i].NbPublicCommitted(), commitmentSubIndexes[i], in)
+				for j, inJ := range committed {
+					privateCommittedValues[i][j].SetBigInt(inJ) // TODO @Tabaie Perf If this takes significant time can read values off the witness vector instead
```
gbotrel, line 78: `SetBigInt` is probably okay overhead vs `Set` (since `in` is already allocating `big.Int`). 1 extra mul per value.

However, I think `DivideByThresholdOrList` should work in-place, and re-slice the `in` input slice into 2. Avoid allocating too many slices of pointers to `big.Int` will matter.

backend/groth16/bls12-377/prove.go
```diff
@@ -102,6 +101,15 @@ func Prove(r1cs *cs.R1CS, pk *ProvingKey, fullWitness witness.Witness, opts ...b
 
 	start := time.Now()
 
+	commitmentsSerialized := make([]byte, fr.Bytes*len(r1cs.CommitmentInfo))
```
gbotrel, line 104: if we commit to many variables, that could slow the prover. It's also running in the "main go routine" of the prover. Any way to avoid the `[]byte` allocation?

constraint/commitment.go
```diff
@@ -64,10 +67,67 @@ func (i *Commitment) PublicCommitted() []int {
 	return i.Committed[:i.NbPublicCommitted()]
 }
 
-func CommitmentIndexes(commitments []Commitment) []uint64 {
-	res := make([]uint64, len(commitments))
-	for i := range res {
-		res[i] = uint64(commitments[i].CommitmentIndex)
+func (c Commitments) CommitmentWireIndexes() []int {
```
gbotrel, line 70: maybe evaluate if instead of computing all the index (wires, commited etc) during proving time, it would not be worth it to store these slices in the `CommitmentInfo` data struct? Particularly to avoid allocations in the prover.

backend/groth16/bls12-377/prove.go
```diff
@@ -288,27 +293,57 @@ func Prove(r1cs *cs.R1CS, pk *ProvingKey, fullWitness witness.Witness, opts ...b
 	return proof, nil
 }
 
+// if len(indexes) == 0, returns slice, nil
+// else, returns a slice with the indexes and another without
+// this assumes the indexes are sorted, without repetition and that len(slice) > len(indexes)
+func split(slice []*big.Int, indexes []int) (with, without []*big.Int) {
```
gbotrel, line 299: deadcode, unused?

backend/groth16/bls12-377/setup.go
```diff
@@ -94,17 +94,24 @@ func Setup(r1cs *cs.R1CS, pk *ProvingKey, vk *VerifyingKey) error {
...
+		r1cs.CommitmentInfo.Interleave(r1cs.GetNbPublicVariables())
+
+	// a commitment is itself defined by a hint so the prover considers it private
+	// but the verifier will need to inject the value itself so on the groth16
+	// level it must be considered public
+	nbPublicWires := r1cs.GetNbPublicVariables() + len(r1cs.CommitmentInfo)
+	nbPrivateWires := r1cs.GetNbSecretVariables() + r1cs.NbInternalVariables - nbPrivateCommittedWires - len(r1cs.CommitmentInfo)
+
+	// we will need to iterate through the private-to-public wires in order
+	/*privToPub := make([], nbPrivateCommittedWires+len(r1cs.CommitmentInfo))
```
gbotrel, line 107: deadcode

backend/groth16/bls12-377/setup.go
```diff
@@ -152,31 +162,42 @@ func Setup(r1cs *cs.R1CS, pk *ProvingKey, vk *VerifyingKey) error {
...
-	var privateCommitted []int
-	if r1cs.GetNbCommitments() != 0 {
-		privateCommitted = r1cs.CommitmentInfo[0].PrivateCommitted()
-	}
+	vI := 0
+	cI := make([]int, len(r1cs.CommitmentInfo))
+	nbPrivateCommittedSeen := 0 // = ∑ᵢ cI[i]
+	nbCommitmentsSeen := 0
 
 	for i := range A {
```
gbotrel, line 170: add a comment to say what the following lines of code are doing

backend/groth16/bls12-377/verify.go
```diff
@@ -62,21 +61,28 @@ func Verify(proof *Proof, vk *VerifyingKey, publicWitness fr.Vector) error {
...
-
-		if err := vk.CommitmentKey.Verify(proof.Commitment, proof.CommitmentPok); err != nil {
-			return err
-		}
-
-		publicCommitted := make([]*big.Int, len(vk.PublicCommitted))
-		for i := range publicCommitted {
+	commitmentsSerialized := make([]byte, len(vk.PublicCommitted)*fr.Bytes)
+	for i := range vk.PublicCommitted {
+		publicCommitted := make([]*big.Int, len(vk.PublicCommitted[i]))
```
gbotrel, line 66: seems the flow is a bit greedy in memory here; each iteration allocate a new slice, convert stuff into big.Int, call `solveCommitmentWire` to convert the big int to `bytes` ... 

Minimizing allocations (including the `commitmentsSerialized` bits) may help; again if we have one or two committed variables we won't see it, but for larger instances.. it will impact perf on the verifier significantly.

backend/groth16/bls12-377/verify.go
```diff
@@ -87,8 +93,8 @@ func Verify(proof *Proof, vk *VerifyingKey, publicWitness fr.Vector) error {
 	}
 	kSum.AddMixed(&vk.G1.K[0])
 
-	if vk.HasCommitment {
-		kSum.AddMixed(&proof.Commitment)
+	for i := range proof.Commitments {
```
gbotrel, line 96: same, my understanding is we probably won't go over comuple of commitments, but if we do, this place might be a good candidate for a batchAddition

backend/groth16/internal/utils.go
```diff
@@ -0,0 +1,23 @@
+package internal
+
+import "math/big"
+
+// DivideByThresholdOrList divides x into two sub-lists. list must be sorted, non-repeating and contain no value less than indexThreshold
+func DivideByThresholdOrList(indexThreshold int, list []int, x []*big.Int) (ltOrInList, gtAndNotInList []*big.Int) {
```
gbotrel, line 6: do that in place?

frontend/cs/r1cs/api.go
```diff
@@ -719,11 +716,27 @@ func (builder *builder) Commit(v ...frontend.Variable) (frontend.Variable, error
 			// it's the same variable ID, do nothing
 			continue
 		} else {
-			// append, it's a new variable ID
-			committed = append(committed, t.VID)
 			if t.VID < builder.cs.GetNbPublicVariables() {
 				nbPublicCommitted++
+			} else {
+				// Cannot commit to a secret variable that has already been committed to
```
gbotrel, line 722: out of curiosity; why can't we just continue here and ignore that? Will comitting to a commitment help in anyway?

internal/utils/search.go
```diff
@@ -0,0 +1,18 @@
+package utils
+
+// BinarySearch attempts to find the target in x. If not found, returns false and the index where the target would be inserted.
```
gbotrel, line 3: why not use go standard std `Find` method?

test/end_to_end.go
```diff
@@ -0,0 +1,135 @@
...
+	t.Run("plonk-e2e", func(t *testing.T) {
+		testPlonk(t, assignment)
+	})
+
+	t.Run("groth16-e2e", func(t *testing.T) {
+		testGroth16(t, assignment)
+	})
+}
+
+func hollow(c frontend.Circuit) frontend.Circuit {
```
gbotrel, line 112: can you document a bit more this file / the functions?

gbotrel (changes requested):

frontend/cs/r1cs/api.go
Tabaie: Because of how the result of an `api.Commit` must be usable as a fiat-shamir challenge, it must have a "one-way" dependence on every variable input to this function. However, the way the commitment scheme is verified is Groth16 necessitates that every private variable can be used in one commitment at most. So if it were needed in multiple commitments, the next best thing is to commit to its commitment (commitments to commitments are easy, they're just conventional hashes.)

backend/groth16/bls12-377/prove.go
Tabaie: I think that wouldn't be expressive enough. This is the list of commitments that are committed to, and they are referred to not even by their wire index, but by their index in the list of committed values in that commitment that commits to them. 🤪

backend/groth16/bls12-377/prove.go
Tabaie: This is not too bad imo. Even though the number of committed variables may be huge, so far the number of _commitments_ themselves have been pretty small. I suggest we keep an eye on it until we find a use-case where there is at least $O(\log n)$ or so commitments.

backend/groth16/internal/utils.go
Tabaie: Replaced, but it still allocates a slice of size `len(list)`, which in this application is small.

backend/groth16/bls12-377/setup.go
```diff
@@ -94,17 +95,19 @@ func Setup(r1cs *cs.R1CS, pk *ProvingKey, vk *VerifyingKey) error {
...
 
-	if r1cs.GetNbCommitments() != 0 { // the commitment itself is defined by a hint so the prover considers it private
-		nbPublicWires++  // but the verifier will need to inject the value itself so on the groth16
-		nbPrivateWires-- // level it must be considered public
-	}
+	commitmentInfo := r1cs.CommitmentInfo.(constraint.Groth16Commitments)
+	commitmentWires := commitmentInfo.CommitmentIndexes()
+	privateCommitted := commitmentInfo.GetPrivateCommitted()
+	nbPrivateCommittedWires := internal.NbElements(privateCommitted)
+	//nbPrivateCommittedWires, privateCommitted, publicAndCommitmentCommitted :=
```
gbotrel, line 103: deadcode

gbotrel (approved):

PR #704, opened 2023-05-25
internal: util package to parse and verify AZTEC ignition ceremony

gbotrel:
To discuss, added it in gnark/internal for convenience, but maybe we should put that in a separate independent repo. 

Advantage of putting it here; the next step, having an API to pull a trusted setup from a web location would be more "discoverable" by gnark users. **Such a "trusted setup" would be usable for example for PlonK circuits using bn254**.

```
// Package ignition is a package that provides helper functions to download, parse
 // and validate the AZTEC Ignition Ceremony data.
 // The specs are described here:
 //
 //	https://github.com/AztecProtocol/ignition-verification/blob/c333ec4775045139f86732abfbbd65728404ab7f/Transcript_spec.md
 //
 // The verification logic follows
 //
 //	https://github.com/AztecProtocol/ignition-verification
 ```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ go.work
 go.work.sum
 
 examples/gbotrel/**
+
+# ignore cached SRS
+internal/ignition/cmd/data/**
--- a/internal/ignition/cmd/main.go
+++ b/internal/ignition/cmd/main.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/consensys/gnark/internal/ignition"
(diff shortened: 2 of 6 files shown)


gbotrel:
mmhh maybe a separate repo + pointing to this in gnark-crypto kzg package to be able to build kzg.ProvingKey & kzg.VerifyingKey from this package if needed would make more sense. 
wdyt @ThomasPiellard / @Tabaie ?

Tabaie:
I'd personally rather stay away from network calls in gnark-crypto.

PR #706, opened 2023-05-29
perf(2-chain/varScalarMul): use DoubleAndAdd to reduce #constraints

yelhousni:
While looking into a question about emulated ScalarMul, I noticed we can further merge some `Double`+ `Add` into a `DoubleAndAdd` to reduce the number of constraints. Hence the tiny PR.

- Old
```
BenchmarkVarScalarMulG1
    g1_test.go:460: groth16 1548
BenchmarkVarScalarMulG1/plonk
    g1_test.go:471: plonk 3861
BenchmarkVarScalarMulG2
    g2_test.go:496: groth16 4250
BenchmarkVarScalarMulG2/plonk
    g2_test.go:507: plonk 12604
```
- New:
```
BenchmarkVarScalarMulG1
    g1_test.go:460: groth16 1544
BenchmarkVarScalarMulG1/plonk
    g1_test.go:471: plonk 3858
BenchmarkVarScalarMulG2
    g2_test.go:496: groth16 4240
BenchmarkVarScalarMulG2/plonk
    g2_test.go:507: plonk 12579
```

PR #707, opened 2023-05-29
perf(2-chain/varScalarMul): DoubleAndAdd to reduce #constraints BLS24

yelhousni:
Forgot about BLS24. Same as #706


ivokub (approved):
LGTM

PR #708, opened 2023-05-30
Feat: fixed-argument emulated pairing

yelhousni:
This PR adds circuits for emulated pairing `e(P,G₂)` and `e(P,G₂)*e(T,Q)` where `G₂` is a fixed point (the canonical generator of the G2 subgroup). The idea is to precompute all the lines and avoid to do EC arithmetic in-circuit. This scenario happens for e.g. BLS signature (minimal-signature-size variant) and KZG.

TODO for later PRs:
- [x] pre-compute product of lines when bit is 1 or -1
- [ ] emulated BLS signature gadget
- [ ] emulated KZG gadget
- [ ] allow fixed arguments other than `G₂` (KZG has 2 pairings, one with fixed `G₂` and one with fixed `[α]G₂` from the SRS)

---
- [x] same for native pairing packages 
  - [x] BLS12-377 
  - [x] BLS24-315


ivokub (approved):
I made a few changes:
* for native I just made the computated lines private.
* for emulated I made the computed lines private and also initalise lazily as there is small overhead. It is not significant, but in the future when we maybe want to provide fixed point as a paremeter, then it is bigger.

Otherwise looks good as always!

ivokub (approved):
Your modifications also approved.

PR #712, opened 2023-06-01
fix: randomize fake commitments

Tabaie:
This PR replaces the fake commitment placeholder with its idealization, true randomness.

### DIFF
--- a/frontend/cs/commitment.go
+++ b/frontend/cs/commitment.go
@@ -2,7 +2,6 @@ package cs
 
 import (
 	"crypto/rand"
-	"crypto/sha256"
 	"fmt"
 	"github.com/consensys/gnark/constraint/solver"
 	"github.com/consensys/gnark/debug"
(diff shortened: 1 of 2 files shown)


gbotrel:
yeah I suspect it's fine, doesn't translate the fact that the commitment should depend on the variable you committed to, but I guess for the placeholder part, that doesn't endanger the soundness of the real protocol. thoughts @ivokub ?

ivokub:
> yeah I suspect it's fine, doesn't translate the fact that the commitment should depend on the variable you committed to, but I guess for the placeholder part, that doesn't endanger the soundness of the real protocol. thoughts @ivokub ?

Should be fine to use true randomness. I previously had digest of the inputs to have make debugging easier. But in practice the commitment does not depend only on the inputs but also on the commitment key (== proving key), so it is not fully deterministic.

Tabaie:
> yeah I suspect it's fine, doesn't translate the fact that the commitment should depend on the variable you committed to, but I guess for the placeholder part, that doesn't endanger the soundness of the real protocol. thoughts @ivokub ?

In fact @ivokub and @ThomasPiellard were recently suggesting to change the name from `api.Commit` to something else emphasizing randomness, in light of which I think this PR would make more sense. The point of `api.Commit(X,Y)` at least as originally envisioned by @AlexandreBelling is "give me a fiat-shamir challenge to a prover who has seen X,Y".

ivokub:
> > yeah I suspect it's fine, doesn't translate the fact that the commitment should depend on the variable you committed to, but I guess for the placeholder part, that doesn't endanger the soundness of the real protocol. thoughts @ivokub ?
> 
> In fact @ivokub and @ThomasPiellard were recently suggesting to change the name from `api.Commit` to something else emphasizing randomness, in light of which I think this PR would make more sense. The point of `api.Commit(X,Y)` at least as originally envisioned by @AlexandreBelling is "give me a fiat-shamir challenge to a prover who has seen X,Y".

Yeah, the current name is not intuitive but I'm starting to reconsider :) I have on top of my head an idea on how to use the commitment for its actual purpose -- carrying some state between different circuits. The main pain-point being that when we use lookup tables then cannot amortise the table construction between several circuits when batching.

But this in principle doesn't prevent renaming now. If I formalise the ideas then I can propose something separately.

PR #719, opened 2023-06-08
Feat/aes128

jplaui:
Please find my initial aes128 and aes128 gcm implementations in this pull request.
I have implemented the naive lookup of s-box values. possible s-box lookup optimizations can be found here: https://github.com/akosba/jsnark/tree/master/JsnarkCircuitBuilder/src/examples/gadgets/blockciphers/sbox


ivokub:
> Please find my initial aes128 and aes128 gcm implementations in this pull request. I have implemented the naive lookup of s-box values. possible s-box lookup optimizations can be found here: https://github.com/akosba/jsnark/tree/master/JsnarkCircuitBuilder/src/examples/gadgets/blockciphers/sbox

Thanks for the contribution. Unfortunately right now AES in not on gnark roadmap and it would take quite some time to review the PR is correct (and most importantly, sound). I think AES is definitely worth it and allows interesting applications. Lets keep the PR open until we have more time to get back to it.

PR #726, opened 2023-06-16
Perf: Add-only emulated scalar multiplication

yelhousni:
I was trying to have a `ScalarMul()` circuit with only additions and no doublings. Ultimately, I could make the circuit work with less constraints than the previous right-to-left double-and-add circuit. So I looked in the literature to check if it was previously known and I found that Marc Joye *of course* did it in 2007 😅
So this PR corresponds actually to Alg.2 of [[CHES:Joye07]](https://www.iacr.org/archive/ches2007/47270135/47270135.pdf) with some tweaks to make it work efficiently in-circuit: 
- select-logic with an additional register;
- isolate first iteration and use [[ELM02]](https://eprint.iacr.org/2002/112) `triple()` method;
- isolate last iteration and discard `R1` and `R2` computations.

Now when we use this new method in precompiles it saves:
- `ECRECOVER`: 
  - 17 609 r1cs
  - 111 590 scs
- `ECMUL`: 
  - 22 997 r1cs
  - 124 945 scs

P.S.: We can ditch the old `ScalarMul()` and replace it with the new `ScalarMulAddOnly()` once the PR is reviewed.

### DIFF
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -531,3 +531,43 @@ func (c *Curve[B, S]) JointScalarMulBase(p *AffinePoint[B], s2, s1 *emulated.Ele
 
 	return c.add(res1, res2)
 }
+
+// ScalarMulAddOnly computes s * p and returns it. It doesn't modify p nor s.
+// This function doesn't check that the p is on the curve. See AssertIsOnCurve.
+//
+// ⚠️  p cannot be (0,0) and s cannot be 0.
+//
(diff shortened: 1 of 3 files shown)


ivokub:
> The table-free windowubg in JointScalarMulBase was difficult to follow but seems correct (two selects in a row thing...).
> 
> I'm not sure about name `ScalarMulAddOnly`, but really cannot suggest anything more succincit (maybe `ScalarMulWithAdds`, but it still is very long).

Oh, we can ditch old ScalarMul and replace with it. Yup, makes also sense. But leave it in comments just in case, maybe is useful in the future, so is good to have close :)

PR #732, opened 2023-06-21
feat/ci: make CI run solidity verification with integartion test. Added test.WithSolidity option. 

gbotrel:
- feat: plonk + sol rebased on latest develop branch
- feat: kept only verifier template
- refactor: plonk solidity generate match expected gnark interface
- tmp: change Library to Contract and add MarshalSolidity on proof
- feat: plonk sodliity change visibility of some funcs
- feat: add SPDIX apache 2 identifier
- fix: move new solidity code into tmpl generation
- fix: use big int string in solidity generation for plonk to avoid negative numbers
- feat: add WithSolidity option in assert helper
- ci: attempt to make gnark-solidity-checker run on CI


ivokub:
Suggested edit:
```diff
diff --git a/test/options.go b/test/options.go
index 0879a365..65aa8a00 100644
--- a/test/options.go
+++ b/test/options.go
@@ -107,7 +107,7 @@ func WithCompileOpts(compileOpts ...frontend.CompileOption) TestingOption {
 
 // WithSolidity is a testing option which enables solidity tests in assertions.
 // If the build tag "solccheck" is not set, this option is ignored.
-// When the tag is set; his requires gnark-solidity-checker to be installed, which in turns
+// When the tag is set; this requires gnark-solidity-checker to be installed, which in turns
 // requires solc and abigen to be reachable in the PATH.
 //
 // See https://github.com/ConsenSys/gnark-solidity-checker for more details.

```

.github/workflows/pr.yml
```diff
@@ -70,10 +70,16 @@ jobs:
...
+        go install golang.org/x/tools/cmd/goimports@latest && go install github.com/klauspost/asmfmt/cmd/asmfmt@latest
+        go install github.com/ethereum/go-ethereum/cmd/abigen@v1.12.0
+        go install github.com/consensys/gnark-solidity-checker@latest
+        sudo add-apt-repository ppa:ethereum/ethereum
+        sudo apt-get update
+        sudo apt-get install solc
     - name: Test
       run: |
-        go test -v -short -timeout=30m ./...
+        go test -v -short -tags=solccheck -timeout=30m ./...
```
ivokub, line 82: The Solitify check seems quite heavy - maybe it would be worth to create separate workflow and call less frequently? I think this could slow down the development speed quite a lot, considering Solidity tests are run for the whole integration test suite.

ivokub: Hmm, I guess we can include solidity tests in CI for now and then later figure out how to make PR tests lighter (I think it would be good to get faster feedback when developing).

test/assert_solidity.go
```diff
@@ -0,0 +1,95 @@
...
+// solidityVerification checks that the exported solidity contract can verify the proof
+// and that the proof is valid.
+// It uses gnark-solidity-checker see test.WithSolidity option.
+func (assert *Assert) solidityVerification(b backend.ID, vk interface {
+	NbPublicWitness() int
+	ExportSolidity(io.Writer) error
+},
+	proof any,
+	validPublicWitness witness.Witness) {
+	if vk.NbPublicWitness() == 0 || !solcCheck {
```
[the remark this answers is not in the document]
gbotrel: at least for Groth16 the template to generate the solidity really doesn't like that :p 

just checked with PlonK, it seems test pass so will update !

PR #736, opened 2023-06-29
feat: [PLONK_AUDIT_4-11] fixes #735

### LINKED ISSUE #735: [PLONK_AUDIT_4-11] missing proof length check
Require that proof length is `0x340+vk_nb_commitments_commit_api*0x60`

ThomasPiellard:
Added missing proof length check, fixes #735

### DIFF
--- a/backend/plonk/bn254/solidity.go
+++ b/backend/plonk/bn254/solidity.go
@@ -277,11 +277,24 @@ contract PlonkVerifier {
       alpha := mod(alpha, r_mod)
       zeta := mod(zeta, r_mod)
 
+      // Derive gamma as Sha256(<transcript>)
+      // where transcript is the concatenation (in this order) of:
+      // * the word "gamma" in ascii, equal to [0x67,0x61,0x6d, 0x6d, 0x61] and encoded as a uint256.
+      // * the commitments to the permutation polynomials S1, S2, S3, where we concatenate the coordinates of those points
+      // * the commitments of Ql, Qr, Qm, Qo, Qk
+      // * the public inputs
+      // * the commitments of the wires related to the custom gates (commitments_wires_commit_api)
+      // * commitments to L, R, O (proof_<l,r,o>_com_<x,y>)
+      // The data described above is written starting at mPtr. "gamma" lies on 5 bytes,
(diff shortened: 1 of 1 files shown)


gbotrel:
```
verifierContract.Verify undefined (type *PlonkVerifier has no field or method Verify)
``` 
---> `Verify` shouldn't be marked `internal`

backend/plonk/bn254/solidity.go
```diff
@@ -422,7 +421,6 @@ contract PlonkVerifier {
   }
 
   function compute_pi(
-        bytes memory proof,
```
DavePearce, line 425: I don't really understand dropping `proof` parameter from `compute_pi()`, since `load_wire_commitments_commit_api()` accepts the proof and uses it.  I guess you're planning further changes?

backend/plonk/bn254/solidity.go
```diff
@@ -422,7 +421,6 @@ contract PlonkVerifier {
   }
 
   function compute_pi(
```
DavePearce, line 423: You've lost some of the nice comments you wrote for the audit (e.g. for `compute_pi()` but also in other places).  Maybe you should copy these over if this is the master copy of the contract?

DavePearce (approved):
Looks good. Added a few comments.

gbotrel (approved):

backend/plonk/bn254/solidity.go
ThomasPiellard: Ah nice catch, the last test that I generated was without the use of api.Commit() and the proof is not needed in that case I forgot about that (I followed the compiler that told me the parameter was unused). I'll add a condition in the template to not take the proof when commit is not used

backend/plonk/bn254/solidity.go
ThomasPiellard: Ah nice catch, the last test that I generated was without the use of api.Commit() and the proof is not needed in that case I forgot about that (I followed the compiler that told me the parameter was unused). I'll add a condition in the template to not take the proof when commit is not used, but it's a mistake on my end

backend/plonk/bn254/solidity.go
ThomasPiellard: I re added the comments

PR #739, opened 2023-06-29
feat: [PLONK_AUDIT_4-9] fixes 738

### LINKED ISSUE #738: [PLONK_AUDIT_4-9] missing inputs size check
Inputs should be bounded above by r_mod

ThomasPiellard:
Added checks on the inputs, to ensure they are below r_mod. 

fixes #738


backend/plonk/bn254/solidity.go
```diff
@@ -540,18 +538,37 @@ contract PlonkVerifier {
       
       return pi;
     }
+  
+  function check_inputs_size(uint256[] memory public_inputs)
```
DavePearce, line 542: Looks good to me

DavePearce (commented):
One thought I had is that there is no length check for `public_inputs`.  Is that a concern?  Is there any minimal or maximal length for `public_inputs`?  For example, a length of `0`?  I guess that the verifier is designed to operate on an arbitrary number of public inputs, so perhaps its fine without a check.

gbotrel (approved):

backend/plonk/bn254/solidity.go
ThomasPiellard: @DavePearce no there is no bound on the number of public inputs, given a proof there is one exact set of public inputs that makes the tuple (proof, public_inputs) valid and vice versa

PR #740, opened 2023-06-29
Feat/gkr testing

Tabaie:
User should be able to use GKR in testing mode, without a full-on prover involved.


constraint/bn254/system.go
```diff
@@ -377,3 +384,14 @@ func getTagSet() cbor.TagSet {
 func (s *system) AddGkr(gkr constraint.GkrInfo) error {
 	return s.System.AddGkr(gkr)
 }
+
+func (s *system) Equal(other constraint.ConstraintSystem) bool {
+	if o, ok := other.(*system); !ok {
+		return false
+	} else {
+		systemDiff := constraint.SystemDiff(s.System, o.System)
```
AlexandreBelling, line 392: I don't see any usage of `SystemDiff` aside the present one. Is it for debugging?

PR #742, opened 2023-06-30
feat: [PLONK_AUDIT_4-4] fixes #741

### LINKED ISSUE #741: [PLONK_AUDIT_4-4] missing verifying pairing check
pairing check is missing, only the success of staticcall is made so far

ThomasPiellard:
added pairing check result: fixes #741

### DIFF
--- a/backend/plonk/bn254/solidity.go
+++ b/backend/plonk/bn254/solidity.go
@@ -741,8 +741,8 @@ contract PlonkVerifier {
         mstore(add(mPtr, 0x140), g2_srs_1_y_0)
         mstore(add(mPtr, 0x160), g2_srs_1_y_1)
         let l_success := staticcall(sub(gas(), 2000),8,mPtr,0x180,0x00,0x20)
-        // l_success := true
-        mstore(add(state, state_success), and(l_success,mload(add(state, state_success))))
+        let res_pairing := mload(0x00)
+        mstore(add(state, state_success), and(l_success,eq(res_pairing,0x1)))
       }
 
       // Fold the opening proofs at ζ:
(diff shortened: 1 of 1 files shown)


backend/plonk/bn254/solidity.go
```diff
@@ -741,8 +741,8 @@ contract PlonkVerifier {
         mstore(add(mPtr, 0x140), g2_srs_1_y_0)
         mstore(add(mPtr, 0x160), g2_srs_1_y_1)
         let l_success := staticcall(sub(gas(), 2000),8,mPtr,0x180,0x00,0x20)
-        // l_success := true
-        mstore(add(state, state_success), and(l_success,mload(add(state, state_success))))
+        let res_pairing := mload(0x00)
+        mstore(add(state, state_success), and(l_success,eq(res_pairing,0x1)))
```
DavePearce, line 745: Hmmm, this doesn't seem equivalent.  If `state_success` is `false` at this point (e.g. because say the quotient check failed) then you're ignoring that information?  Seems to me like you need to `and` against `mload(add(state, state_success)` and `res_pairing`.  

Also, do you really need the `eq(...,0x1)`.  The yellow paper seems to me to say that the SNARKV precompiled contract either returns `0` or `1`.

DavePearce (commented):
See my comment about a missing `and` against `mload(add(state, state_success)`.

backend/plonk/bn254/solidity.go
ThomasPiellard: yep you're right, somehow I manage to make mistakes at each commits^^

DavePearce: All good, that's what the process is for.

DavePearce (approved):
Great, looks good to me now!

Also, you might want to think about adding some tests at some point (because the CI pipeline didn't fail with an error for an incorrect pairing check)

ivokub (approved):

PR #747, opened 2023-07-02
fix: plonk scs serialization issues

gbotrel:
1. 
```
panic: error reading scs : cbor: exceeded max number of elements 134217728 for CBOR array
```
--> seems now we can do `2147483647` as max size, easy fix, lucky for us :) 

2. 
```
panic: cbor: cannot unmarshal map into Go struct field cs.system.Blueprints of type constraint.Blueprint
```

--> suspect this comes from `BlueprintLookupHint` 
We need to figure out a better way to serialize blueprints, this fix should do for now, but we need to re-serialize the circuits (cc @AlexandreBelling )


ivokub:
To confirm, looks good to me. I guess it is not so bad that lookup blueprint is in constraint package. Maybe we want to put the blueprints into separate package in the future so that constraint package wouldn't be too big.

PR #754, opened 2023-07-04
feat: status of staticcalls are checked, fixes #753

### LINKED ISSUE #753: [PLONK_AUDIT_4-5] missing status check for staticcall
Currently the staticcalls are not checked (the valued is poped from the stack), that could result in computing with some tampered data if a staticcall runs out of gas for instance.

ThomasPiellard:
I added checks on each of the staticcalls instructions following this doc: 
- https://blog.soliditylang.org/2021/04/21/custom-errors/.

The contract now reverts if for instance not enough gas is supplied in the static calls.

The other solution was to update the `state_success` in the memory but it was messy, and there was no error message...

fixes #753

### DIFF
--- a/backend/plonk/bn254/solidity.go
+++ b/backend/plonk/bn254/solidity.go
@@ -243,10 +243,10 @@ contract PlonkVerifier {
   uint256 constant state_success = 0x240;
   uint256 constant state_check_var = 0x260; // /!\ this slot is used for debugging only
 
-
   uint256 constant state_last_mem = 0x280;
 
-  event PrintUint256(uint256 a);
+  // -------- errors
+  uint256 constant error_string_id = 0x08c379a000000000000000000000000000000000000000000000000000000000; // selector for function Error(string)
 
   function derive_gamma_beta_alpha_zeta(bytes memory proof, uint256[] memory public_inputs)
   internal view returns(uint256, uint256, uint256, uint256) {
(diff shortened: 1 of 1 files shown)


ivokub (approved):
LGTM

backend/plonk/bn254/solidity.go
```diff
@@ -332,27 +341,31 @@ contract PlonkVerifier {
...
         mstore(add(_mPtr, 0x60), mload(add(aproof, proof_r_com_y)))
         mstore(add(_mPtr, 0x80), mload(add(aproof, proof_o_com_x)))
         mstore(add(_mPtr, 0xa0), mload(add(aproof, proof_o_com_y)))
-        // pop(staticcall(sub(gas(), 2000), 0x2, add(mPtr, 0x1b), 0x365, mPtr, 0x20)) //0x1b -> 000.."gamma"
 
         let size := add(0x2c5, mul(mload(pub_inputs), 0x20)) // 0x2c5 = 22*32+5
         size := add(size, mul(vk_nb_commitments_commit_api, 0x40))
-        pop(staticcall(sub(gas(), 2000), 0x2, add(mPtr, 0x1b), size, mPtr, 0x20)) //0x1b -> 000.."gamma"
+        let success := staticcall(sub(gas(), 2000), 0x2, add(mPtr, 0x1b), size, mPtr, 0x20) //0x1b -> 000.."gamma"
+        if eq(success, 0) {
```
DavePearce, line 355: You could use `iszero()` here, so it would be:
```Solidity
if iszero(success) { ... }
```
But, I don't think it really matters.

DavePearce (commented):
Looks good to me

PR #756, opened 2023-07-04
fix: loop counter corrected fixes #755

### LINKED ISSUE #755: [PLONK_AUDIT_4-13] loading arbitrary data when loading wire commitments (loop too big)
in `load_wire_commitments_commit_api`, the loop counter is twice the required size (in the loop we load both the x and y coordinates so no need to set the loop counter as `2*vk_nb_commitments_commit_api`

ThomasPiellard:
in `load_wire_commitments_commit_api`, replaced the loop counter by `vk_nb_commitments_commit_api` (instead of `2*vk_nb_commitments_commit_api` )

fixes #755

### DIFF
--- a/backend/plonk/bn254/solidity.go
+++ b/backend/plonk/bn254/solidity.go
@@ -248,6 +248,31 @@ contract PlonkVerifier {
 
   event PrintUint256(uint256 a);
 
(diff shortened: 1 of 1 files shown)


DavePearce (approved):
Yup, looks good to me

ivokub (approved):

PR #760, opened 2023-07-05
perf: ELM03+Joye07 for emulated scalarMul

yelhousni:
Actually with more select logic and since we are using affine formulas, we can turn double-and-add to add-only while skipping some computations. This PR trades off the y-coordinate computation of the second addition for a select. This would correspond to Alg.1' [[Joye07]](https://www.iacr.org/archive/ches2007/47270135/47270135.pdf) + [[ELM03]](https://arxiv.org/abs/math/0208038).

*new bench*
```
ecrecover (r1cs):  303521
ecrecover (scs):  1455938
ecmul (r1cs):  183756
ecmul (scs):  876507
```

*old bench*
```
ecrecover (r1sc):  365237
ecrecover (scs):  1767044
ecmul (r1sc):  244988
ecmul (scs):  1183191
```


ivokub (approved):
Looks good. Great one!

PR #769, opened 2023-07-06
fix: fixes #768

### LINKED ISSUE #768: The first G1 point in the KZG SRS is not always (1,2)...
In the solidity contract it is assumed that the first entry of the KZG srs (on G1) is the 'canonical' point (1,2).

The template should take in account that it's not always the case.

It's this block in the solidity contract:
```
let folded_evals_commit := mPtr
        mPtr := add(folded_evals_commit, 0x40)
        mstore(folded_evals_commit, 1)                      <---- needs to be generated
        mstore(add(folded_evals_commit, 0x20), 2)  <---- needs to be generated
        mstore(add(folded_evals_commit, 0x40), mload(folded_evals))
        let check_staticcall := staticcall(sub(gas(), 2000),7,folded_evals_commit,0x60,folded_evals_commit,0x40)
        if eq(check_staticcall, 0) {
          error_verify()
        }
```

ThomasPiellard:
G1 point in kzg SRS is generated by the template now...

fixes #768

### DIFF
--- a/backend/plonk/bn254/solidity.go
+++ b/backend/plonk/bn254/solidity.go
@@ -807,8 +807,8 @@ contract PlonkVerifier {
 
         let folded_evals_commit := mPtr
         mPtr := add(folded_evals_commit, 0x40)
-        mstore(folded_evals_commit, 0x1)
-        mstore(add(folded_evals_commit, 0x20), 0x2)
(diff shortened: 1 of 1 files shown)


yelhousni (approved):
Ok LGTM pending CI which should be okay 👍

PR #775, opened 2023-07-12
feat: zeta_power_n_minus_one save and reused in compute_pi

### LINKED ISSUE #774: [PLONK_AUDIT_4-23] z**n-1 computed multiple times
z**n-1 is recomputed several times, it should be stored in the state and computed only once.

ThomasPiellard:
The value `z**n-1` is needed several times in the code and was recomputed. Now it is computed once in `compute_pi`. The function returns `z**n-1` in addition to the public inputs contribution. `z**n-1` is then stored in the memory dedicated to the state.

Fixes #774


ivokub (approved):
Looks good to merge. Waiting for @DavePearce to confirm.

Failing test is due to #776, but didn't cherry-pick here for better observability of audit fixes.

DavePearce (commented):
So, looks fine as is but seems incomplete to me.  In particular `zeta_power_n_minus_one` is returned from `compute_pi()` but then discarded.  Instead, its recomputed in `compute_alpha_square_lagrange_0()` and, at that point, assigned to `state_zeta_power_n_minus_one`.

PR #777, opened 2023-07-13
Gas optimisation and private constant explicitly named

thedarkjester:
Utilises the following techniques:

1. Conversion of fixed variables to constants
2. Use default value initialisation for zero based loops
3. `unchecked`  for limited size collections when incrementing or performing operations with known variables that cannot over/underflow
4. Conversion of some `view` functions to `pure` as no state will be modified (no static calls to precompiles)

PR #781, opened 2023-07-20
Thedarkjester develop

ThomasPiellard:
cf pr #777 , it's a copy of it I just fixed the conflicts.


backend/plonk/bn254/solidity.go
```diff
@@ -23,125 +23,133 @@ pragma solidity ^0.8.0;
...
+      res += uint256(xmsg[47 - i]) << (8 * i);
+      unchecked {
+        i++;
       }
-      res = res % r_mod;
-      uint256 tmp;
-      for (uint i=0; i<16; i++){
-          tmp += uint256(xmsg[15-i])<<(8*i);
+    }
+    res = res % r_mod;
```
thedarkjester, line 94: this can actually be
```
   unchecked {
      res = res % r_mod;
    }
```
drops another 147 gas

ThomasPiellard: 👍 what does the 'unchecked' unchecks here ?

backend/plonk/bn254/solidity.go
```diff
@@ -603,30 +611,36 @@ contract PlonkVerifier {
...
-          
-          uint256 hash_res = Utils.hash_fr(wire_committed_commitments[2*i], wire_committed_commitments[2*i+1]);
-          uint256 a = compute_ith_lagrange_at_z(zeta, zeta_power_n_minus_one, commitment_indices[i]+public_inputs.length);
-          assembly {
-            a := mulmod(hash_res, a, r_mod)
-            pi := addmod(pi, a, r_mod)
-          }
+
+      for (uint256 i; i < vk_nb_commitments_commit_api; ) {
+        uint256 hash_res = Utils.hash_fr(wire_committed_commitments[2 * i], wire_committed_commitments[2 * i + 1]);
```
thedarkjester, line 627: this can also be replaced with 
```
      uint256 hash_fr_x;
      uint256 hash_fr_y;
      unchecked{
        hash_fr_x = wire_committed_commitments[2 * i];
        hash_fr_y = wire_committed_commitments[2 * i + 1];
      }

      uint256 hash_res = Utils.hash_fr(hash_fr_x, hash_fr_y);
```

Another 627 gas saved

backend/plonk/bn254/solidity.go
thedarkjester: this checks the mathematical `%` mod doesn't over/underflow - it really shouldn't be checking, but I think the compiler adds the checks around any math

backend/plonk/bn254/solidity.go
ThomasPiellard: oh nice catch^^ I would never have found that actually, even when you assign values there is a check??

ivokub (approved):
I'm not good to review but approving understanding that both @ThomasPiellard and @thedarkjester have reviewed.

backend/plonk/bn254/solidity.go
ThomasPiellard: ah right

PR #783, opened 2023-07-24
Perf: save 1 `Select` at each iteration in the emulated `scalarMul`

yelhousni:
At each iteration we have 4 `Select` and a `doubleAndAdd` which is either `2R0+R1` or `2R1+R0`. This PR implements a conditional `doubleAndAdd` that first always computes the x-coordinate of `R0+R1` and then adds either `R0` or `R1` based on a `Select`. This saves 1 `Select` at each iteration. 
The PR also substitutes `AddUnified` with simple `add` in the last `bit=0` iteration in `JointScalarMulBase`, which is only used in ECRECOVER where the `scalar=0` edge case does not occur.

The total savings in PLONK for precompiles are:
*ECRECOVER: -11427 scs*
*ECMUL: -2016 scs*

### DIFF
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -330,6 +330,54 @@ func (c *Curve[B, S]) doubleAndAdd(p, q *AffinePoint[B]) *AffinePoint[B] {
 
 }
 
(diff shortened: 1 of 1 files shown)


ivokub (approved):
Looks good!

PR #788, opened 2023-07-25
Perf: Improve MultiLin.Eval number of constraints

Tabaie:
Evaluation of large multilinear extensions is a major bottleneck in GKR verification. Currently the hypercube bases are evaluated at the requested point and the linear combination of these results with the corresponding hypercube values is taken. A massive reduction in the number of constraints can be achieved by performing successive foldings instead. For the 20-variable case we get
```
develop:
r1cs: 2100657 constraints
scs : 4194301 constraints

opt1 (successive fold):
r1cs: 1048576 constraints 49%
scs: 3145726 constraints 75%
```
a 25% improvement for SCS.

A second optimization defers scaling of folding operations which is corrected at the end. My experiments suggest that this starts paying off at around the $2^{16}$ size. It doesn't help with R1CS at all, but doesn't hurt significantly either. With SCS we see significant additional gains:

ops 1&2 (successive fold + defer scaling for large sizes):
```
r1cs: 1048627 constraints 50%
scs: 2097226 constraints 50%
```

### DIFF
--- a/std/polynomial/polynomial.go
+++ b/std/polynomial/polynomial.go
@@ -9,14 +9,26 @@ import (
 type Polynomial []frontend.Variable
 type MultiLin []frontend.Variable
 
+var minFoldScaledLogSize = 16
+
 // Evaluate assumes len(m) = 1 << len(at)
 // it doesn't modify m
 func (m MultiLin) Evaluate(api frontend.API, at []frontend.Variable) frontend.Variable {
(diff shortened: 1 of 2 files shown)


AlexandreBelling (approved):

ivokub (approved):
Confirming @AlexandreBelling approval.

PR #789, opened 2023-07-25
Feat: Export multicommit

AlexandreBelling:
# Description

- Export the Multicommit package
- ~Run defer in reverse order to match golangs behaviour~


ivokub:
> * Run defer in reverse order to match golangs behaviour

Did the current order order lead to implementation problems? It was intentional that we do FIFO as this allows calling additional defers inside deferred functions. We use this in the range checking implementation (we first defer the range checks which defers constructing logderiv argument) imo

AlexandreBelling:
@ivokub 

> > * Run defer in reverse order to match golangs behaviour
> 
> Did the current order order lead to implementation problems? It was intentional that we do FIFO as this allows calling additional defers inside deferred functions. We use this in the range checking implementation (we first defer the range checks which defers constructing logderiv argument) imo

Yes it creates some scoping issues, for instance if I want to use "gkr-for-hashing" and the "multicommit", then I need things to be finalized "GkrHasher" before "MultiCommit". This means that I must wait for the hasher to be initialized before calling multi-commit. But if gnark is already currently dependant on the FIFO order I can carry on.

PR #791, opened 2023-07-26
cache array index - 200 gas

thedarkjester:


ivokub:
Btw, @ThomasPiellard please change the base to `develop`. Right now it was merged to `thedarkjester-develop` which is already merged to `develop`

PR #794, opened 2023-07-26
feat: hash_fr in assembly + removed Utils

ThomasPiellard:
* removed Utils library
* hash_fr is written in assembly

@julien-marchand there might be some conflicts with the 'calldataisation' :/

### DIFF
--- a/backend/plonk/bn254/solidity.go
+++ b/backend/plonk/bn254/solidity.go
@@ -20,57 +20,8 @@ const tmplSolidityVerifier = `// SPDX-License-Identifier: Apache-2.0
 
 pragma solidity ^0.8.19;
 
-library Utils {
-  uint256 private constant r_mod = 21888242871839275222246405745257275088548364400416034343698204186575808495617;
-  uint8 private constant zero = 0;
-  uint8 private constant lenInBytes = 48;
-  uint8 private constant sizeDomain = 11;
-  string private constant dst = "BSB22-Plonk";
-  uint256 private constant b = 6350874878119819312338956282401532410528162663560392320966563075034087161851;
-  bytes private constant zeroBuffer =
-    hex"00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000";
-  uint8 private constant one = 1;
-  uint8 private constant two = 2;
-
-  /**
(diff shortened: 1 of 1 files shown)


backend/plonk/bn254/solidity.go
```diff
@@ -422,15 +384,134 @@ contract PlonkVerifier {
...
+      if iszero(success) {
+        error_sha2_256()
+      }
+
+      // at this point we have mPtr = [ b1 || b2] where b1 is on 32byes and b2 in 16bytes.
+      // we interpret it as a big integer mod r in big endian (similar to regular decimal notation)
+      // the result is then 2**(8*16)*mPtr[32:] + mPtr[32:48]
+      res := mulmod(mload(mPtr), bb, r_mod) // <- res = 2**128 * mPtr[:32]
+      offset := add(mPtr, 0x10)
+      for {let i:=0} lt(i, 0x10) {i:=add(i,1)} // mPtr <- [xx, xx, ..,  | 0, 0, .. 0  ||    b2   ]
```
ivokub, line 499: Not sure about optimisations. Would unrolling the loop help? Otherwise there are implicit checks for overflow when doing add, no?

ivokub (approved):
A remark about comment and a few remarks about optimisations, but I don't know if they help.

backend/plonk/bn254/solidity.go
```diff
@@ -422,15 +384,134 @@ contract PlonkVerifier {
...
+
+      // size domain
+      mstore8(add(mPtr, 0x8e), sizeDomain)
+
+      let success := staticcall(gas(), 0x2, mPtr, 0x8f, mPtr, 0x20)
+      if iszero(success) {
+        error_sha2_256()
+      }
+
+      let b0 := mload(mPtr)
```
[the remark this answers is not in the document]
ThomasPiellard: normally b0 is not needed from here, it is only used to populate the first entry of the next hash. So I leave it in mPtr and use it directly for the next hash

backend/plonk/bn254/solidity.go
ThomasPiellard: I could try to unroll it. But I might get a 'stack too deep' error if the function is too big

PR #795, opened 2023-07-27
Feat/clean compute pi

ThomasPiellard:
Plonk verifier modifs:
---------------------

* Verify is now in a single assembly block -> compute_pi and derive_gamma_beta_alpha_zeta are now in the same block
* Sanity checks are now in the main block in Verify


backend/plonk/bn254/solidity.go
```diff
@@ -236,9 +317,10 @@ contract PlonkVerifier {
       // The data described above is written starting at mPtr. "gamma" lies on 5 bytes,
       // and is encoded as a uint256 number n. In basis b = 256, the number looks like this
       // [0 0 0 .. 0x67 0x61 0x6d, 0x6d, 0x61]. The first non zero entry is at position 27=0x1b
-      function derive_gamma(aproof, pub_inputs) {
+      function derive_gamma(aproof, ins)->gamma_not_reduced {
         
-        let mPtr := mload(0x40)
+        let state := mload(0x40)
+        let mPtr := add(mload(0x40), state_last_mem)
```
ivokub, line 323: Couldn't we use `state` as the first argument here?

ivokub (approved):
LGTM. It is mostly refactoring, but with slight changes (some functions return non-reduced values for challenge computation).

backend/plonk/bn254/solidity.go
ThomasPiellard: yep 👍

PR #816, opened 2023-08-18
Perf: save some negations in emulated pairings

yelhousni:
# Description

@feltroidprime pointed out this morning that pre-computing `-x/y` in emulated pairings would save lines negations at each double/add step in the Miller loop. While it shouldn't in native fields, the idea is interesting for emulated fields. I tried this for 2-chains and emulated BN254 and BLS12-381:
- For 2-chains it does not save anything for both Groth16 and Plonk, as expected.
- For emulated pairings however, the results are a bit weird because of how field emulation works (@ivokub any explanation?): 
  - For BLS12-381, it does not save anything in Groth16.
  - For BN254, we lose -385 constraints in Groth16.
 
However this PR is worth it for Plonk because the addition gates are not free. We save 5960 constraints for BLS12-381 and 2896 for BN254. I applied the changes to 2-chains as well, to homogenise the implementations in native and emulated packages.

## Type of change

Please delete options that are not relevant.

# How has this been tested?

Please describe the tests that you ran or implemented to verify your changes. Provide instructions so we can reproduce.

- [x] Test suite for native and emulated packages.

Benchmarks on fixed and variable input pairings of different sizes (1 to 10).

- [x] Benchmark, on Macbook air M1

# Checklist:

### DIFF
--- a/std/algebra/emulated/sw_bn254/pairing.go
+++ b/std/algebra/emulated/sw_bn254/pairing.go
@@ -347,7 +347,6 @@ func (pr Pairing) MillerLoop(P []*G1Affine, Q []*G2Affine) (*GTEl, error) {
 		// So, 1/y is well defined for all points P's.
 		yInv[k] = pr.curveF.Inverse(&P[k].Y)
 		xNegOverY[k] = pr.curveF.MulMod(&P[k].X, yInv[k])
-		xNegOverY[k] = pr.curveF.MulMod(&P[k].X, yInv[k])
 		xNegOverY[k] = pr.curveF.Neg(xNegOverY[k])
 	}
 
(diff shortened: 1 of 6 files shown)


ivokub:
> * For emulated pairings however, the results are a bit weird because of how field emulation works (@ivokub any explanation?):

Yes, I would assume this is due to how field emulation works. When we negate then we increase the overflow of the limbs (because we have to add the subtraction padding to avoid underflowing individual limbs). When the overflow is larger, then we may hit the amortized reduction (for additions) sooner and we have to carry bigger partitions in the equality check.

std/algebra/emulated/sw_bn254/pairing.go
```diff
@@ -346,7 +346,9 @@ func (pr Pairing) MillerLoop(P []*G1Affine, Q []*G2Affine) (*GTEl, error) {
 		// Anyway (x,0) cannot be on BN254 because -3 is a cubic non-residue in Fp.
 		// So, 1/y is well defined for all points P's.
 		yInv[k] = pr.curveF.Inverse(&P[k].Y)
-		xOverY[k] = pr.curveF.MulMod(&P[k].X, yInv[k])
+		xNegOverY[k] = pr.curveF.MulMod(&P[k].X, yInv[k])
```
ivokub, line 349: Duplicate line?

ivokub (changes requested):
I think there is a duplicate line in bn254 precomputation. Otherwise looks good.

PS, if you removed the benchmark from bls12377, then maybe also remove it from bls24315? It is not a proper benchmark anyway (doesn't depend on the number of loops).

ivokub (approved):
LGTM

PR #817, opened 2023-08-23
feat: add sha3 primitive

NikitaMasych:
This PR is intended to bring SHA3 into the gnark. 
Mainly, it implements sponge construction over Keccak F-[1600] permutation and adds following schemes:

### DIFF
--- a/std/hash/sha3/doc.go
+++ b/std/hash/sha3/doc.go
@@ -1,0 +1,5 @@
+// Package sha3 provides ZKP circuits for SHA3 hash algorithms applying sponge construction over
+// Keccak f-[1600] permutation function.
+//
+// Instances correspond golang.org/x/crypto/sha3, except SHA224, which is not x64 compatible.
+package sha3
--- a/std/hash/sha3/hashes.go
+++ b/std/hash/sha3/hashes.go
@@ -1,0 +1,94 @@
+package sha3
+
+import (
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/std/hash"
+	"github.com/consensys/gnark/std/math/uints"
+)
+
+// New256 creates a new SHA3-256 hash.
+// Its generic security strength is 256 bits against preimage attacks,
+// and 128 bits against collision attacks.
+func New256(api frontend.API) (hash.BinaryHasher, error) {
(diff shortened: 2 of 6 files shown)


ivokub (approved):
Looks good to me. Thanks for the contribution.
Not necessary for this PR, but maybe you have ideas how to implement `std/hash.BinaryFixedLengthHasher` interface?

I suggested some edits to remove the global variable in the test cases so that we can run the tests in parallel and then added different SHA instances as a subtest.

I also checked against the full solver and it works well. I didn't run against the full prover right now.

NikitaMasych:
Applied your suggestion! 

Straightforward approach for `std/hash.BinaryFixedLengthHasher` would be to trunc input array, and send it for padding, but for that it is necessary to convert argument `length frontend.Variable` to int; I tried to use `api.Compiler().ConstantValue(length)` for this purpose, but it doesn't seem to work (I am definitely missing something here). #809 confirms that it is not possible. For now I have no other ideas of resolution.
I am interested in possible applications of the `hash.BinaryFixedLengthHasher`, can you provide me some examples where having such a method might be useful?

ivokub:
Thanks for the update. I'll wait for the CI and merge. Really appreciate your contribution.

> I am interested in possible applications of the `hash.BinaryFixedLengthHasher`, can you provide me some examples where having such a method might be useful?

The sizes of the circuits are fixed, so we have to allocate sufficient number of variables to be able to fit maximal possible input size. This also means that for SHA3 (when implementing hash.BinaryHasher) we only allow inputs of fixed length if we want to prove the preimage-knowledge of the digest. However, it may happen that the preimage size is not known at compile-time or we may have different inputs with different input sizes. This happens for example in zkEVM where if we want to support precompiles and keccak opcode, we have to allow inputs of differents size.

liyue201:
> Looks good to me. Thanks for the contribution. Not necessary for this PR, but maybe you have ideas how to implement `std/hash.BinaryFixedLengthHasher` interface?
> 
> I suggested some edits to remove the global variable in the test cases so that we can run the tests in parallel and then added different SHA instances as a subtest.
> 
> I also checked against the full solver and it works well. I didn't run against the full prover right now.

I have implemented this interface at https://github.com/Consensys/gnark/pull/821, please review

PR #819, opened 2023-08-25
fs in solidity compliant with the go implem

ThomasPiellard:
# Description

Fiat Shamir changes in plonk are added in the solidity template


gbotrel:
@ThomasPiellard the solidity template is missing the function `hex` . that is: 
```
func (vk *VerifyingKey) ExportSolidity(w io.Writer) error {
	funcMap := template.FuncMap{
		// The name "inc" is what the function will be called in the template text.
		"inc": func(i int) int {
			return i + 1
		},
--> missing hex here, you probably forgot to modify the .go.tmpl file
```

gbotrel (approved):

PR #825, opened 2023-09-07
perf, feat: `assert.CheckCircuit(...)`

gbotrel:
# Description

This PR tackles the issue of heterogeneous tests for circuits and long runs in the CI.
Previously some tests call the test engine directly, sometime we commit tests that only check one curve and one backend to go faster, ... ; 

Proposition here is to replace `assert.ProverSucceeded` `assert.ProverFailed` ``assert.SolverSucceeded` ... with `assert.CheckCircuit`. 

Through build tags and `-short` test option, an increasing number of checks are performed;
```
//	go test -short 				--> testEngineChecks
//	go test 						--> testEngineChecks  + constraintSolverChecks
//	go test -tags=prover_checks 	--> ... + proverChecks
//	go test -tags=release_checks 	--> ... + releaseChecks (solidity, serialization, ...)
```

This allow a fast dev workflow (test engine only with `-short` option) and various CI workflows to do extra validations.

This PR also factorize / clean up quite a bit of repeated code (for example for testing serialization round trips).

Last but not least, `test.Engine` has a `BatchInvert` method called in `logderivarg.go` `Build()` method; it saves 30% of test time since a lot of inverses using `big.Int` are performed in the test engine. A potential follow up to this PR would be to have the test engine parametrized with field element instead of `big.Int`.

Fixes # (issue)

This PR also fixes some minor issues;

- `groth16.Proof` serialization was missing commitment info.
- `witness.ReadFrom` returned incorrect number of bytes read when deserializing
- 

## Breaking changes

- `test.NoSerialization()` -> `test.NoSerializationChecks()`

## Type of change

- [x ] This change requires a documentation update

`go test -short ./...` up to 90% faster.

# Checklist:

- [ x] I have performed a self-review of my code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] I did not modify files generated from templates
- [ x] `golangci-lint` does not output errors locally
- [ x] New and existing unit tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules


internal/stats/stats_test.go
```diff
@@ -1,15 +1,16 @@
...
 
 	"github.com/consensys/gnark/backend"
-	"github.com/consensys/gnark/test"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCircuitStatistics(t *testing.T) {
 	const refPath = "latest.stats"
-	assert := test.NewAssert(t)
+	assert := require.New(t)
```
ivokub, line 13: Why use require directly instead of assert?

test/assert_options.go
```diff
@@ -29,14 +15,58 @@ import (
...
+	for _, option := range opts {
+		err := option(&opt)
+		assert.NoError(err, "parsing TestingOption")
+	}
+
+	return opt
+}
+
+// WithValidAssignment is a testing option which adds a valid assignment
+func WithValidAssignment(validAssignment frontend.Circuit) TestingOption {
```
ivokub, line 57: What about having a variation of this option (in addition to this one) called `WithValidAssignmentMaker` (etc instead of `Maker`) which constructs a new witness every time is called? Sometimes when we have some complex witness types (a la `Element` in non-native package) then we change some internal fields here and there to keep track of the usage inside circuit (for example, if something has been range-checked).

This may lead to failures in determinism tests. Even though the circuits can be fixed (and I usually have, by implementing cleanup or storing some stuff in internal caches of the non-native API), it may be quite inconvenient.

std/evmprecompiles/bn_test.go
```diff
@@ -127,10 +123,7 @@ func TestECMulCircuitFull(t *testing.T) {
 	t.Skip("skipping very long test")
 	assert := test.NewAssert(t)
 	circuit, witness := testRoutineECMul(t)
-	assert.ProverSucceeded(circuit, witness,
-		test.NoFuzzing(), test.NoSerialization(),
-		test.WithBackends(backend.GROTH16, backend.PLONK), test.WithCurves(ecc.BN254),
-	)
+	assert.CheckCircuit(circuit, test.WithValidAssignment(witness))
```
ivokub, line 126: I think in addition to changing from `ProverSucceeded` to `CheckCircuit`, we can also enable the test. Previously as there wasn't a way to distinguish between release test and PR test, I had it skipped. But now I think it would be OK to re-enable.

std/evmprecompiles/01-ecrecover_test.go
```diff
@@ -134,8 +133,9 @@ func TestECRecoverCircuitFull(t *testing.T) {
 	t.Skip("skipping very long test")
```
ivokub, line 133: Remove skip?

ivokub:
Suggested edit:
```diff
diff --git a/test/assert.go b/test/assert.go
index 35ed4a45..81c346db 100644
--- a/test/assert.go
+++ b/test/assert.go
@@ -72,7 +72,7 @@ func (assert *Assert) Log(v ...interface{}) {
 	assert.t.Log(v...)
 }
 
-// ProverSucceeded is deprecated: use CheckCircuit instead
+// ProverSucceeded is deprecated: use [Assert.CheckCircuit] instead
 func (assert *Assert) ProverSucceeded(circuit frontend.Circuit, validAssignment frontend.Circuit, opts ...TestingOption) {
 	// copy the options
 	newOpts := make([]TestingOption, len(opts), len(opts)+2)
@@ -82,7 +82,7 @@ func (assert *Assert) ProverSucceeded(circuit frontend.Circuit, validAssignment
 	assert.CheckCircuit(circuit, newOpts...)
 }
 
-// ProverSucceeded is Deprecated use CheckCircuit instead
+// ProverSucceeded is deprecated use [Assert.CheckCircuit] instead
 func (assert *Assert) ProverFailed(circuit frontend.Circuit, invalidAssignment frontend.Circuit, opts ...TestingOption) {
 	// copy the options
 	newOpts := make([]TestingOption, len(opts), len(opts)+2)
@@ -92,7 +92,7 @@ func (assert *Assert) ProverFailed(circuit frontend.Circuit, invalidAssignment f
 	assert.CheckCircuit(circuit, newOpts...)
 }
 
-// SolvingSucceeded is deprecated: use CheckCircuit instead
+// SolvingSucceeded is deprecated: use [Assert.CheckCircuit] instead
 func (assert *Assert) SolvingSucceeded(circuit frontend.Circuit, validWitness frontend.Circuit, opts ...TestingOption) {
 
 	// copy the options
diff --git a/test/assert_checkcircuit.go b/test/assert_checkcircuit.go
index e5bb6b96..f3502ee1 100644
--- a/test/assert_checkcircuit.go
+++ b/test/assert_checkcircuit.go
@@ -14,19 +14,19 @@ import (
 
 // CheckCircuit performs a series of check on the provided circuit.
 //
-//	go test -short 					--> testEngineChecks
-//	go test 						--> testEngineChecks  + constraintSolverChecks
-//	go test -tags=prover_checks 	--> ... + proverChecks
-//	go test -tags=release_checks 	--> ... + releaseChecks (solidity, serialization, ...)
+//	go test -short                  --> testEngineChecks
+//	go test                         --> testEngineChecks  + constraintSolverChecks
+//	go test -tags=prover_checks     --> ... + proverChecks
+//	go test -tags=release_checks    --> ... + releaseChecks (solidity, serialization, ...)
 //
 // Depending on the above flags, the following checks are performed:
-// - the circuit compiles
-// - the circuit can be solved with the test engine
-// - the circuit can be solved with the constraint system solver
-// - the circuit can be solved with the prover
-// - the circuit can be verified with the verifier
-// - the circuit can be verified with gnark-solidity-checker
-// - the circuit, witness, proving and verifying keys can be serialized and deserialized
+//   - the circuit compiles
+//   - the circuit can be solved with the test engine
+//   - the circuit can be solved with the constraint system solver
+//   - the circuit can be solved with the prover
+//   - the circuit can be verified with the verifier
+//   - the circuit can be verified with gnark-solidity-checker
+//   - the circuit, witness, proving and verifying keys can be serialized and deserialized
 func (assert *Assert) CheckCircuit(circuit frontend.Circuit, opts ...TestingOption) {
 	// get the testing configuration
 	opt := assert.options(opts...)

```

ivokub (approved):
I think the approach is really good. A few questions about additional behaviour control options (which I'm not sure are worth it).

Also added a few suggestions to make godoc output look nicer.

test/assert_checkcircuit.go
```diff
@@ -0,0 +1,285 @@
...
+		assert.Run(func(assert *Assert) {
+
+			// parse valid / invalid assignments
+			var invalidWitnesses, validWitnesses []_witness
+			for _, a := range opt.validAssignments {
+				w := assert.parseAssignment(circuit, a, curve, opt.checkSerialization)
+				validWitnesses = append(validWitnesses, w)
+
+				// check that the assignment is valid with the test engine
+				err := IsSolved(circuit, w.assignment, curve.ScalarField())
```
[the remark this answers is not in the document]
ivokub: Another use case for avoiding test engine - when I'm trying to debug hint functions (particularly a case where I replace a standard hint with a malicious one), then because test engine does not use hint registry but instead calls the hint functions directly, then I'm not able to override hint function in the test engine.

And I'm not able to make tests pass as I'm not getting to solver as the test will fail with the test engine (but should succeed with solver).

internal/stats/stats_test.go
gbotrel: before used to import `gnark/test` ; but we don't need this package here, in this context we just do a "normal unit test", no circuit stuff going on.

test/assert.go
```diff
@@ -87,237 +72,45 @@ func (assert *Assert) Log(v ...interface{}) {
...
-	}
-
-	if opt.fuzzing {
-		// TODO may not be the right place, but ensures all our tests call these minimal tests
-		// (like filling a witness with zeroes, or binary values, ...)
-		assert.Run(func(assert *Assert) {
-			assert.Fuzz(circuit, 5, opts...)
-		}, "fuzz")
-	}
+	assert.CheckCircuit(circuit, newOpts...)
```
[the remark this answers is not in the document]
gbotrel: mmhh yeah I thought about that, but then we would need to refactor all the places we use these methods to use CheckCircuit else we are back to having a veryyyyy long test flow. My plan was to deprecate this anyway --

PR #837, opened 2023-09-23
Refactor/cleanupplonkprover

gbotrel:
wip refactor of feat/plonk_memory_optim by @ThomasPiellard


ThomasPiellard:
`solveConstraints` and `buildRatioCopyConstraint` could both be split in two: first the actual computation of the vectors and then the commitment to the vectors. Also 'deriveAlpha' should be a separate function with a dedicated channel. The 'evaluateConstraints' depends only on alpha, so you just just need to wait for the corresponding chan (since computing alpha requires committing to everything prior to the computation of the quotient)

PR #855, opened 2023-10-10
perf, refactor: plonk prover

gbotrel:
# Description

Sister PR of https://github.com/Consensys/gnark-crypto/pull/456 . 

Goal is to make PlonK prover less memory greedy, more performant and more readable.

* uses (more) FFT but on the small domain (by @ThomasPiellard )
* stores the KZG SRS in canonical and Lagrange form which avoids some FFT and may result in a future PR in some additional perf gains (by @ThomasPiellard )

4M constraints;

```
benchmark                    old ns/op      new ns/op      delta
BenchmarkProver/bn254-96     5232334900     4284055453     -18.12%

benchmark                    old allocs     new allocs     delta
BenchmarkProver/bn254-96     20763          18963          -8.67%

benchmark                    old bytes      new bytes      delta
BenchmarkProver/bn254-96     6348612832     4301645176     -32.24%
``` 

The memory gain is actually much larger than this, since the `extended trace` in the `ProvingKey` is gone, for large circuits that's dozen if not hundreds of GB.


ThomasPiellard (approved):

ThomasPiellard:
I added just one thing: the commitment to the bsb22 polynomials now use the srs in Lagrange basis, those polynomials are sparse in Lagrange form, otherwise looks good

PR #846, opened 2023-10-04
feat: BW6-761 emulated pairing

yelhousni:
# Description

This PR implements BW6-761 emulated pairing circuit. It is based on #589 (mainly for the tower arithmetic). It follows https://hackmd.io/@gnark/BW6-761-changes:
- da202f354469be1b0b91ff649eaf218c23c49a0f implements equation (4') 
- 118edc34b6cc3319fabf38b77e97cf0f07a202de implements equation (2)
- a266590f70147956726228ed9414ec40b78ef1bb implements equation (1)

All these equations are implemented using similar tricks to BN254 and BLS12-381 emulated pairings (tower hints, affine coordinates, sparser lines, iterations isolation...).
I was expecting (4') to be the optimal choice, for a single pairing, in terms of constraints but it seems not. I will need to try with the optimal Tate equations.

Todo:
- [ ] ~~equation (5)~~
- [ ] ~~equation (5')~~
- [x] equation (6) ⭐️ 7fc5925
- [ ] ~~equation (6')~~
- [x] ~~Torus-based final exponentiation~~ 
(not enough multiplications in `Expt` for torus-based FE to be worth it vs. 46 Karabina squarings in a row)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

- [x] Pairing (and multi-pairing) and final exponentiation results compared to gnark-crypto. ~~This needs to change the gnark-crypto branch to `feat/bw6761-pairing`~~. The final chosen algorithm is the same as gnark-crypto `master`, so no nothing to do there anymore. 

# How has this been benchmarked?

- [x] Number of constraints:

|                     | Miller Loop | Final exp | 1 pairing | 2 pairings |
| ----------- | ----------- | ----------- | -------------- | ------------- |
| R1CS           | 1,548,496    | 1,037,242    | 2,585,738       | 3,746,835     |
| SCS             | 16,520,533 | 11,372,377  | 27,892,910     |  39,656,303  |

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #749, opened 2023-07-03
perf: non-native modular multiplication

ivokub:
# Description

This PR implements improved non-native modular multiplication as described in https://hackmd.io/3cpvkONzQl-TF5Oj8VXEHQ.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How has this been tested?

The existing test suite passes.

# How has this been benchmarked?

One of the biggest circuits we have - BW6-761 PLONK verifier in BN254 PLONK reduced from 180M constraints to 80M.

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #861, opened 2023-10-13
chore: avoid nonnative dereferences

ivokub:
# Description

With the upcoming pr #749 it is necessary that we do not change the limb values of `emulated.Element` values. Mostly the API is designed in a way that this doesn't happen (we pass in and out pointers and do not perform in-place modifications), but there are a few cases in nonnative algebra where the API is overridden. This is mostly due to that for structs containing nonnative elements we need to use the value instead of pointer to count the number of field elements to allocate and have initialized values.

For example, if we have definition like:
```go
type G1Affine struct {
    X, Y emulated.Element[emparams.BN254Fp]
}
```
and we compute
```go
var P G1Affine
P = curve.Add(c.G, c.Q)
P.X = *emulated.Add(&p.X, other)
```
then we modify the value `P.X`. Now, as the multiplication checks are deferred for amortizing some computations, we get a failed check.

This PR changes such occurrences, making #749 more streamlined. Similar changes are already incorporated in #846.

## Type of change

# How has this been tested?

- [x] Rebased #749 on top and all tests are working.

Not benchmarked, circuit stats are the same are the same.

# Checklist:

### DIFF
--- a/std/algebra/emulated/fields_bls12381/e12_pairing.go
+++ b/std/algebra/emulated/fields_bls12381/e12_pairing.go
@@ -80,8 +80,7 @@ func (e Ext12) ExptTorus(x *E6) *E6 {
 //	}
 func (e *Ext12) MulBy014(z *E12, c0, c1 *E2) *E12 {
 
-	a := z.C0
-	a = *e.MulBy01(&a, c0, c1)
+	a := e.MulBy01(&z.C0, c0, c1)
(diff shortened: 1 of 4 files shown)


ivokub:
@gbotrel - pinging. This PR is a pre-requisite for #749.

gbotrel (approved):

PR #868, opened 2023-10-17
Fix: edge cases in the Karabina cyclotomic square decompression

### LINKED ISSUE #858: Handle Karabina decompression edge cases in-circuit
Some edge cases in Karabina decompression might occur when using `PairingCheck()`. These need to be handled in-circuit for 2-chains and BW6-761 emulated pairing similarly to [gnark-crypto](https://github.com/Consensys/gnark-crypto/blob/206544105834ddf11090b556d6c7b055469d5116/ecc/bw6-761/internal/fptower/e6.go#L225).

yelhousni:
# Description

Fixes #858 

TODO: 
- [x] Emulated BW6-671
- [x] Native BLS12-377
- [ ] Native BLS24-315

## Type of change

# How has this been tested?

The multi-pairing computation `e(P,Q)*e(-P,Q)` returns 1. In this case the `g3 == g2 == 0` edge case is triggered. The function `PairingCheck()` and corresponding test were added to emulated bw6-761 and 2-chains.

WIP

# Checklist:

### DIFF
--- a/std/algebra/native/fields_bls12377/e12.go
+++ b/std/algebra/native/fields_bls12377/e12.go
@@ -263,28 +263,44 @@ func (e *E12) CyclotomicSquareCompressed(api frontend.API, x E12) *E12 {
 // Decompress Karabina's cyclotomic square result
 func (e *E12) Decompress(api frontend.API, x E12) *E12 {
 
-	// TODO: hadle the g3==0 case with MUX
-
 	var t [3]E2
+	var _t [2]E2
 	var one E2
 	one.SetOne()
 
-	// t0 = g1²
+	// if g3 == 0
+	// t0 = 2 * g1 * g5
+	// t1 = g2
+	selector1 := x.C1.B0.IsZero(api)
+	_t[0].Square(api, x.C0.B1)
+	_t[0].Double(api, _t[0])
+	_t[1] = x.C0.B2
+
+	// if g2 == g3 == 0
+	selector2 := _t[1].IsZero(api)
+
+	// if g3 != 0
+	// t0 = E * g5^2 + 3 * g1^2 - 2 * g2
+	// t1 = 4 * g3
 	t[0].Square(api, x.C0.B1)
-	// t1 = 3 * g1² - 2 * g2
 	t[1].Sub(api, t[0], x.C0.B2).
 		Double(api, t[1]).
 		Add(api, t[1], t[0])
-	// t0 = E * g5² + t1
 	t[2].Square(api, x.C1.B2)
 	t[0].MulByNonResidue(api, t[2]).
 		Add(api, t[0], t[1])
-	// t1 = 4 * g3
 	t[1].Double(api, x.C1.B0).
 		Double(api, t[1])
-	// z4 = g4 / t1
+
+	// g4 = (E * g5^2 + 3 * g1^2 - 2 * g2)/4g3 or (2 * g1 * g5)/g2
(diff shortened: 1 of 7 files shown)


std/algebra/native/fields_bls24315/e24.go
```diff
@@ -261,23 +261,38 @@ func (e *E24) CyclotomicSquareCompressed(api frontend.API, x E24) *E24 {
...
 	var t [3]E4
+	var _t [2]E4
 	var one E4
 	one.SetOne()
 
-	// t0 = g1²
+	// if g3 == 0
+	// t0 = 2 * g1 * g5
+	// t1 = g2
+	selector1 := x.D1.C0.IsZero(api)
```
ivokub, line 271: Don't we have to also cover the edge case where g4==0? (as in fields_bw6761)

std/algebra/native/fields_bls12377/e12.go
```diff
@@ -263,28 +263,42 @@ func (e *E12) CyclotomicSquareCompressed(api frontend.API, x E12) *E12 {
...
 	var t [3]E2
+	var _t [2]E2
 	var one E2
 	one.SetOne()
 
-	// t0 = g1²
+	// if g3 == 0
+	// t0 = 2 * g1 * g5
+	// t1 = g2
+	selector1 := x.C1.B0.IsZero(api)
```
ivokub, line 274: Don't we have to also cover the edge case where g4==0? (as in fields_bw6761)

ivokub (approved):
I'm not sure if `g4=0` is relevant for two-chains. But otherwise looks good.

yelhousni:
> I'm not sure if `g4=0` is relevant for two-chains. But otherwise looks good.

It is covered implicitly to save some constraints. When `g3==g2==0` we should return 1 but we do nothing because the result in 2-chains happen to be 1 implicitly. First, `g4` is set to 0` because `g4=e.C1.B1.DivUnchecked(api, t[0], t[1])` uses `E2` Inverse which is a `fp.Element` Inverse which returns conventionally 0 when denominator is 0. Then the remaining computation returns `E * (2 * g4² + g3 * g5 - 3 * g2 * g1) + 1` which is exactly 1 because `g2==g3==g4==0`. We cannot do the same implicit trick in emulated BW6-761 because there the hinted division use big.Int ModInverse. There is a comment explaining this here: https://github.com/Consensys/gnark/blob/baaf4ba836f8bfb7c468b7e46c101f3c8f436269/std/algebra/native/fields_bls12377/e12.go#L295

ivokub:
> > I'm not sure if `g4=0` is relevant for two-chains. But otherwise looks good.
> 
> It is covered implicitly to save some constraints. When `g3==g2==0` we should return 1 but we do nothing because the result, in 2-chains, happen to be 1 implicitly. First, `g4` is set to 0 because `g4=e.C1.B1.DivUnchecked(api, t[0], t[1])` uses `E2` Inverse which is a `fp.Element` Inverse which returns conventionally 0 when denominator is 0. Then the remaining computation returns `E * (2 * g4² + g3 * g5 - 3 * g2 * g1) + 1` which is exactly 1 because `g2==g3==g4==0`. We cannot do the same implicit trick in emulated BW6-761 because there the hinted division uses `big.Int` `ModInverse`. There is a comment explaining this here:
> 
> https://github.com/Consensys/gnark/blob/baaf4ba836f8bfb7c468b7e46c101f3c8f436269/std/algebra/native/fields_bls12377/e12.go#L295

Makes sense!

PR #870, opened 2023-10-17
fix: non-native arithmetic autoreduction for division, inversion and sqrt

### LINKED ISSUE #867: bug: omitted automatic reductions for Div, Inverse and Sqrt in field emulation
There was an edge case in #866 which is due to not automatically reducing before division operation.

## Description
We keep track of the overflows and usually reduce before the overflow of the result would overflow the scalar field. But division does many things internally:
1. result using hint
2. multiplication of the result and divisor
3. equality check of previous and dividend

The problem appears in step 3 - the overflow of of the result is already 189, but inside equality check we have to add the subtraction padding, adding two bits of overflow, so we are at 191 bits. But this is overflow only, so the whole limb is on 255 bits, which is the same than bls12-381 scalar field modulus width. We get a scalar field overflow and nothing works anymore.

The issue has always been around, it only appears more clearly when we have a lot of limbs because right now the multiplication overflow (step 2) depends on the number of limbs. And in #866 unreduced values have 37+ limbs, so overflow grows very quickly.

## Expected Behavior

We check if all the intermediate values do not overflow the scalar field.

## Possible Fix

Use the auto-reduce machinery for division. The issue also appears for Sqrt and Inverse, so should also add auto-reduce for those methods.

ivokub:
# Description

Adds automatic modular reductions to division, inversion and square roots in field emulation when the intermediate results do not fit into the scalar field of the curve.

Fixes #867 

## Type of change

# How has this been tested?

- [x] Added regression test where the bug lead to arithmetic failure and is fixed with the fix.

Not benchmarked.

# Checklist:


ivokub:
@gbotrel - I cannot reproduce locally. It is timeouting, but I do not understand if only for a single test (TestInverse over Goldilocks) or for the whole suite. If only for a single test, then may be some infinite recursion (which doesn't happen locally), but makes sense for the whole suite as we have added emulated Groth16/KZG proofs etc.

gbotrel (approved):

PR #875, opened 2023-10-19
test: refactor and avoid concurrent call to register

gbotrel:
# Description

* Make `std/gkr` use `assert.CheckCircuit`
* When `NoTestEngine` is set, skip `ccs.Solve` checks in `test` package.

### DIFF
--- a/test/assert_checkcircuit.go
+++ b/test/assert_checkcircuit.go
@@ -78,6 +78,9 @@ func (assert *Assert) CheckCircuit(circuit frontend.Circuit, opts ...TestingOpti
 					// 2- if we are not running the full prover;
 					// we need to run the solver on the constraint system only
 					if !opt.checkProver {
+						if opt.skipTestEngine {
+							return
+						}
 						for _, w := range invalidWitnesses {
 							w := w
 							assert.Run(func(assert *Assert) {
--- a/std/gkr/api_test.go
+++ b/std/gkr/api_test.go
@@ -2,27 +2,25 @@ package gkr
 
 import (
 	"fmt"
-	"github.com/consensys/gnark-crypto/kzg"
-	"github.com/consensys/gnark/backend/plonk"
-	bn254r1cs "github.com/consensys/gnark/constraint/bn254"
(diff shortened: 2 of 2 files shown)


test/assert_checkcircuit.go
```diff
@@ -78,6 +78,9 @@ func (assert *Assert) CheckCircuit(circuit frontend.Circuit, opts ...TestingOpti
 					// 2- if we are not running the full prover;
 					// we need to run the solver on the constraint system only
 					if !opt.checkProver {
+						if opt.skipTestEngine {
```
ivokub, line 81: Why we're skipping the solving when `NoTestEngine` set? Usually I'm skipping the test engine when I have modified the solver(e.g. for testing hint override option. This doesn't work with test engine as test engine calls hints directly and cannot be overidden), but now I cannot call solver on its own at all?

ivokub (commented):
Otherwise looks good, I'm not sure about skipping solver with `NoTestEngine`

test/assert_checkcircuit.go
ivokub: As an example, I have used it here: https://github.com/Consensys/gnark/blob/master/internal/regression_tests/issue_836_test.go#L92

gbotrel: Yeah issue on this one... is that for GKR for now some circuits are "unsolvable" by the solver, totally a misused of the flag here I removed it. 

i.e. with GKR the hint is valid only with a backend, need to replace the placeholder hint with something valid for the solver.

ivokub: Indeed, when we run with test engine then we need to have a working hint. Or we could right now shortcut it by having a specific error type which essentially says "I'm a hint only working with actual backend - skip solving with test engine" etc.

ivokub: Working on right now on GKR - actually the GKR hints are not dependent on the backend. We can overload the hints already in the solver. This would fix this issue.

Should I do a separate PR for that or add to this one?

PR #879, opened 2023-10-20
feat: emulated mimc (WIP)

### LINKED ISSUE #848: feat: emulated MiMC
Needed for #847.

Should be relatively straight-forward. Should support all known parameters from gnark-crypto, but it is sufficient if we start with the parameters required for proof aggregation (BLS12-377 scalar field).

ivokub:
Work in progress WIP for emulated MiMC. Do not merge!

Open issues right now:
* [ ] we get an import cycle because we use mimc in for logderivargument and multicommit. Both in both cases we do not need to use MiMC specifically. There, we can also do the squaring instead of using MiMC. Actually this is already implemented in the mulmod PR. After we have merged that we can change mimc.NewMiMC to return hash.FieldHasher instead. Then it is clearer what the function is about. Most importantly, then we can also implement `hash.FieldHasherGeneric[t emulated.FieldParam]` as an interface.
* [ ] implement emulated mimc
* [ ] implement Fiat-Shamir using `hash.FieldHasherGeneric`

cc @ThomasPiellard 

# Description

Fixes #848 

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/examples/rollup/circuit.go
+++ b/examples/rollup/circuit.go
@@ -21,6 +21,7 @@ import (
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/std/accumulator/merkle"
 	"github.com/consensys/gnark/std/algebra/native/twistededwards"
+	"github.com/consensys/gnark/std/hash"
 	"github.com/consensys/gnark/std/hash/mimc"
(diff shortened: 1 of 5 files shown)


ivokub:
We use alternative approach in PLONK verifier.

PR #880, opened 2023-10-20
feat: add PLONK in-circuit verifier

### LINKED ISSUE #847: feat: implement Plonk verifier in-circuit
With #840 it becomes easier to implement generic Plonk verifier in-circuit for emulated and native fields (2-chains).

This requires:
* [x] hash-to-field. Hash to field is used to map the commitment as a public input. Right now we use the standard hash-to-field but this is too expensive as uses SHA2 internally. We should consider using MiMC directly.
* [x] for that, we need emulated MiMC hashing.
* [x] we also compute a Fiat-Shamir challenge which right now is hardcoded with SHA2. We need to make the hash function for the FS and hash-to-field configurable. This means that we need to add prover/verifier options in gnark which allows to change the implementations.
* [x] add batch KZG verification to gadget. We have implemented KZG verification in #840, but for a single instance. Using folding it is possible to make the verification more efficient.

ivokub:
# Description

This PR adds in-circuit verifier supporting commitments for both 2-chain and emulated computations.

Fixes #847 

## Type of change

# How has this been tested?

- [x] 2-chain test without commitments `TestBLS12InBW6WoCommit`
- [x] emulated test without commitments `TestBW6InBN254WoCommit`
- [x] 2-chain test with commitments `TestBLS12InBW6Commit`
- [x] emulated test with commitments `TestBW6InBN254Commit`

# Checklist:

### DIFF
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -647,7 +647,7 @@ func (c *Curve[B, S]) MultiScalarMul(p []*AffinePoint[B], s []*emulated.Element[
 	res := c.ScalarMul(p[0], s[0])
 	for i := 1; i < len(p); i++ {
 		q := c.ScalarMul(p[i], s[i])
-		c.AddUnified(res, q)
+		res = c.AddUnified(res, q)
 	}
 	return res, nil
 }
--- a/std/algebra/native/sw_bls12377/pairing2.go
+++ b/std/algebra/native/sw_bls12377/pairing2.go
@@ -76,7 +76,6 @@ func (c *Curve) Add(P, Q *G1Affine) *G1Affine {
 // AssertIsEqual asserts the equality of P and Q.
 func (c *Curve) AssertIsEqual(P, Q *G1Affine) {
 	P.AssertIsEqual(c.api, *Q)
-	panic("todo")
 }
 
 // Neg negates P and returns the result. Does not modify P.
--- a/std/algebra/native/sw_bls12377/pairing2.go
+++ b/std/algebra/native/sw_bls12377/pairing2.go
@@ -126,7 +126,12 @@ func (c *Curve) MultiScalarMul(P []*G1Affine, scalars []*Scalar) (*G1Affine, err
 	res := c.ScalarMul(P[0], scalars[0])
 	for i := 1; i < len(P); i++ {
 		q := c.ScalarMul(P[i], scalars[i])
-		res = c.Add(res, q)
+
+		// check for infinity
+		isInfinity := c.api.And(c.api.IsZero(P[i].X), c.api.IsZero(P[i].Y))
+		tmp := c.Add(res, q)
+		res.X = c.api.Select(isInfinity, res.X, tmp.X)
+		res.Y = c.api.Select(isInfinity, res.Y, tmp.Y)
 	}
 	return res, nil
 }
--- a/std/algebra/native/sw_bls12377/pairing2.go
+++ b/std/algebra/native/sw_bls12377/pairing2.go
@@ -126,7 +126,7 @@ func (c *Curve) MultiScalarMul(P []*G1Affine, scalars []*Scalar) (*G1Affine, err
 	res := c.ScalarMul(P[0], scalars[0])
 	for i := 1; i < len(P); i++ {
 		q := c.ScalarMul(P[i], scalars[i])
-		c.Add(res, q)
+		res = c.Add(res, q)
 	}
 	return res, nil
 }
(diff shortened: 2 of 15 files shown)


ivokub:
Ready from my side. @ThomasPiellard - if you can confirm that also good from your side then I think it is good for review. I assigned @yelhousni for review as we both have worked on it :)

std/algebra/native/sw_bls12377/pairing2.go
```diff
@@ -126,7 +126,12 @@ func (c *Curve) MultiScalarMul(P []*G1Affine, scalars []*Scalar) (*G1Affine, err
 	res := c.ScalarMul(P[0], scalars[0])
 	for i := 1; i < len(P); i++ {
 		q := c.ScalarMul(P[i], scalars[i])
-		c.Add(res, q)
+
```
yelhousni, line 129: I think `ScalarMul` accounts for when `P[i]=(0,0)` no? It's `Add` that doesn't. Maybe use `AddUnified`?

std/algebra/native/sw_bls24315/pairing2.go
```diff
@@ -126,7 +126,12 @@ func (c *Curve) MultiScalarMul(P []*G1Affine, scalars []*Scalar) (*G1Affine, err
 	res := c.ScalarMul(P[0], scalars[0])
 	for i := 1; i < len(P); i++ {
 		q := c.ScalarMul(P[i], scalars[i])
-		c.Add(res, q)
+
```
yelhousni, line 129: ditto

yelhousni (commented):
LGTM but I need a second round of review

std/algebra/native/sw_bls12377/pairing2.go
ivokub: Yes, `Add` doesn't. Imo we do not have `AddUnified` for 2-chains, only for emulated?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -647,7 +647,7 @@ func (c *Curve[B, S]) MultiScalarMul(p []*AffinePoint[B], s []*emulated.Element[
 	res := c.ScalarMul(p[0], s[0])
 	for i := 1; i < len(p); i++ {
 		q := c.ScalarMul(p[i], s[i])
-		c.AddUnified(res, q)
+		res = c.AddUnified(res, q)
```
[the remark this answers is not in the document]
ivokub: 👍  will add. I spent whole day debugging :)

std/algebra/native/sw_bls12377/pairing2.go
ThomasPiellard: @yelhousni at some point you contemplated using a twisted Edwards model of the curve for this ? Is it still on the table ?

std/algebra/emulated/sw_emulated/point.go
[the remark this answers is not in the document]
ivokub: Added scalar mul test.

ivokub:
Added MSM tests for emulated and 2-chain curves. This revealed another bug :)

I think the PR is done from my side. We could merge from my side after approving review. I think we can address additional optimisations in different PRs, to keep number of different open PRs small and avoid needing to rebase/merge.

Some other discussed optimisations:
* unsafe additions for scalar mul
* bounded scalar mul for KZG folding (not to use all bits)
* endomorphisms for in circuit scalar mul
* using precomputed lines (#925/#930). This should be almost drop-in fix without needing to modify this PR much.
* for proof aggregation extending native PLONK prover to have uniform verifying key (only needing to switch between selector polynomial commitments, keeping SRS and `Size` fixed)
* include #749 
* VDF for exponentiation

Ping @yelhousni for review.

yelhousni:
> Added MSM tests for emulated and 2-chain curves. This revealed another bug :)
> 
> I think the PR is done from my side. We could merge from my side after approving review. I think we can address additional optimisations in different PRs, to keep number of different open PRs small and avoid needing to rebase/merge.
> 
> Some other discussed optimisations:
> 
> * unsafe additions for scalar mul
> * bounded scalar mul for KZG folding (not to use all bits)
> * endomorphisms for in circuit scalar mul
> * using precomputed lines ([Perf: Groth16 verifier circuit with pre-computed lines #925](https://github.com/Consensys/gnark/pull/925)/[perf: use G2 precomputed lines for Miller loop #930](https://github.com/Consensys/gnark/pull/930)). This should be almost drop-in fix without needing to modify this PR much.
> * for proof aggregation extending native PLONK prover to have uniform verifying key (only needing to switch between selector polynomial commitments, keeping SRS and `Size` fixed)
> * include [perf: non-native modular multiplication #749](https://github.com/Consensys/gnark/pull/749)
> * VDF for exponentiation
> 
> Ping @yelhousni for review.

Alright on my end. Let's merge this and address optimizations in other PRs. I gave the Wesolowski VDF approach a try to compute efficiently `z^(2^T) mod n` but it's not sound since `n` a known prime (the root finding problem is easy). Also `2^T` is small in our use-case compared to `n` so reducing the exponent modulo `n-1` (the Euler totient) is not helpful neither. I would add to the list of optimizations:
- MSM in-circuit 
- For 2-chains, use twisted Edwards complete formulae for windowed scalar multiplication might be worth it.

yelhousni (approved):

ivokub:
Thanks for the review and co-authoring the PR! I gathered the ideas for improvements in #935.

PR #891, opened 2023-10-25
Feat/marshal g1 scalar

ThomasPiellard:
# Description

Adds method `MarshalG1(G1El, int) []frontend.Variable`, `MarshalScalar(S, int) []frontend.Variable` on the `Curve` interface, following gnark-crypto's Marshal (the result a slice of bits while it is a []byte in gnark-crypto but the layout is the same, same endianness).

## Type of change

# How has this been tested?

For both emulated and native curves, see:
- [x] TestMarshalG1
- [x] TestMarshalScalar

# Checklist:


ivokub (commented):
Right now the MarshalScalar and MarshalG1 methods seem to be returning bits. In the short-mimc the fieldhasher interfaces is taking in field elements.

I'll try if it would be possible to here instead return compatible fr.Elements (by recomposing bits such that the high byte is 0). Then we can also have compatibility when hashing the bits (not only bitwise compatibility with native marshaling)

ThomasPiellard:
Yes my plan was to return bits in the exact same layout as in gnark-crypto and let the user reconstruct the variable according to gnark-crypto. But it makes sense to return frontendVariables now that there is the short hash.

ivokub (approved):
I changed a bit not to include the number of bits for marshalling and are now deducing from the parameters. Also added bitmode to short hash so that we can have compatible hash values.

I'll implement custom FS transcript binding (with `BindBits` separately).

From my side good to go. Have a look yourself and you can merge if seems good.

PR #899, opened 2023-11-01
perf: lookup blueprint compile time improvement

### LINKED ISSUE #898: perf: lookup tables with many queries are slow to compile and solve
Large lookup tables with many queries would (greatly) benefit from some caching at the compile time and solving time.

gbotrel:
Partially fixes #898  (compile part).

Essentially, the `Blueprint` `WireWalker(inst Instruction) (w WireWalker, maxLevel int)` now returns a `maxLevel`  which, alongside the fact that the callback passed into the walker returns the `maxLevel` of a wire, allows stateful blueprints like `BlueprintLookupHint` to cache some info avoiding useless computation by the compiler to figure out the instruction dependency tree.

Added a reference benchmark in `logderivlookup`:
```
BenchmarkCompileManyLookup/scs-10      12830216291     480989306     -96.25%
BenchmarkCompileManyLookup/r1cs-10     13002215583     687194500     -94.71%
```


Tabaie (approved):

constraint/blueprint.go
```diff
@@ -18,11 +18,16 @@ type Blueprint interface {
...
+	// WireWalker returns a function that walks the wires appearing in the blueprint
+	// and optionally the minimum level (minus one) at which the instruction can be solved;
+	// that is; maxLevel = max{level(wire) | wire in blueprint}.
+	// If maxLevel is unknown at compile time, implementation must return 0.
 	// This is used by the level builder to build a dependency graph between instructions.
-	WireWalker(inst Instruction) func(cb func(wire uint32))
+	WireWalker(inst Instruction) (w WireWalker, maxLevel int)
 }
 
+type WireWalker func(cb func(wire uint32) int)
```
ivokub, line 29: I'm not understanding the return value of `cb` here - is it `cb`'s definition of wire level? Does it allow the blueprint to learn the level of the wire from the system's perspective?

ivokub: To confirm - essentially we use the return value of WireWalker and `maxLevel` to keep the blueprint and system in sync what is the max level of solved wires. And when we update the level in the blueprint side then we assume that the wires for which the callback is called on should not be called again, and we internally update the starting conditions of the loops.

ivokub (approved):

constraint/blueprint.go
```diff
@@ -18,11 +18,16 @@ type Blueprint interface {
...
 	NbOutputs(inst Instruction) int
 
-	// WireWalker returns a function that walks the wires appearing in the blueprint.
+	// WireWalker returns a function that walks the wires appearing in the blueprint
+	// and optionally the minimum level (minus one) at which the instruction can be solved;
+	// that is; maxLevel = max{level(wire) | wire in blueprint}.
+	// If maxLevel is unknown at compile time, implementation must return 0.
 	// This is used by the level builder to build a dependency graph between instructions.
-	WireWalker(inst Instruction) func(cb func(wire uint32))
+	WireWalker(inst Instruction) (w WireWalker, maxLevel int)
```
ivokub, line 26: So, for me it seems correct. One thing which bothers me a bit is that imo this change in lookup blueprint a bit breaks the contract - lookup blueprint does not actually walk all the wires anymore - just the new ones. Maybe we can add an argument to `Blueprint.WireWalker` method if we want to allow caching or not?

gbotrel:
@ivokub did replace the weird `WireWalker` with double callback pattern with a simpler `InstructionTree` interface (see `constraint/instruction_tree.go`) .

Yields further perf improvements (against previoous commit on this branch):
```
benchmark                              old ns/op     new ns/op     delta
BenchmarkCompileManyLookup/scs-10      480989306     412605000     -14.22%
BenchmarkCompileManyLookup/r1cs-10     687194500     676015438     -1.63%

benchmark                              old allocs     new allocs     delta
BenchmarkCompileManyLookup/scs-10      8603276        5336556        -37.97%
BenchmarkCompileManyLookup/r1cs-10     9624679        8598475        -10.66%

benchmark                              old bytes      new bytes      delta
BenchmarkCompileManyLookup/scs-10      942962434      846228050      -10.26%
BenchmarkCompileManyLookup/r1cs-10     2033656932     2002813188     -1.52%
```

ivokub (approved):
Now makes much more sense. Looks good! Dunno about having some checks behing debug tag - if these cases happen then it is a bug in gnark anyway, but I think it would be slightly easier to report and debug when having unique identifiers (instead of generic out-of-bounds err)

constraint/instruction_tree.go
```diff
@@ -0,0 +1,64 @@
...
+	// if wireID == maxUint32, it's a constant.
+	return (wireID - offset) < uint32(len(system.lbWireLevel))
+}
+
+func (system *System) GetWireLevel(wireID uint32) Level {
+	return system.lbWireLevel[wireID-system.internalWireOffset()]
+}
+
+func (system *System) InsertWire(wireID uint32, level Level) {
+	if debug.Debug {
```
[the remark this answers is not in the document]
gbotrel: ha I was not planning on leaving them -- but maybe the slight extra slowdown due to these sanity checks is acceptable. (when debug.Debug is set, this is removed by the compiler)

PR #900, opened 2023-11-02
feat: Fiat-Shamir transcript using a short hash

ivokub:
# Description

Implement gnark-crypto compatible Fiat-Shamir transcript using a short MiMC hash for proof recursion.

All tests seem to be working, but I'm not sure about the option `WithDomainSeparation`. I think it is too agressive on gnark-crypto side to enforce domain separation with the `string:` string when hasher has `WriteString` method. I would make it optional as in gnark. Imo right now the behaviour is not documented and quite difficult to debug.

Related #847 

## Type of change

# How has this been tested?

- [x] Transcript test with 2-chains
- [x] Transcript test with field emulation

Not benchmarked

# Checklist:

### DIFF
--- a/std/fiat-shamir/transcript.go
+++ b/std/fiat-shamir/transcript.go
@@ -53,7 +53,7 @@ type challenge struct {
 // NewTranscript returns a new transcript.
 // h is the hash function that is used to compute the challenges.
 // challenges are the name of the challenges. The order is important.
-func NewTranscript(api frontend.API, h hash.FieldHasher, challengesID ...string) Transcript {
+func NewTranscript(api frontend.API, h hash.FieldHasher, challengesID []string) *Transcript {
 	n := len(challengesID)
 	t := Transcript{
 		challenges: make(map[string]challenge, n),
@@ -65,7 +65,7 @@ func NewTranscript(api frontend.API, h hash.FieldHasher, challengesID ...string)
 		t.challenges[challengesID[i]] = challenge{position: i}
 	}
 
-	return t
+	return &t
 }
 
 // Bind binds the challenge to value. A challenge can be binded to an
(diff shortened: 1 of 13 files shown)


ivokub:
One existing Fiat-Shamir test still failing, looking into it.

ivokub:
Hmm, I think I'm still missing correct marshalling of zeros for emulated targets. Will add.

std/fiat-shamir/transcript.go
```diff
@@ -110,19 +119,34 @@ func (t *Transcript) ComputeChallenge(challengeID string) (frontend.Variable, er
 	t.h.Reset()
 
 	// write the challenge name, the purpose is to have a domain separator
-	cChallenge := []byte(challengeID) // if we send a string, it is assumed to be a base10 number
-	if challengeName, err := constant.HashedBytes(t.api, cChallenge); err == nil {
-		t.h.Write(challengeName)
+	challengeInput := []byte(challengeID)
+	var challengeHashInput frontend.Variable = challengeInput
+	if t.config.withDomainSeparation {
```
ThomasPiellard, line 124: So in gnark-crypto we will need to have a fiatShamirConfig also at some point to tune the domain separation variable ? Currently it's hardcoded in WriteString

ThomasPiellard (approved):

std/fiat-shamir/transcript.go
ivokub: Yes, I do think it would be better to be able to configure it in gnark-crypto also. Right now the domain separation is enforced automatically with a fixed string, but this would prevent (simple) compatiblity with other software. We could have it by wrapping MiMC in gnark-crypto to remove `WriteString` method (replacing it), but imo it is unintuitive. Maybe we can even do an option:
```go
func WithDomainSeparationString(prefix string)
```
or something like that.

ivokub: Created issue: https://github.com/Consensys/gnark-crypto/issues/467

PR #901, opened 2023-11-03
refactor: use emulated.FieldParams as type parameter to generic Curve and Pairing

ivokub:
# Description

For the PLONK and KZG verifiers we need to work on the scalar fields. However, in the existing parametrization we didn't have access to the field parameters and the 2-chain scalars weren't compatible.

This PR makes the generic `Curve` and `Pairing` interfaces parametric with `emulated.FieldParams` which defines the scalar field. Accordingly, we update all the implementations to follow the convention. As now the scalars are emulated elements, then we also defined 2-chain scalars to be emulated elements which is more conventional.

NB! This PR depends on #900 so that should be merged before.

## Type of change

# How has this been tested?

All existing tests pass as is.

Not benchmarked.

# Checklist:


ivokub:
Also @yelhousni - this also impact KZG verifier.

PR #906, opened 2023-11-07
Feat/kzg gadget full

ThomasPiellard:
# Description

Addition of `BatchVerifySinglePoint`, `BatchVerifyMultiPoints`, `FoldProof` in KZG gadget.

Will conflict with #900

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

- [x] TestBatchVerifySinglePoint
- [x] TestBatchVerifyMultiPoints
- [x] TestFoldProof 

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #908, opened 2023-11-08
feat: batched KZG

ivokub:
This PR is #906, but rebased on top of #900 and #901.


ivokub:
@ThomasPiellard set you as a reviewer for this PR. It is checked on my side and looks good to me. It doesn't differ significantly from #906, but I simplified a bit (for example the shorthash initialisation etc.). I still have some ideas on how to make the implementation a bit cleaner, but I would do it in other PRs to avoid creating merge conflicts with @yelhousni work.

However, it is rebased on top of #901, so we should merge that before this one.

ThomasPiellard (approved):

PR #912, opened 2023-11-13
Perf/multisymbol 4bw

Tabaie:
# Description

This implements an LZSS based compression scheme and the means to verify its correctness in a SNARK.

## Type of change


std/compress/io.go
gbotrel, line 1: Do we need this in this PR ?
Also we have a `Gcd` and `Pack` (circuit stuff) in this file

std/compress/lzss/compress.go
```diff
@@ -0,0 +1,218 @@
...
+	inputSa    [maxInputSize]int32 // suffix array space.
+
+	dictData  []byte
+	dictIndex *suffixarray.Index
+	dictSa    [maxDictSize]int32 // suffix array space.
+}
+
+type CompressionMode uint8
+
+const (
```
gbotrel, line 27: need doc to understand difference between "BestCompression" and "BestSnarkCompression" for example

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/consensys/gnark
 
-go 1.19
+go 1.21
```
gbotrel, line 3: need to keep to go 1.20 I think -- probably my bad wanted to use the (cough cough) awesome builtin "max" and "min" function

std/compress/lzss/dict_fun
gbotrel, line 1: remove

std/compress/lzss/compile_large/main.go
gbotrel, line 1: remove (can be in compress_test.go if needed, with a skip if testing.Short() is set)

std/compress/profile001.pdf
gbotrel, line 1: remove

std/compress/profile002.pdf
gbotrel, line 1: remove

std/utils/test_vectors_utils/test_vector_utils.go
gbotrel, line 1: why in this PR?

gbotrel (changes requested):

std/compress/profile002.pdf
gbotrel: + I think the compress/test_case folder can be cleaned up a bit

std/compress/lzss/compress.go
Tabaie: That's exactly why I think it's better to give it a name like `WordLength` that actually denotes its function, and then explain in docs that lower values increase the compression ratio and higher value decrease the decompressor's number of constraints.

frontend/cs/scs/api.go
```diff
@@ -657,6 +702,22 @@ func (builder *builder) Commit(v ...frontend.Variable) (frontend.Variable, error
 	})
 }
 
+// NewCombination in the form of res = qL.a + qR.b + qM.ab + qC
+func (builder *builder) NewCombination(a, b frontend.Variable, qL, qR, qM, qC int) frontend.Variable {
```
gbotrel, line 706: suggestion;
rename to `api.EvaluatePlonkExpression` + in this function, check that a, b are not constant and do the appropriate path if they are. (else this function will panic). 

also add a corresponding interface so that in a circuit, one can check if `_, ok := api.(...); ok` and have a fail over path for R1CS for example

std/compress/io.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"bytes"
+	"compress/gzip"
+	"github.com/consensys/gnark/frontend"
+	"golang.org/x/exp/constraints"
+	"io"
+	"os"
+)
+
+func GzWrite(outFileName string, o io.WriterTo) error {
```
gbotrel, line 12: why do we need to expose that exactly? if you only need it in test, put it iin _test.go files 👍 (but I don't understand from the code why we need it)

std/compress/io.go
```diff
@@ -0,0 +1,103 @@
...
+		for a[0] != 0 {
+			a[1], a[0] = a[0], a[1]%a[0]
+		}
+		a = a[1:]
+	}
+
+	return a[0]
+}
+
+func Pack(api frontend.API, words []frontend.Variable, wordLen int) []frontend.Variable {
```
gbotrel, line 89: should not be here

gbotrel (changes requested):

std/compress/io.go
Tabaie: Currently using it for code that outputs the compiled cs, which has to run as a standalone application and not a test.

PR #928, opened 2023-11-20
feat: exit when condition is not filled

ThomasPiellard:
# Description

The sanity checks in the plonk verifier (solidity) now return an error as soon as a condition is not matched instead of waiting for all checks to be done. It should save some gas when a condition is not matched (minor optim).

## Type of change

# Checklist:


ivokub (approved):
Otherwise, looks good. But now wouldn't it make more sense to use `gt(..., R_MOD-1)` instead of `iszero(lt(..., R_MOD))`? Maybe additional storage for `R_MOD-1` is bigger than additional opcode?

PR #874, opened 2023-10-19
Perf: KZG verify gadget

yelhousni:
# Description

**(Refactored as in #930 )**

The current KZG verify gadget computes (in `AssertProof`) the multi-pairing `e(totalG1, vk.SRS[0]) * e(negQuotientPoly, vk.SRS[1]) == 1` assuming all arguments are variable. However `vk.SRS[0]=G₂` and `vk.SRS[1]=[α]G₂` are fixed points in G2. This PR pre-computes the lines used in the multi-pairing hence avoiding G2 arithmetic in-circuit.

TODO: 
- [x] pre-compute lines from `vk.SRS[0]=G₂`
- [x] pre-compute lines from `vk.SRS[1]=[α]G₂`
- [x] implement `DoublePairingFixedQCheck` which does `e(var, fix) * e(var, var) == 1`
  - [x] BN254
  - [x] BLS12-381
  - [x] BW6-761 ~(Naive algorithm for now)~
  - [x] BLS12-377
  - [x] BLS24-315
- [x] implement `DoublePairingFixedQCheck` which does `e(var, fix) * e(var, fix) == 1`
- [x] make the fixed pairing agnostic/generic wrt fixed argument + lazy pre-computation
- [x] use new function in KZG verify gadget

N.B.: To do the same for BW6-761, we need to revisit the Miller loop algortihm to be an ate loop instead of a Tate. Will do in a separate PR.
*Edit: This was done in https://github.com/Consensys/gnark/pull/876 and merged here*

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

Same existing tests work for the new feature.

# How has this been benchmarked?

- *Emulated case:* verifying a KZG proof on BW6-761 with a BN254-SNARK:

|     | Old  | New |
|---|---|---|
|R1CS | 5,171,863 | 4,147,830 |
|SCS   | 49,547,799 | 38,776,383 |

This is a 10.7M+ saving in SCS which largely offsets the 24k loss in https://github.com/Consensys/gnark/pull/876/ due to Miller algorithm change.

-  *Native case:* verifying a KZG proof on BLS12-377 with a BW6-761-SNARK:

|     | Old  | New |
|---|---|---|
|R1CS | 18,390 | 16,414 |
|SCS   | 74,128 | 68,502 |

This is a 5.6k saving in SCS. A pairing in native already does not cost much.

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
[...]

PR #930, opened 2023-11-21
perf: use G2 precomputed lines for Miller loop

ivokub:
# Description

This PR adds G2 line precomputations as in #874, #925, but initialized lazily and embeded inside G2 element. This lazy computation allows to have all the benefits of using ML with precomputed lines, but without having to change the existing interfaces and allowing for some inputs be provided using lines and other as points (as needed in Groth16).

Additionally, added a way to fix the verification key completely inside the circuit. This provides less versatility than providing the verification key as a witness, but allows for considerable decrease in the number of public witness elements, which matters for Solidity verifier. The decrease of constraints due to fixing VK inside circuit is not substantial compared to providing precomputed lines as our non-native arithmetic is still not sufficiently flexible to handle all cases of operations with constant arguments.

This PR is still not complete - the precomputations and witness assignments are implemented only for BW6-761 currently as it is most critical for our use cases. But I think extending the approach to BLS12-377 first and then to other curves is fairly straightforward. It is not copy-paste though as we have different extension towers and optimisations for line computations.

## Type of change

# How has this been tested?

- [x] `TestPairFixedTestSolve` - Miller Loop with input G2 element with precomputation
- [x] `TestKZGVerificationEmulated3ConstantVk` - constant KZG verification key
- [x] `TestKZGVerificationEmulated3Precomputed` - witness KZG verification key with precomputation
- [x] `TestBW6InBN254` - witness G16 verification key without precomputation 
- [x] `TestBW6InBN254Precomputed` - witness G16 verification key with precomputation
- [x] `TestBW6InBN254Constant` - constant G16 verification key

All cases BW6 key emulated in BN254. R1CS constraint system. Decreasing order "no precomputation" -> "precomputed as witness" -> "precomputed as constant":
* KZG verification 5389344 -> 4087448 -> 4066748
* Groth16 verification 6086776 -> 4839928 -> 4748682

# Checklist:


ivokub:
I haven't run the full set of benchmarks, but I think the initial results are indicative. Verifying Groth16 over BW6 in BN254 emulated using R1CS I have the decrease from 6086776 constraints to 4748682. This is the case when the verifying key is fully constant.

Additionally, this construction allows for the case of providing precomputed lines as a witness. This is a bit more versatile, allowing to verify different circuits, but then would have to figure out how to provide precomputed lines verifiably. If we provide them as a secret witness, then the outer prover could cheat, but providing as public witness adds 18k elements to public witness. In that case the circuit size is 4839928.

Similarly, for KZG verification only have 5389344->4087448->4066748 constraints.

ivokub:
@yelhousni - I currently have implemented only for BW6-761 as imo this is most critical for PLONK compression BW6-761->BN254 (currently at 80M constraints without this PR). We could try to implement for all curves (second priority for BLS12-377), but I would rather try to merge this quickly and then do other curves separately in other PR.

backend/plonk/bls12-377/marshal.go
```diff
@@ -349,6 +349,7 @@ func (vk *VerifyingKey) writeTo(w io.Writer, options ...func(*curve.Encoder)) (n
 		&vk.Kzg.G1,
 		&vk.Kzg.G2[0],
 		&vk.Kzg.G2[1],
+		&vk.Kzg.Lines,
```
yelhousni, line 352: don't we need marshalling lines anymore? the out-circuit PLONK uses gnark-crypto KZG package which uses Lines, no? https://github.com/Consensys/gnark-crypto/blob/e77308824822cefa0bb066a9457eee0d426540ac/ecc/bls12-377/kzg/kzg.go#L247

std/algebra/emulated/sw_bw6761/pairing.go
```diff
@@ -255,114 +261,50 @@ func (pr Pairing) MillerLoop(P []*G1Affine, Q []*G2Affine) (*GTEl, error) {
...
-	// only care about the Miller loop result in Fp6 and not the point itself.
 	result = pr.Square(result)
 	for k := 0; k < n; k++ {
-		l0 = pr.tangentCompute(accQ[k])
-		l0 = &lineEvaluation{
-			R0: *pr.curveF.MulMod(&l0.R0, xNegOverY[k]),
-			R1: *pr.curveF.MulMod(&l0.R1, yInv[k]),
-		}
-		result = pr.MulBy014(result, &l0.R1, &l0.R0)
+		result = pr.MulBy014(result,
```
yelhousni, line 304: I think we can keep this iteration inside the loop because the special case is handled by `precomputeLines` (with `tangentCompute`)

yelhousni (approved):
This is a great refactoring! it takes away the necessity to understand the fixed-arg vs. variable-arg pairing implementation for a user. Great job! I think we can merge and then add the other curves.

std/algebra/emulated/sw_bw6761/pairing.go
ivokub: Merged the iterations. But as I'm not computing second line at all for `i=0`, then added additional conditional `i>0` here.

backend/plonk/bls12-377/marshal.go
ivokub: If I'm understanding correctly then you propose omitting Lines from PLONK verification key serialization/deserilization? I tried, but then the tests which check for structure stability after IO roundtrip, failed. In this PR we write the Lines when writing PLONK verification key (and also read when reading).

PR #933, opened 2023-11-23
Perf: variant of the Karabina cyclotomic squaring

yelhousni:
# Description

Currently we implement Theorem 3.2 from https://eprint.iacr.org/2010/542.pdf whenever we have 3 repeated squaring or more in the final exponentiation. It is the fastest variant of cyclotomic squaring but in-circuit the selector logic to handle branching cases when denominators are zero makes it not always the best in some cases. This PR compares different variants in Sec. 5 and proposes a tradeoff:
- For the native case, when repeated squaring size is:
    - 1 or 2 --> we choose Granger-Scott
    - 3 --> we choose Karabina SQR12345 (*implemented in this PR*)
    - 4 or more --> we choose Karabina SQR2345 (*implemented previously*)
So nothing changes concretely for BLS12-377 since we do not encouter the case of size 3.

- For emulated case, it difficult to theoretically set a threshold due to emulated arithmetic but empirically for BW6-761:
    - for sizes 1 and 2 --> we use Granger-Scott
    - for sizes 3, 4, 5, 7 and 11 --> we use Karabina SQR12345 (*implemented in this PR*)
    - for size 46 and 92 --> we use Karabina SQR2345 (*implemented previously*)

## Type of change

# How has this been tested?

`TestExptFp6` and `TestFinalExponentiationTestSolve` are used here to test the new Karabina variants. Otherwise we need to implement the go version of these in gnark-crypto.

For BW6-761, this saves 290826 SCS in the Final exp.

# Checklist:

### DIFF
--- a/std/algebra/emulated/fields_bw6761/e6.go
+++ b/std/algebra/emulated/fields_bw6761/e6.go
@@ -105,10 +105,103 @@ func (e Ext6) Square(x *E6) *E6 {
 	}
 }
 
-// Karabina's compressed cyclotomic square
+// Karabina's compressed cyclotomic square SQR12345
+// https://eprint.iacr.org/2010/542.pdf
+// Sec. 5.6 with minor modifications to fit our tower
+func (e Ext6) CyclotomicSquareKarabina12345(x *E6) *E6 {
+	x = e.Reduce(x)
+
+	// h4 = -g4 + 3((g3+g5)(g1+c*g2)-g1g5-c*g3g2)
+	g1g5 := e.fp.Mul(&x.B0.A1, &x.B1.A2)
+	g3g2 := e.fp.Mul(&x.B1.A0, &x.B0.A2)
+	h4 := mulFpByNonResidue(e.fp, &x.B0.A2)
+	h4 = e.fp.Add(h4, &x.B0.A1)
+	t := e.fp.Add(&x.B1.A0, &x.B1.A2)
+	h4 = e.fp.Mul(h4, t)
+	h4 = e.fp.Sub(h4, g1g5)
+	t = mulFpByNonResidue(e.fp, g3g2)
+	h4 = e.fp.Sub(h4, t)
+	h4 = e.fp.MulConst(h4, big.NewInt(3))
+	h4 = e.fp.Sub(h4, &x.B1.A1)
+
+	// h3 = 2(g3+3c*g1g5)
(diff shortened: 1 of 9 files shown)


std/algebra/emulated/fields_bw6761/e6.go
```diff
@@ -105,10 +105,103 @@ func (e Ext6) Square(x *E6) *E6 {
...
+	h3 = e.fp.Add(h3, &x.B1.A0)
+	h3 = e.fp.MulConst(h3, big.NewInt(2))
+
+	// h2 = 3((g1+g5)(g1+c*g5)-c*g1g5-g1g5)-2g2
+	t = mulFpByNonResidue(e.fp, &x.B1.A2)
+	t = e.fp.Add(t, &x.B0.A1)
+	h2 := e.fp.Add(&x.B1.A2, &x.B0.A1)
+	h2 = e.fp.Mul(h2, t)
+	h2 = e.fp.Sub(h2, g1g5)
+	t = mulFpByNonResidue(e.fp, g1g5)
```
ivokub, line 139: mulFpByNonResiude seems multiplication by -4 -- cannot we inline `-c*g1g5 - g1g5` as `-(c+1)g1g5`?

std/algebra/emulated/fields_bw6761/e6.go
```diff
@@ -105,10 +105,103 @@ func (e Ext6) Square(x *E6) *E6 {
...
+	t = e.fp.MulConst(&x.B0.A2, big.NewInt(2))
+	h2 = e.fp.Sub(h2, t)
+
+	// h1 = 3((g3+g2)(g3+c*g2)-c*g3g2-g3g2)-2g1
+	t = mulFpByNonResidue(e.fp, &x.B0.A2)
+	t = e.fp.Add(t, &x.B1.A0)
+	h1 := e.fp.Add(&x.B0.A2, &x.B1.A0)
+	h1 = e.fp.Mul(h1, t)
+	h1 = e.fp.Sub(h1, g3g2)
+	t = mulFpByNonResidue(e.fp, g3g2)
```
ivokub, line 151: Same as for `g1g5` before

ivokub (approved):

std/algebra/emulated/fields_bw6761/e6.go
yelhousni: absolutely :)

PR #940, opened 2023-12-01
Chore/compression v1

Tabaie:
# Description

Minor improvements to compression:

* Compressor can decide to bypass compression when appropriate
* One byte at the beginning of the compressed file is dedicated to the compression algorithm version

Fixes # (issue)

## Type of change

# How has this been tested?

All previous tests pass. No new tests were added.

No benchmarks were run as the change is minimal.

# Checklist:

### DIFF
--- a/frontend/cs/scs/api_assertions.go
+++ b/frontend/cs/scs/api_assertions.go
@@ -128,6 +128,33 @@ func (builder *builder) AssertIsBoolean(i1 frontend.Variable) {
 
 }
 
+func (builder *builder) AssertIsCrumb(i1 frontend.Variable) {
+	const errorMsg = "AssertIsCrumb: input is not a crumb"
+	if c, ok := builder.constantValue(i1); ok {
+		bytes := c.Bytes()
+		for i := 1; i < len(bytes); i++ {
+			if bytes[i] != 0 {
+				panic(errorMsg)
+			}
+		}
+		for i := 0; i < 4; i++ {
+			if bytes[0] == byte(i) {
+				return
+			}
+		}
+		panic(errorMsg)
+	}
+
+	x := builder.MulAcc(builder.Mul(-3, i1), i1, i1).(expr.Term)
+
+	builder.addPlonkConstraint(sparseR1C{
+		xa: x.VID,
+		xb: x.VID,
+		qL: builder.cs.FromInterface(2),
+		qM: builder.tOne,
+	})
+}
+
 // AssertIsLessOrEqual fails if  v > bound
 func (builder *builder) AssertIsLessOrEqual(v frontend.Variable, bound frontend.Variable) {
 	cv, vConst := builder.constantValue(v)
--- a/frontend/cs/scs/api_assertions.go
+++ b/frontend/cs/scs/api_assertions.go
@@ -131,16 +131,8 @@ func (builder *builder) AssertIsBoolean(i1 frontend.Variable) {
 func (builder *builder) AssertIsCrumb(i1 frontend.Variable) {
 	const errorMsg = "AssertIsCrumb: input is not a crumb"
 	if c, ok := builder.constantValue(i1); ok {
-		bytes := c.Bytes()
-		for i := 1; i < len(bytes); i++ {
-			if bytes[i] != 0 {
-				panic(errorMsg)
-			}
-		}
-		for i := 0; i < 4; i++ {
-			if bytes[0] == byte(i) {
-				return
-			}
+		if i, ok := builder.cs.Uint64(c); ok && i < 4 {
+			return
 		}
 		panic(errorMsg)
 	}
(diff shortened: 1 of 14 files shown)


frontend/cs/scs/api_assertions.go
```diff
@@ -128,6 +128,33 @@ func (builder *builder) AssertIsBoolean(i1 frontend.Variable) {
 
 }
 
+func (builder *builder) AssertIsCrumb(i1 frontend.Variable) {
+	const errorMsg = "AssertIsCrumb: input is not a crumb"
+	if c, ok := builder.constantValue(i1); ok {
+		bytes := c.Bytes()
```
gbotrel, line 134: I think doing something like 
```
if c.IsUint64() && c.Uint64() <= 3 {return} else {panic(...)} 
```
is more readable

frontend/cs/scs/api_assertions.go
```diff
@@ -128,6 +128,33 @@ func (builder *builder) AssertIsBoolean(i1 frontend.Variable) {
...
+		}
+		for i := 0; i < 4; i++ {
+			if bytes[0] == byte(i) {
+				return
+			}
+		}
+		panic(errorMsg)
+	}
+
+	x := builder.MulAcc(builder.Mul(-3, i1), i1, i1).(expr.Term)
```
gbotrel, line 148: can you comment a bit the logic? i.e. reading the code it's not obvious what is the constraint / if it's sound

std/compress/lzss/backref.go
```diff
@@ -47,22 +47,32 @@ type backref struct {
...
+}
+
 func (b *backref) writeTo(w *bitio.Writer, i int) {
-	w.TryWriteByte(b.bType.delimiter)
-	w.TryWriteBits(uint64(b.length-1), b.bType.nbBitsLength)
-	if b.bType.dictOnly {
-		w.TryWriteBits(uint64(b.address), b.bType.nbBitsAddress)
-	} else {
-		w.TryWriteBits(uint64(i-b.address-1), b.bType.nbBitsAddress)
+	panicIfErr(w.WriteByte(b.bType.delimiter))
```
gbotrel, line 57: why the change? The purpose of the `TryWrite...` methods is to make things more readable and test the error afterwards -- probably also has a perf impact over a large stream to do so many branches at each write

std/compress/lzss/compress.go
```diff
@@ -103,7 +103,7 @@ func (compressor *Compressor) Compress(d []byte) (c []byte, err error) {
 
 	// reset output buffer
 	compressor.buf.Reset()
-	compressor.buf.WriteByte(byte(compressor.level))
+	compressor.buf.Write([]byte{0, byte(compressor.level)}) // 0 -> compressor release version
```
gbotrel, line 106: can you extract in a `writeMetadata` or `writeHeader`/ corresponding `readMetadata` / `readHeader`?

std/compress/lzss/decompress.go
```diff
@@ -57,59 +65,49 @@ func DecompressGo(data, dict []byte) (d []byte, err error) {
...
-
-	bDict := backref{bType: dictBackRefType}
-	bShort := backref{bType: shortBackRefType}
-	bLong := backref{bType: longBackRefType}
-
+	_ = in.TryReadByte()
 	levelFromData := Level(in.TryReadByte())
-	if levelFromData != NoCompression && levelFromData != level {
-		panic("compression mode mismatch")
+	outLenBits := 16
```
gbotrel, line 77: what is outLen ?

std/compress/lzss/compress.go
```diff
@@ -210,6 +210,14 @@ func (compressor *Compressor) Compress(d []byte) (c []byte, err error) {
 		return nil, err
 	}
 
+	if compressor.buf.Len() >= len(d)+2 {
+		// compression was not worth it
+		out := compressor.buf.Bytes()
+		out[1] = byte(NoCompression)
```
gbotrel, line 216: use a clean `writeHeader` or `writeMetadata` func so the logic to read / write that is not dispersed

std/compress/stream.go
```diff
@@ -31,12 +31,21 @@ func (s *Stream) At(i int) int {
 	return s.D[i]
 }
 
-func NewStreamFromBytes(in []byte) Stream {
-	d := make([]int, len(in))
-	for i := range in {
-		d[i] = int(in[i])
+func panicIfErr(err error) {
+	if err != nil {
```
gbotrel, line 35: I would rather panics be real panics; if they are actually reachable due to a malformed input or something like that, it's a potential (unlikely but still ^^) ddos vector

frontend/cs/scs/api_assertions.go
gbotrel: + also interested to know the justification of this API (i.e clearly say; using the `std/math/bits` package generates X constraints while this generates Y constraint for R1CS, Z for SCS)

gbotrel (changes requested):

std/compress/lzss/backref.go
Tabaie: The errors overwrite each other so if we encounter an error in one of the middle writes (don't know how likely that is; to fail on one write and succeed on the following ones) we won't know that there ever was an error.

gbotrel (approved):

PR #953, opened 2023-12-06
refactor: plonk.Setup takes kzg srs in canonical and lagrange form

gbotrel:
This PR changes the signature of `plonk.Setup(...)` and adds an additional parameter; the `kzg.SRS` in Lagrange form.

In production, this can be computed using `gnark-crypto` `ToLagrangeG1` helper from a canonical SRS.

For test purposes when we know the `tau` used for generating the canonical srs, we can take a quicker path (see `test/unsafekzg` package which offers a convenient api to generate and cache in memory / on disk srs for test purposes).


test/unsafekzg/kzgsrs.go
```diff
@@ -0,0 +1,393 @@
...
+	bw6761 "github.com/consensys/gnark-crypto/ecc/bw6-761"
+)
+
+var (
+	cache           map[uint64]cacheEntry
+	memLock, fsLock sync.RWMutex
+)
+
+func init() {
+	cache = make(map[uint64]cacheEntry)
```
ivokub, line 60: Do not necessarily need init() function. Can just define global var `cache = make(map[uint64]cacheEntry)`. For example see https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/database/sql/sql.go;l=35

test/unsafekzg/kzgsrs.go
```diff
@@ -0,0 +1,393 @@
...
+	return canonical, lagrange, nil
+}
+
+type cacheEntry struct {
+	canonical kzg.SRS
+	lagrange  kzg.SRS
+}
+
+func cacheKey(curveID ecc.ID, size uint64) uint64 {
+	return uint64(curveID)<<59 | size
```
ivokub, line 126: I would rather use readable key names. When we cache to disk then it may be beneficial to have intelligible names for sharing etc. For example `kzg_srs-{{ CurveID }}-size_{{ size }}`.

internal/generator/backend/template/zkpschemes/plonk/plonk.setup.go.tmpl
```diff
@@ -89,7 +88,7 @@ type ProvingKey struct {
 }
 
 // TODO modify the signature to receive the SRS in Lagrange form (optional argument ?)
-func Setup(spr *cs.SparseR1CS, kzgSrs kzg.SRS) (*ProvingKey, *VerifyingKey, error) {
+func Setup(spr *cs.SparseR1CS, srs, srsLagrange kzg.SRS) (*ProvingKey, *VerifyingKey, error) {
```
ivokub, line 91: Maybe if `srsLagrange == nil` we compute it inside Setup?

The function documentation doesn't reflect anymore that also have to provide srs in Lagrange form or how to get it.

ivokub (approved):
Looks good. A few comments though but they are not blocking.

ivokub:
Imo on thing we could consider in the future is to have a lazy SRS. We can store it on disk and only when we actually we read it to memory (using mmap). Or more better, because the key is very structured we can have method a la `Get(index int)` and then we read exactly from the location we need by computing the exact byte location on disk.

PR #968, opened 2023-12-15
Fix std/recursion/plonk native and emulated examples

wzmuda:
# Description

While studying in-circuit verification with PLONK, I found the existing examples to be of great help. During my research, I discovered some flaws in them and decided to correct these issues for others who might evaluate the same feature in the future.

~Fixes # (issue)~ Currently, there is no dedicated ticket reporting this problem. This PR is primarily a convenience change that does not alter the actual library code. Therefore, I did not think it was necessary to open a ticket. However, please let me know if doing so is required; I would be happy to create one and update this paragraph accordingly.

## Type of change

# How has this been tested?

These examples are not actual Golang tests, but converting them for such use is straightforward. What I did was simply rename `func Example_emulated()` to `func Example_emulated(t *testing.T)` (and did the same for the other example). This change is sufficient to transform them into Golang-style test cases and run them accordingly.

One could also create a separate project and copy the code, which is a reasonable first step for anyone developing their own program with in-circuit PLONK verification.

Since this is just an example, I did not perform any benchmarking.

# Checklist:

No, but this does not seem relevant since the examples serve, to some extent, as tests themselves.

No, but since I'm modifying only examples, this does not seem to be relevant.

### DIFF
--- a/std/recursion/plonk/native_doc_test.go
+++ b/std/recursion/plonk/native_doc_test.go
@@ -10,10 +10,12 @@ import (
 	"github.com/consensys/gnark/test/unsafekzg"
 )
 
-// Example of verifying recursively BLS12-371 PLONK proof in BW6-761 PLONK circuit using field emulation
+// Example of verifying recursively BLS12-377 PLONK proof in BW6-761 PLONK circuit using field emulation
 func Example_native() {
 	// compute the proof which we want to verify recursively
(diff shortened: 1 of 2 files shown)


ivokub:
Hi @wzmuda - thanks for the contribution. I have been busy before holidays to review but will definitely get to it next year. Sorry for the delay.

ivokub (approved):
Perfect! Thanks for the contribution!

I wrote the buggy examples and it is good that they are really being read!

PR #976, opened 2023-12-20
perf(ecmul): use GLV with safe handling of edge cases in EVM ecmul

yelhousni:
# Description

Another tiny PR to use GLV with safe handling of edge cases in EVM `ecmul` precompile.

## Type of change

# How has this been tested?

Tests in `bn_test.go` pass.
 
ECMUL precompile cost goes from **781,437** to **673,659** SCS.

# Checklist:

### DIFF
--- a/std/algebra/emulated/sw_emulated/point.go
+++ b/std/algebra/emulated/sw_emulated/point.go
@@ -480,13 +480,18 @@ func (c *Curve[B, S]) scalarMulGLV(Q *AffinePoint[B], s *emulated.Element[S]) *A
 	if err != nil {
 		panic(fmt.Sprintf("compute GLV decomposition: %v", err))
 	}
-	s1, s2 := sd[0], sd[1]
+	s1, s2, s3, s4 := sd[0], sd[1], sd[3], sd[4]
+
 	c.scalarApi.AssertIsEqual(
 		c.scalarApi.Add(s1, c.scalarApi.Mul(s2, c.eigenvalue)),
 		c.scalarApi.Add(s, c.scalarApi.Mul(frModulus, sd[2])),
 	)
-	selector1 := c.scalarApi.IsZero(c.scalarApi.Sub(sd[3], s1))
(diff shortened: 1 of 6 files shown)


yelhousni:
@ivokub just re-pinging you here

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -532,13 +540,157 @@ func (c *Curve[B, S]) scalarMulGLV(Q *AffinePoint[B], s *emulated.Element[S], op
...
+		R0 = c.Select(sBits[i], Rb, R0)
+		R1 = c.Select(sBits[i], R1, Rb)
+	}
+
+	// i = n-1
+	Rb = c.doubleAndAddSelect(sBits[n-1], R0, R1)
+	R0 = c.Select(sBits[n-1], Rb, R0)
+
+	// i = 0
+	R0 = c.Select(sBits[0], R0, c.Add(R0, c.Neg(p)))
```
ivokub, line 579: Do I understand correctly that here is the only difference for safe and unsafe? I feel the optional `Safe` option would make sense as we would only be changing one function. Then we can have less duplication.

ivokub (approved):
PR looks good. I'll accept it, but maybe I can try implementing the safe scalarmul with option instead to see how it looks. I have actually also thought about adding an option which would allow to define for MSM which inputs may be unsafe (i.e. 0 scalar or 0 point) and then these options would complement each other quite nicely.

ivokub:
@yelhousni - I implemented using an option and merged the implementations. See how it looks and feels. If seems excessive then I'm good with reverting. On one hand I do not want to overuse using options, but I still like having unified implementation.

Another thing - currently the default is to use unsafe and we use option to use safe version. But I guess it would make more sense to have the defaults reversed -- i.e. without options we use safe formulas and then can optionally toggle to unsafe formulas for performance. I think we may have incoming bug reports due to the defaults being the other way.
But I didn't want to change now as it would have required a lot of refactoring all over, so dunno.

PR #977, opened 2023-12-20
perf: mark the result of `builder.IsZero` as boolean to save constraints when used in future

winderica:
# Description

Hi there, thanks for your amazing work!

In the current implementation of gnark, `builder.Select` introduces a constraint to enforce that the condition is boolean. However, when the condition is from the result of `builder.IsZero`, this constraint is unnecessary, as `builder.IsZero` always returns a value `m` that is guaranteed to be either 0 or 1.

This PR marks `m` in `builder.IsZero` as boolean, so that we can save 1 constraint when using the result in subsequent operations like `Select`.

## Type of change

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas (N/A)
- [ ] I have made corresponding changes to the documentation (N/A)
- [ ] I have added tests that prove my fix is effective or that my feature works (N/A)
- [ ] Any dependent changes have been merged and published in downstream modules (N/A)

### DIFF
--- a/frontend/cs/r1cs/api.go
+++ b/frontend/cs/r1cs/api.go
@@ -565,6 +565,8 @@ func (builder *builder) IsZero(i1 frontend.Variable) frontend.Variable {
 
 	builder.cs.AttachDebugInfo(debug, []int{c1, c2})
 
+	builder.MarkBoolean(m)
+
 	return m
 }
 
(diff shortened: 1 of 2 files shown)


ivokub:
Hi @winderica - thanks for the contribution. I have been busy before holidays to review but will definitely get to it next year. Sorry for the delay.

I think the fix would be useful but I would first have to see thoroughly the test suite to see if there are not unintended side-effects.

gbotrel (approved):
lgtm 👍

Issue #966, opened 2023-12-15
feat: allow arbitrary circuit verification in PLONK aggregation


ivokub:
See #960.

Currently they are constants for making the circuits a bit more efficient. Actually `SizeInv` being a constant doesn't improve the constraint count as we do not have short path for computing var-const multiplication in field emulation. But `Size` currently defines the number of iterations we need for exponentiating `zeta`, the challenge used inside the circuit.

If we would make `Size` variable and fix it size to be approx 30 bits, then the overhead for additional multiplications is relatively small compared to rest of the operations (MSMs, pairing computation, hashing etc.).

Secondly, right now we are using `Lookup2` for switching between the verification keys which means we can only switch between four different keys. This is somewhat limiting. It would be better if we would use a proper lookup table for key switching, but this requires making the lookup table generic for the input type (or implement it for emulated elements and extend to G1).

Or maybe we can have even better approach where we store all verification keys inside a Merkle Tree, provide the root as an input to the aggregation circuit and then use inclusion proof to verify that all verification keys are inside Merkle Tree. I guess we can have several approaches.

PR #1010, opened 2024-01-20
feat: different PLONK circuit verification

ivokub:
# Description

Currently we didn't switch on the circuit size and generator when doing multi-proof verification. This PR includes these parts in PLONK verification key in circuit specific VK. Additionally, had to change the exponentiation by the circuit size to use variable instead of a constant.

Related #966 

## Type of change

# How has this been tested?

All current tests work. 

Previously verifying over 10 proofs cost was 291527, now 313833, so approx 7% increase. BUT, for the example circuits the circuits we were aggregating were small and we did fixed exponetiation, so only essentially a few non-native muls. Now we do variable exponentiation for 30-bit exponents. But circuits in practice are actually closer to 30 bits (25-26 bits imo), so the difference shouldn't be that significant.

# Checklist:

### DIFF
--- a/std/recursion/plonk/doc.go
+++ b/std/recursion/plonk/doc.go
@@ -1,2 +1,5 @@
 // Package plonk implements in-circuit PLONK verifier.
+//
+// NB! The circuit allows verifying proofs of PLONK circuits of size up to 2**30
+// constraints.
 package plonk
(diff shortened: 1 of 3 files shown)


yelhousni (approved):
LGTM 👍 Maybe add a comment in `plonk/doc.go` to say that we assume circuits of maximum size 2**30.

ivokub:
> LGTM 👍 Maybe add a comment in `plonk/doc.go` to say that we assume circuits of maximum size 2**30.

Thanks, added!

PR #1017, opened 2024-01-22
feat: use n-bit mux for switching PLONK verification keys

### LINKED ISSUE #966: feat: allow arbitrary circuit verification in PLONK aggregation

ivokub:
# Description

This PR changes the mux from `Lookup2` to `selector.Mux` which allows arbitrary switching. The change is backwards compatible (previously we threw an error if had more than 4 keys).

Fixes #966 

## Type of change

# How has this been tested?

- [x] TestMux for non-native elements
- [x] TestMux for G1 elements (native and 2-chains)
- [x] Implemented PLONK aggregate verification with 5 keys instead of four.

PLONK recursion over 10 proofs increase from 313833 to 314402.

# Checklist:

### DIFF
--- a/std/recursion/plonk/verifier.go
+++ b/std/recursion/plonk/verifier.go
@@ -2,6 +2,7 @@ package plonk
 
 import (
 	"fmt"
+	"reflect"
 
 	fr_bls12377 "github.com/consensys/gnark-crypto/ecc/bls12-377/fr"
 	fr_bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381/fr"
@@ -28,6 +29,7 @@ import (
 	"github.com/consensys/gnark/std/math/bits"
 	"github.com/consensys/gnark/std/math/emulated"
 	"github.com/consensys/gnark/std/recursion"
+	"github.com/consensys/gnark/std/selector"
 )
 
 // Proof is a typed PLONK proof of SNARK. Use [ValueOfProof] to initialize the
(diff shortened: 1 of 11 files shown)


ivokub:
@gbotrel - included you for review as I'm doing some reflect stuff to switch over all fields in verification key.

std/recursion/plonk/verifier.go
```diff
@@ -1242,33 +1241,128 @@ func (v *Verifier[FR, G1El, G2El, GtEl]) SwitchVerificationKey(bvk BaseVerifying
...
-	}
+	v.muxCircuitVerifyingKey(idx, cvks, &cvk)
 	return VerifyingKey[FR, G1El, G2El]{
 		BaseVerifyingKey:    bvk,
 		CircuitVerifyingKey: cvk,
 	}, nil
 }
+
+func (v *Verifier[FR, G1El, G2El, GtEl]) muxCircuitVerifyingKey(idx frontend.Variable, cvks []CircuitVerifyingKey[FR, G1El], ret *CircuitVerifyingKey[FR, G1El]) {
+	var cvkT CircuitVerifyingKey[FR, G1El]
```
gbotrel, line 1256: so, discovering this, I'm a bit confused as to what is the purpose of the method and why `reflect` is a good choice here. Can you add couple of lines in comments to contextualise this?

gbotrel (approved):
high level tests looks decent to me, not sure about the specifics of the `f.enforceWidthConditional` and `normalize` each time I read `emulated` package I'm confused as to when to / when not to use them TBH.

ivokub:
> high level tests looks decent to me, not sure about the specifics of the `f.enforceWidthConditional` and `normalize` each time I read `emulated` package I'm confused as to when to / when not to use them TBH.

Regarding `f.enforceWidthConditional` - it is essentially to assert that witness limbs are in range. If it is already constrained the it is no-op, otherwise we range check.

I need normalize, as the the inputs may have different number of limbs or overflow. So the result needs to be with maximum number of limbs and overflow for further correct computations. I only need to use normalize for Select, Lookup2 and Mux.

PR #1022, opened 2024-01-26
fix: Decompressor to return -1 when output doesn't fit

Tabaie:
# Description

Currently, when the output buffer `d` is not long enough to contain the decompressed stream, `Decompress` returns `dLength = 0` which is seen as a valid output. This PR changes that to -1 which in any cryptographic prime-order field is so large it cannot be confused with a valid output length.

## Type of change

# How has this been tested?

[`TestOutBufTooShort`](https://github.com/Consensys/gnark/blob/fix/compressor/truncated-input/std/compress/lzss/snark_test.go#L106)

No benchmark. Performance impact expected to be negligible.

# Checklist:


AlexandreBelling (approved):
Looking good to me :)

PR #1023, opened 2024-01-26
Fix: edge cases in `std/algebra` elliptic curve arithmetic circuit (`emulated` and `2-chains)`

### LINKED ISSUE #964: fix: PLONK verifier gadget in-circuit without commitments
From #949.

The issue is that if we do not use commitments then verification key points may be zero. But for perf we use unsafe operations. 

> I would still like to get PLONK verification without commitment working. I'm actually thinking about adding new algopt option algopts.ForceSafe which performs safe arithmetic inside MSM. It is suboptimal (we could also only omit the selector we know is 0), but atleast would cover different cases. And I think we already know inside the circuit if there is commitment or not.

yelhousni:
# Description

- Fix edge cases in `std/algebra` involving scalars `0` and points `(0,0)` using the `UseSafe` option.
- Fix bug when scalar is `1` in `scalarMulGeneric`, `scalarBitsMulGeneric` and related methods.
- Fix #964 (but the fix is commented out to save constraints for the cases with `appi.Commit` that we use in Linea)

## Type of change

# How has this been tested?

- Current tests
- New tests for missing methods
- New tests taking into accounts all edge cases

Constraints count doesn't change when `UseSafe` option is not used (except for `scalarMulGeneric`), which is what interests us the most for our use-cases. The option is to capture all edge cases in all use-cases. 
Best thing forward is to add relevant emulated circuits to `internal/stats`.

# Checklist:


std/algebra/native/sw_bls12377/g1.go
```diff
@@ -405,7 +323,15 @@ func (P *G1Affine) constScalarMul(api frontend.API, Q G1Affine, s *big.Int) *G1A
 	s.Mod(s, cc.fr)
 	cc.phi1(api, &phiQ, &Q)
 
-	k := ecc.SplitScalar(s, cc.glvBasis)
+	var k [2]big.Int
+	// if s=0, assign dummy 1s to k[0] and k[1]
+	if s.BitLen() == 0 {
```
ivokub, line 328: In constant case we can just return infinity point already here without needing to record any constraints. And similarly for s == 1 we can return Q directly. And in this case I would move the check before we perform any computations a few lines up (cc.phi1 and s.Mod)

std/algebra/native/sw_bls12377/g1.go
```diff
@@ -422,47 +348,64 @@ func (P *G1Affine) constScalarMul(api frontend.API, Q G1Affine, s *big.Int) *G1A
...
-}
-
-// Assign a value to self (witness assignment)
-func (p *G1Jac) Assign(p1 *bls12377.G1Jac) {
-	p.X = (fr.Element)(p1.X)
-	p.Y = (fr.Element)(p1.Y)
-	p.Z = (fr.Element)(p1.Z)
-}
+	// if s=0, return P=(0,0)
+	if s.BitLen() == 0 {
```
ivokub, line 403: Assuming we already check `s == 0` above then do not need this check here.

std/algebra/native/sw_bls12377/g2.go
```diff
@@ -366,7 +333,15 @@ func (P *g2AffP) constScalarMul(api frontend.API, Q g2AffP, s *big.Int) *g2AffP
 	s.Mod(s, cc.fr)
 	cc.phi2(api, &phiQ, &Q)
 
-	k := ecc.SplitScalar(s, cc.glvBasis)
+	var k [2]big.Int
+	// if s=0, assign dummy 1s to k[0] and k[1]
+	if s.BitLen() == 0 {
```
ivokub, line 338: Can directly return infinity point here if s == 0 and Q if s==1

std/algebra/native/sw_bls12377/g2.go
```diff
@@ -383,47 +358,65 @@ func (P *g2AffP) constScalarMul(api frontend.API, Q g2AffP, s *big.Int) *g2AffP
...
-
-// Assign a value to self (witness assignment)
-func (p *G2Jac) Assign(p1 *bls12377.G2Jac) {
-	p.X.Assign(&p1.X)
-	p.Y.Assign(&p1.Y)
-	p.Z.Assign(&p1.Z)
-}
+	// if s=0, return P=(0,0)
+	zero := fields_bls12377.E2{A0: 0, A1: 0}
+	if s.BitLen() == 0 {
```
ivokub, line 414: Can remove if applying short path above

std/algebra/native/sw_bls24315/g1.go
```diff
@@ -389,7 +322,15 @@ func (P *G1Affine) constScalarMul(api frontend.API, Q G1Affine, s *big.Int) *G1A
 	s.Mod(s, cc.fr)
 	cc.phi1(api, &phiQ, &Q)
 
-	k := ecc.SplitScalar(s, cc.glvBasis)
+	var k [2]big.Int
+	// if s=0, assign dummy 1s to k[0] and k[1]
+	if s.BitLen() == 0 {
```
ivokub, line 327: Use fast path return

std/algebra/native/sw_bls24315/g2.go
```diff
@@ -366,7 +333,15 @@ func (P *g2AffP) constScalarMul(api frontend.API, Q g2AffP, s *big.Int) *g2AffP
 	s.Mod(s, cc.fr)
 	cc.phi2(api, &phiQ, &Q)
 
-	k := ecc.SplitScalar(s, cc.glvBasis)
+	var k [2]big.Int
+	// if s=0, assign dummy 1s to k[0] and k[1]
+	if s.BitLen() == 0 {
```
ivokub, line 338: Add fast path return

std/recursion/plonk/verifier.go
```diff
@@ -1002,11 +1002,11 @@ func (v *Verifier[FR, G1El, G2El, GtEl]) PrepareVerification(vk VerifyingKey[FR,
 		_s1, _s2, // second & third part
 	)
 
-	linearizedPolynomialDigest, err := v.curve.MultiScalarMul(points, scalars)
+	linearizedPolynomialDigest, err := v.curve.MultiScalarMul(points, scalars /*, algopts.WithUseSafe()*/) // in PLONK Wo Commit ==> use algopts.WithUseSafe()
```
ivokub, line 1005: Maybe we can implement PLONK verifier option for enabling `WithUseSafe` here?

std/recursion/plonk/verifier.go
```diff
@@ -1002,11 +1002,11 @@ func (v *Verifier[FR, G1El, G2El, GtEl]) PrepareVerification(vk VerifyingKey[FR,
...
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("linearized polynomial digest MSM: %w", err)
 	}
-	linearizedPolynomialDigest = v.curve.Add(linearizedPolynomialDigest, &vk.Qk.G1El) // Qk=0 in PLONK W\ Commit ==> use AddUnified
+	linearizedPolynomialDigest = v.curve. /*AddUnified*/ Add(linearizedPolynomialDigest, &vk.Qk.G1El) // in PLONK Wo Commit ==> use AddUnified
```
ivokub, line 1009: Same

std/recursion/plonk/verifier_test.go
```diff
@@ -42,10 +42,12 @@ func (c *OuterCircuit[FR, G1El, G2El, GtEl]) Define(api frontend.API) error {
 }
 
 /*
-TODO: Tests without api.Commit fail because the current optimized MultiScalarMul (JointScalarMul in particular) requires input points to be distinct.
-
-//-----------------------------------------------------------------
+///-----------------------------------------------------------------
```
ivokub, line 45: If we have PLONK verifier option then can enable the test to serve as an example? I agree that by default should use unsafe MSM and Add in verifier for efficinecy, the edge cases happen only for very trivial inner circuits.

ivokub (changes requested):
Looks good, however I think for scalarmul by constant we can have fast paths which do not perform computations.

Dunno about PLONK verifier options though - I think they are good to have but not essential. I think we can implement them later (but creating an issue for tracking), it is up to you.

std/algebra/native/sw_bls12377/g1.go
yelhousni: Yes you're right. Will do.

std/recursion/plonk/verifier.go
yelhousni: Yes that would be the best thing we can do

std/recursion/plonk/verifier_test.go
yelhousni: Agree

yelhousni:
> Looks good, however I think for scalarmul by constant we can have fast paths which do not perform computations.
> 
> Dunno about PLONK verifier options though - I think they are good to have but not essential. I think we can implement them later (but creating an issue for tracking), it is up to you.

Applied fast path for `constScalarMul` when `s==0`. For the case `s==1` I didn't do it since we never use this method anyway we can also specialize paths for `s==2` and so on.. but `s==0` is just to have the exposed method working for all cases even if it's never used in our use-cases. For the plonk option I will create an issue and close this PR. Wdyt?

ivokub (approved):
> Applied fast path for constScalarMul when s==0. For the case s==1 I didn't do it since we never use this method anyway and we can also specialize paths for s==2 and so on.. but s==0 is just to have the exposed method working for all cases even if it's never used in our use-cases. For the plonk option I will create an issue and close this PR. Wdyt?

Yep, I agree - lets only handle `s==0` in the fast path and everything else in regular path. Your changes look good 👍 !

And I also agree with handling PLONK verifier options in separate PR. It is nice to have but nothing critical.

PR #1028, opened 2024-01-30
Feat/plonk verifier options

### LINKED ISSUE #1025: Add PLONK verifier option to use safe EC arithmetic.
From https://github.com/Consensys/gnark/pull/1023#discussion_r1468257308.

When PLONK verifier recursion circuit doesn't use `api.Commit`, some points in the verifying key are not set, so `(0,0)` by default. In this case we need to use `algopts.WithUseSafe()` option [here](https://github.com/Consensys/gnark/blob/f3c3eebfc1bdd28aeb90135472a24e4e8c1713fd/std/recursion/plonk/verifier.go#L1005
) and `AddUnified` [here](https://github.com/Consensys/gnark/blob/f3c3eebfc1bdd28aeb90135472a24e4e8c1713fd/std/recursion/plonk/verifier.go#L1009
). We can add an option to the PLONK verifier to handle this.

ivokub:
# Description

Adds PLONK verifier options

Fixes #1025 

## Type of change

# How has this been tested?

PLONK recursive verifier without commitments added and succeeding.

Not benchmarked

# Checklist:

### DIFF
--- a/std/recursion/plonk/opts.go
+++ b/std/recursion/plonk/opts.go
@@ -72,3 +72,34 @@ func GetNativeVerifierOptions(outer, field *big.Int) backend.VerifierOption {
 		return nil
 	}
 }
+
+type verifierCfg struct {
+	withUseSafe bool
+}
+
+// VerifierOption allows to modify the behaviour of PLONK verifier.
+type VerifierOption func(cfg *verifierCfg) error
+
+// WithUseSafe forces the usage of safe formulas for point addition and
+// multi-scalar multiplication. The option is necessary when recursing simple
+// inner circuits whose selector polynomials may have exceptional cases (zeros,
+// equal to each other, inverses of each other).
+//
+// Safe formulas are less efficient to use, so using this option has performance
+// impact on the outer circuit size.
+func WithUseSafe() VerifierOption {
+	return func(cfg *verifierCfg) error {
+		cfg.withUseSafe = true
+		return nil
+	}
+}
+
+func newCfg(opts ...VerifierOption) (*verifierCfg, error) {
+	cfg := new(verifierCfg)
+	for i := range opts {
+		if err := opts[i](cfg); err != nil {
+			return nil, fmt.Errorf("option %d: %w", i, err)
+		}
+	}
+	return cfg, nil
+}
--- a/std/recursion/plonk/verifier_test.go
+++ b/std/recursion/plonk/verifier_test.go
@@ -37,11 +37,10 @@ func (c *OuterCircuit[FR, G1El, G2El, GtEl]) Define(api frontend.API) error {
 	if err != nil {
 		return fmt.Errorf("new verifier: %w", err)
 	}
-	err = verifier.AssertProof(c.VerifyingKey, c.Proof, c.InnerWitness)
+	err = verifier.AssertProof(c.VerifyingKey, c.Proof, c.InnerWitness, WithUseSafe())
 	return err
 }
 
-/*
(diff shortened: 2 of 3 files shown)


std/recursion/plonk/opts.go
```diff
@@ -72,3 +72,34 @@ func GetNativeVerifierOptions(outer, field *big.Int) backend.VerifierOption {
...
+type VerifierOption func(cfg *verifierCfg) error
+
+// WithUseSafe forces the usage of safe formulas for point addition and
+// multi-scalar multiplication. The option is necessary when recursing simple
+// inner circuits whose selector polynomials may have exceptional cases (zeros,
+// equal to each other, inverses of each other).
+//
+// Safe formulas are less efficient to use, so using this option has performance
+// impact on the outer circuit size.
+func WithUseSafe() VerifierOption {
```
gbotrel, line 90: I prefer `WithSafeFormulas` or `WithSafeArithmetic` or `WithCompleteFormulas` -- 

doc wise; as a circuit developer, how do I know if my inner circuit have selector polynomial with exceptional cases? 
Shouldn't we put the "safe arithmetic formulas" as default, and have an option that disable it when we know what we are doing and don't expect such edge cases??

gbotrel (changes requested):

std/recursion/plonk/opts.go
ivokub: Yup, we thought about with @yelhousni and it seems that doing unsafe by default was better. The exceptional cases happen when:
* Qk = 0
* Ql = Qr
* Ql = -Qr
* (a few more related to relations of the selector polynomial commitments)
* batchedproofs claimed values are 0-s

These cases usually happen quite rarely when we generate dummy internal circuits for testing or when the proofs are malformed. The option is rather for the cases when we test to avoid needing to patch gnark.

cc: @yelhousni, @ThomasPiellard to confirm my intuition?

ivokub: re renaming the option - indeed, I'm not very good at naming things. I'll rename.

yelhousni: right edge-cases should be trivial and I prefer to have the unsafe path as default because some of the decisions are only related to Ethereum (e.g. `(0,0)` is not the infinity point normally). Also making the safe default adds more constraints to the unsafe path, at least in the way we're are handling it right now.

ivokub: Renamed 👍

std/recursion/plonk/verifier_test.go
```diff
@@ -37,11 +37,10 @@ func (c *OuterCircuit[FR, G1El, G2El, GtEl]) Define(api frontend.API) error {
...
 	}
-	err = verifier.AssertProof(c.VerifyingKey, c.Proof, c.InnerWitness)
+	err = verifier.AssertProof(c.VerifyingKey, c.Proof, c.InnerWitness, WithCompleteArithmetic())
 	return err
 }
 
-/*
 ///-----------------------------------------------------------------
 // Without api.Commit
 //
```
yelhousni, line 46: I guess we should remove this comment now

yelhousni (approved):
LGTM 👍

std/recursion/plonk/verifier_test.go
yelhousni: @ivokub removed it in last commit + renamed option in `std/algebra`

ivokub:
@gbotrel, request for re-review. Merging is blocked otherwise.

gbotrel (approved):

PR #1033, opened 2024-02-04
feat: test engine is a range checker

Tabaie:
This PR fits the test engine to the `frontend.Rangechecker` interface, with the effect that now its range assertions are checked during the `Define` method, giving a clearer error in case of failure as compared with a failed deferred call.

# How has this been tested?

Previous rangecheck tests pass. A [new test](https://github.com/Consensys/gnark/blob/feat/engine-rangechecker/test/engine_test.go#L109) has also been added but I am not sure if it's necessary.

No benchmarking.

# Checklist:

### DIFF
--- a/test/engine_test.go
+++ b/test/engine_test.go
@@ -2,6 +2,8 @@ package test
 
 import (
 	"fmt"
+	"github.com/consensys/gnark/backend"
+	"github.com/consensys/gnark/std/rangecheck"
 	"math/big"
 	"testing"
 
@@ -100,4 +102,29 @@ func TestPreCompileHook(t *testing.T) {
 	if !isDeferCalled {
 		t.Error("callback not called")
 	}
+}
+
+// TODO @ivokob is this test necessary, considering that the rangechecker is already tested against the test engine?
+
+func TestRangeCheck(t *testing.T) {
+	NewAssert(t).CheckCircuit(&rangeCheckTestCircuit{bits: 4}, WithCurves(ecc.BN254), WithBackends(backend.PLONK),
+		WithValidAssignment(&rangeCheckTestCircuit{X: 0}),
+		WithValidAssignment(&rangeCheckTestCircuit{X: 15}),
+		WithInvalidAssignment(&rangeCheckTestCircuit{X: 16}),
+		WithInvalidAssignment(&rangeCheckTestCircuit{X: -1}),
+	)
+
+	NewAssert(t).CheckCircuit(&rangeCheckTestCircuit{bits: 254}, WithCurves(ecc.BN254), WithBackends(backend.PLONK),
+		WithValidAssignment(&rangeCheckTestCircuit{X: -1}),
+	)
+}
+
+type rangeCheckTestCircuit struct {
+	X    frontend.Variable
+	bits int
+}
+
+func (c *rangeCheckTestCircuit) Define(api frontend.API) error {
+	rangecheck.New(api).Check(c.X, c.bits)
+	return nil
 }
--- a/test/engine_test.go
+++ b/test/engine_test.go
@@ -125,6 +125,15 @@ type rangeCheckTestCircuit struct {
 }
 
 func (c *rangeCheckTestCircuit) Define(api frontend.API) error {
-	rangecheck.New(api).Check(c.X, c.bits)
+	r := rangecheck.New(api)
+
+	r.Check(c.X, c.bits)
+
+	// 0 ≤ X ≤ 2ᵇⁱᵗˢ - 1 ⇔ - 2ᵇⁱᵗˢ + 1 ≤ -X ≤ 0 ⇔ 0 ≤ (2ᵇⁱᵗˢ - 1) - X ≤ 2ᵇⁱᵗˢ - 1
+	// i.e. X is in range iff (2ᵇⁱᵗˢ - 1) - X is
+	bound := big.NewInt(1)
+	bound.Lsh(bound, uint(c.bits)).Sub(bound, big.NewInt(1))
+	r.Check(api.Sub(bound, c.X), c.bits)
+
 	return nil
 }
--- a/test/engine.go
+++ b/test/engine.go
@@ -781,3 +781,40 @@ func (e *engine) MustBeLessOrEqCst(aBits []frontend.Variable, bound *big.Int, aF
 		panic(fmt.Sprintf("%d > %d", v, bound))
 	}
 }
+
+// Check asserts that the value given by v is less than 2^bits.
+func (e *engine) Check(v frontend.Variable, bits int) {
+
+	var i big.Int
+
+	switch vv := v.(type) {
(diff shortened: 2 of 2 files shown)


test/engine_test.go
```diff
@@ -101,3 +103,37 @@ func TestPreCompileHook(t *testing.T) {
 		t.Error("callback not called")
 	}
 }
+
+// TODO @ivokob is this test necessary, considering that the rangechecker is already tested against the test engine?
```
ivokub, line 107: Indeed I think this test is not particularly necessary. But I would add a test which would assert that the test engine implements `frontend.RangeChecker` to ensure that we won't have regression in the future.

ivokub (commented):
PR looks good, but I'll have a bit of thought about it. You are completely right that it makes debugging easier, but it also doesn't test other unit tests through the logderivative range checker, which I think can possibly lead to bugs.

What about implementing as an option - which when called wraps the test engine in
```go
type testChecker struct {
    *engine
}

func (c *testChecker) Check(v ....)
....
```

In this case we can keep the current test engine as is and if necessary wrap it in another implementation.

test/engine_test.go
```diff
@@ -101,3 +103,37 @@ func TestPreCompileHook(t *testing.T) {
...
+
+func TestRangeCheck(t *testing.T) {
```
ivokub, line 109: We have implemented only for test engine. CheckCircuit runs for both test engine and actual prover.

ivokub:
Suggested edit:
```diff
diff --git a/test/engine.go b/test/engine.go
index 3fdc80c1a..5c559449f 100644
--- a/test/engine.go
+++ b/test/engine.go
@@ -59,6 +59,7 @@ type engine struct {
 	kvstore.Store
 	blueprints        []constraint.Blueprint
 	internalVariables []*big.Int
+	isRangechecker    bool
 }
 
 // TestEngineOption defines an option for the test engine.
@@ -75,6 +76,16 @@ func SetAllVariablesAsConstants() TestEngineOption {
 	}
 }
 
+// AsRangechecker makes the test engine to implement [frontend.Rangechecker]
+// interface. Useful for checking alternative solving paths instead of using
+// in-circuit range checking mechanism.
+func AsRangechecker() TestEngineOption {
+	return func(e *engine) error {
+		e.isRangechecker = true
+		return nil
+	}
+}
+
 // WithBackendProverOptions is a test engine option which allows to define
 // prover options. If not set, then default prover configuration is used.
 func WithBackendProverOptions(opts ...backend.ProverOption) TestEngineOption {
@@ -136,7 +147,11 @@ func IsSolved(circuit, witness frontend.Circuit, field *big.Int, opts ...TestEng
 		}
 	}
 
-	if err = c.Define(e); err != nil {
+	var eapi frontend.API = e
+	if e.isRangechecker {
+		eapi = &rangecheckerEngine{engine: e}
+	}
+	if err = c.Define(eapi); err != nil {
 		return fmt.Errorf("define: %w", err)
 	}
 	if err = callDeferred(e); err != nil {
@@ -782,8 +797,12 @@ func (e *engine) MustBeLessOrEqCst(aBits []frontend.Variable, bound *big.Int, aF
 	}
 }
 
+type rangecheckerEngine struct {
+	*engine
+}
+
 // Check asserts that the value given by v is less than 2^bits.
-func (e *engine) Check(v frontend.Variable, bits int) {
+func (e *rangecheckerEngine) Check(v frontend.Variable, bits int) {
 
 	var i big.Int
 
diff --git a/test/engine_test.go b/test/engine_test.go
index 4934a0f50..d806b74b6 100644
--- a/test/engine_test.go
+++ b/test/engine_test.go
@@ -2,8 +2,6 @@ package test
 
 import (
 	"fmt"
-	"github.com/consensys/gnark/backend"
-	"github.com/consensys/gnark/std/rangecheck"
 	"math/big"
 	"testing"
 
@@ -104,36 +102,31 @@ func TestPreCompileHook(t *testing.T) {
 	}
 }
 
-// TODO @ivokob is this test necessary, considering that the rangechecker is already tested against the test engine?
-
-func TestRangeCheck(t *testing.T) {
-	NewAssert(t).CheckCircuit(&rangeCheckTestCircuit{bits: 4}, WithCurves(ecc.BN254), WithBackends(backend.PLONK),
-		WithValidAssignment(&rangeCheckTestCircuit{X: 0}),
-		WithValidAssignment(&rangeCheckTestCircuit{X: 15}),
-		WithInvalidAssignment(&rangeCheckTestCircuit{X: 16}),
-		WithInvalidAssignment(&rangeCheckTestCircuit{X: -1}),
-	)
-
-	NewAssert(t).CheckCircuit(&rangeCheckTestCircuit{bits: 254}, WithCurves(ecc.BN254), WithBackends(backend.PLONK),
-		WithValidAssignment(&rangeCheckTestCircuit{X: -1}),
-	)
-}
-
-type rangeCheckTestCircuit struct {
+type rangecheckerEngineCircuit struct {
 	X    frontend.Variable
 	bits int
 }
 
-func (c *rangeCheckTestCircuit) Define(api frontend.API) error {
-	r := rangecheck.New(api)
-
-	r.Check(c.X, c.bits)
-
-	// 0 ≤ X ≤ 2ᵇⁱᵗˢ - 1 ⇔ - 2ᵇⁱᵗˢ + 1 ≤ -X ≤ 0 ⇔ 0 ≤ (2ᵇⁱᵗˢ - 1) - X ≤ 2ᵇⁱᵗˢ - 1
-	// i.e. X is in range iff (2ᵇⁱᵗˢ - 1) - X is
-	bound := big.NewInt(1)
-	bound.Lsh(bound, uint(c.bits)).Sub(bound, big.NewInt(1))
-	r.Check(api.Sub(bound, c.X), c.bits)
-
+func (c *rangecheckerEngineCircuit) Define(api frontend.API) error {
+	rc, ok := api.(frontend.Rangechecker)
+	if !ok {
+		return fmt.Errorf("not rangechecker")
+	}
+	rc.Check(c.X, c.bits)
 	return nil
 }
+
+func TestIsRangechecker(t *testing.T) {
+	circuit := rangecheckerEngineCircuit{bits: 4}
+	valid := rangecheckerEngineCircuit{X: 15}
+	invalid := rangecheckerEngineCircuit{X: 16}
+	if err := IsSolved(&circuit, &valid, ecc.BN254.ScalarField(), AsRangechecker()); err != nil {
+		t.Fatal(err)
+	}
+	if err := IsSolved(&circuit, &invalid, ecc.BN254.ScalarField(), AsRangechecker()); err == nil {
+		t.Fatal(err)
+	}
+	if err := IsSolved(&circuit, &valid, ecc.BN254.ScalarField()); err == nil {
+		t.Fatal(err)
+	}
+}

```

ivokub:
See my proposed implementation. In this case, the current implementations stay as is but it is possible to enforce range checking optionally for debugging.

PR #1047, opened 2024-02-09
Refac/blob decompressor mirror

Tabaie:
Minor API changes and bugfixes for the compressor integration to use.


std/compress/io.go
```diff
@@ -114,7 +114,7 @@ func UnpackIntoBytesHint(_ *big.Int, ins, outs []*big.Int) error {
 }
 
 // ReadNum reads the slice c as a big endian number in base radix
-func ReadNum(api frontend.API, c []frontend.Variable, radix int) frontend.Variable {
+func ReadNum(api frontend.API, c []frontend.Variable, radix frontend.Variable) frontend.Variable {
```
gbotrel, line 117: why is the radix a variable now? is it really "variable" ??

std/compress/io.go
```diff
@@ -143,3 +143,99 @@ func ShiftLeft(api frontend.API, slice []frontend.Variable, shiftAmount frontend
...
+		api:         api,
+		toRead:      toRead,
+		radix:       twoPow(api, wordNbBits),
+		numBound:    twoPow(api, numNbBits),
+		wordsPerNum: wordsPerNum,
+	}
+}
+
+func twoPow(api frontend.API, n int) frontend.Variable {
+	res := frontend.Variable(1)
```
gbotrel, line 180: working with a constant here; I would just use `1 << n` or if it's larger than 64bits, bit.Int Rsh method, no need to use the API and incur the extra frontend compile cost

gbotrel (changes requested):

std/compress/io.go
Tabaie: Encountered an `n=248` in the zkevm monorepo.

std/compress/io.go
Tabaie: Just to cover possible cases where the radix is large(r than 64 bits)

gbotrel: then use a big.Int here ? It conveys more the "this is a constant" part IMO.

std/compress/io.go
gbotrel: then just big.Int LSH 👍

std/compress/io.go
```diff
@@ -170,18 +170,16 @@ func NewNumReader(api frontend.API, toRead []frontend.Variable, numNbBits, wordN
...
 
-func twoPow(api frontend.API, n int) frontend.Variable {
-	res := frontend.Variable(1)
-	for n > 0 {
-		res = api.Mul(res, 2)
-		n--
-	}
+func twoPow(n int) *big.Int {
+	_n := min(n, 62)
+	res := big.NewInt(1 << _n)
```
gbotrel, line 181: res := big.NewInt(1)
res.Lsh(res, n) ?

gbotrel (approved):

PR #1052, opened 2024-02-13
feat: add option for enforcing number of goroutines for the solver

### LINKED ISSUE #1048: feat: add solver option to define number of used parallel processes
Currently, the witness solver runs in parallel, but returns first encountered error. When debugging solver failures, this makes the behaviour non-deterministic and more difficult to debug. We should have a solver option to allow defining the number of parallel goroutines (instead of `runtime.NumCPU()`) for more predictable behaviour.

ivokub:
# Description

This PR allows restricting the number of threads in solver. This is mainly useful for debugging the solver as the solving order is deterministic.

Fixes #1048 

## Type of change

# How has this been tested?

# Checklist:

### DIFF
--- a/constraint/solver/options.go
+++ b/constraint/solver/options.go
@@ -17,7 +17,7 @@ type Option func(*Config) error
 type Config struct {
 	HintFunctions map[HintID]Hint // defaults to all built-in hint functions
 	Logger        zerolog.Logger  // defaults to gnark.Logger
-	NbThreads     int             // defaults to runtime.NumCPU()
+	NbTasks       int             // defaults to runtime.NumCPU()
 }
 
 // WithHints is a solver option that specifies additional hint functions to be used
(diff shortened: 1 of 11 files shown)


constraint/solver/options.go
```diff
@@ -53,11 +57,29 @@ func WithLogger(l zerolog.Logger) Option {
...
+// WithNbThreads sets the number of gorutines to use for the solver. If not set,
+// then the number of goroutines is set to runtime.NumCPU().
+//
+// The option may be useful for debugging the solver behaviour and restricting
+// the CPU usage. Note that this is not hard limit - in case the solver calls a
+// hint function which may create more goroutines, the number of goroutines may
+// exceed the limit.
+func WithNbThreads(nbThreads int) Option {
+	return func(opt *Config) error {
+		if nbThreads <= 0 {
```
gbotrel, line 69: would also check against a upper limit (say 512) -- it's a audit return we got for similar code in gnark-crypto;

+ naming wise, to be more "Go" idiomatic, I would co with naming this "WithNbThreads" "nbThreads" to... "WithNumCPU(...)" "numCPU".

2 other remarks;
* having a unit test somewhere testing that the solver outputs the same thing for 1 vs runtime.NuMCPU() might come handy
* probably not the intended use of the PR, but, if we known at compile time that we are going to run the solver with a single CPU, then, we d on't need to build the dependency tree and can benefit from much faster compile time. again, probably overkill / overcomplicating code base for no good reason.

gbotrel (changes requested):

constraint/solver/options.go
gbotrel: (or with nbTasks / WithNbTasks to be consistent with gnark-crypto fft and msm 🙄 )

ivokub: I like NbTasks more, it is more consistent with gnark-crypto side. I was also thinking about the name, but didn't want to have something Go-specific (NbGoroutines etc.).

ivokub: And I'll try to figure out how to reliably create the situation where the solver outputs different results. I think I can run a hint with short sleep and then returning error.

ivokub:
@gbotrel - renamed option and added test which checks that we solve in parallel and in sequence (with high probability).

gbotrel (approved):

PR #1061, opened 2024-02-18
Perf: optimize EC arithmetic

yelhousni:
# Description

A collection of small ideas to optimize the elliptic curve arithmetic in `emulated` and 2-chains packages (BLS12 and BLS24). Comments in the code explain the tricks.

## Type of change

# How has this been tested?

Same tests for the existing methods.

1- Emulated case: SNARK curve: BN254

- Emulated curve: e.g. *secp256k1*:
`ScalarMul`: **385,458** scs vs. 656,839 scs
`JointScalarMul`: **451,908** scs vs. 680,740 scs

- Emulated curve: e.g. *BW6-761*:
`ScalarMul`: **1,367,064** scs vs. 2,291,619 scs
`JointScalarMul`: **1,589,533** scs vs. 2,355,964 scs

2- Native case: e.g. *BLS12-377/BW6-761* chain
`varScalarMul G1` **3,851** scs vs. 3,858 scs
`JointScalarMul G1` **6,191** scs vs. 6,705 scs
`varScalarMul G2` **9,956** scs vs. 12,579 scs

3- Some circuits of interest:
- PLONK verifier (BW6-761 in BN254) goes from 50,750,530 scs to **33,860,282** scs.
- ECRECOVER (secp256k1 in BN254) goes from 688,362 scs to **459,082** scs.

# Checklist:


std/algebra/emulated/sw_emulated/point.go
```diff
@@ -522,60 +522,160 @@ func (c *Curve[B, S]) scalarMulGLV(Q *AffinePoint[B], s *emulated.Element[S], op
...
 	if err != nil {
 		panic(fmt.Sprintf("compute GLV decomposition: %v", err))
 	}
-	s1, s2, s3, s4 := sd[0], sd[1], sd[3], sd[4]
+	s1, s2, s3, s4, s5, s6 := sd[0], sd[1], sd[3], sd[4], sd[5], sd[6]
 
 	c.scalarApi.AssertIsEqual(
-		c.scalarApi.Add(s1, c.scalarApi.Mul(s2, c.eigenvalue)),
+		c.scalarApi.Add(s5, c.scalarApi.Mul(s6, c.eigenvalue)),
 		c.scalarApi.Add(s, c.scalarApi.Mul(frModulus, sd[2])),
```
ivokub, line 533: Do we need to output `sd[2]` and check here using `sd[2]`? Wouldn't it be sufficient if we only check
```go
c.scalarApi.AssertIsEqual(
		c.scalarApi.Add(s5, c.scalarApi.Mul(s6, c.eigenvalue)),
        s
)
```
as AsssertIsEqual perform equality check modulo `frModulus`?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -522,60 +522,160 @@ func (c *Curve[B, S]) scalarMulGLV(Q *AffinePoint[B], s *emulated.Element[S], op
...
-
-	var Acc, B1 *AffinePoint[B]
+
+	// s1, s2 can be negative (bigints) in the hint, but will be reduced
+	// in-circuit modulo the SNARK scalar field and not the emulated field. So
+	// we return in the hint both |s1|, |s2| and the flags s3=0/1, s4=0/1 to
+	// negate the point instead of the corresponding scalar. Since s3, s4 are
+	// either 0 or 1, we only need to check the first limb is zero or not.
+	selector1 := c.api.IsZero(s3.Limbs[0])
+	selector2 := c.api.IsZero(s4.Limbs[0])
```
ivokub, line 542: Should we add check that `(s1, s3, s5)` and `(s2, s4, s6)` correspond to each other? Otherwise we have asserted that `s5 + lambda*s6 == s`, but actually use `s1` and `s2` for binary decomposition. Malicious hint could return `s1` and `s2` which do not correspond to `s5` and `s6`.

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -522,60 +522,160 @@ func (c *Curve[B, S]) scalarMulGLV(Q *AffinePoint[B], s *emulated.Element[S], op
...
+
+	s1bits := c.scalarApi.ToBits(s1)
+	s2bits := c.scalarApi.ToBits(s2)
+	nbits := st.Modulus().BitLen()>>1 + 2
+
+	var Acc *AffinePoint[B]
 	// precompute -Q, -Φ(Q), Φ(Q)
```
ivokub, line 549: Now the tables are bigger, do we have `tableQ = Q, -Q, -3Q` and `tablePhiQ = phi(Q), -phi(Q), -3phi(Q)`?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -522,60 +522,160 @@ func (c *Curve[B, S]) scalarMulGLV(Q *AffinePoint[B], s *emulated.Element[S], op
...
+	// P = B4 and P' = B3
+	T15 := c.Neg(T12)
+	// T = [3](Φ(Q) - Q)
+	// P = B4 and P' = B4
+	T16 := c.Neg(T11)
+
+	// when nbits is odd, we need to handle the first iteration separately
+	if nbits%2 == 0 {
+		// Acc = [2]Acc ± Q ± R ± Φ(Q) ± Φ(R)
+		T := &AffinePoint[B]{
```
ivokub, line 638: Comment says that add `±R±phi(R)`, but in the implementation not? Should this be a random point?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -758,59 +858,64 @@ func (c *Curve[B, S]) jointScalarMulGLV(p1, p2 *AffinePoint[B], s1, s2 *emulated
...
+	// s1, s2 can be negative (bigints) in the hint, but will be reduced
+	// in-circuit modulo the SNARK scalar field and not the emulated field. So
+	// we return in the hint both |s1|, |s2| and the flags s3=0/1, s4=0/1 to
+	// negate the point instead of the corresponding scalar. Since s3, s4 are
+	// either 0 or 1, we only need to check the first limb is zero or not.
+	// Same goes for t1, t2.
+	selector1 := c.api.IsZero(s3.Limbs[0])
+	selector2 := c.api.IsZero(s4.Limbs[0])
+	selector3 := c.api.IsZero(t3.Limbs[0])
+	selector4 := c.api.IsZero(t4.Limbs[0])
```
ivokub, line 893: Same comment that should assert that `sd1, sd3, sd5` etc. should correspond to each other.

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -822,51 +927,104 @@ func (c *Curve[B, S]) jointScalarMulGLVUnsafe(Q, R *AffinePoint[B], s, t *emulat
...
-	t2 = c.scalarApi.Select(selector4, t2, t4)
+	B1 := Acc
+	// and then add the base point G to it to avoid incomplete additions in the
+	// loop, because when doing doubleAndAdd(Acc, Bi) as (Acc+Bi)+Acc it might
+	// happen that Acc==Bi or -Bi. But now we force Acc to be different than
+	// the stored Bi.  However we need at the end to subtract [2^nbits]G
+	// (harcoded) from the result.
+	//
+	// Acc = Q + R + Φ(Q) + Φ(R) + G
+	Acc = c.Add(Acc, c.Generator())
```
ivokub, line 953: Still - in this case still we would hit the incomplete addition when input is `-G`? What about the idea to select between two values to add in the beginning and later to subtract depending if the input is `-G` or not? We could even use `G` and `2G`.

ivokub (changes requested):
I think we should add the checks that the scalar decompositions `s0` and `|s0|` correspond to each other corresponding to the returned bit.

And have a look if we should also need to choose between different inputs which we add-sub in beginning/end of the loop. Imo we still have a chance to hit the incomplete case, but I might be wrong.

But otherwise, looks good! I think the performance improvements are significant.

std/algebra/emulated/sw_emulated/point.go
yelhousni: you're right! that was an old copy-paste from 2-chains

yelhousni:
> I think we should add the checks that the scalar decompositions `s0` and `|s0|` correspond to each other corresponding to the returned bit.
> 
> And have a look if we should also need to choose between different inputs which we add-sub in beginning/end of the loop. Imo we still have a chance to hit the incomplete case, but I might be wrong.
> 
> But otherwise, looks good! I think the performance improvements are significant.

Made the changes. You can review now. One thing I'm still not 100% sure about is the 0/1 (bigints) in the hint. Maybe we should assert that they are booleans in-circuit? or just revert to subtracting the subscalars and their absolute values instead.

ivokub:
> > I think we should add the checks that the scalar decompositions `s0` and `|s0|` correspond to each other corresponding to the returned bit.
> > And have a look if we should also need to choose between different inputs which we add-sub in beginning/end of the loop. Imo we still have a chance to hit the incomplete case, but I might be wrong.
> > But otherwise, looks good! I think the performance improvements are significant.
> 
> Made the changes. You can review now. One thing I'm still not 100% sure about is the 0/1 (bigints) in the hint. Maybe we should assert that they are booleans in-circuit? or just revert to subtracting the subscalars and their absolute values instead.

Seems good, I think the conditional select between `G` and `phi2(G)` is good.

I'll check a bit if I can optimize a bit the consistency check between `s1, s3, s5`. Currently it involves multiplication `s1*s1 == s5*s5`, but imo we could avoid multiplications by trying to check `(-1)^(1-s3)*s5 - s1 == 0`. And I think I'll also see if it makes sense to output the sign bits in a separate hint as when we use non-native hint calling, then we also range check all limbs of the bit which is wasteful. If we would output the sign bits directly from a hint, then can only do `AssertIsBoolean` on native API.

ivokub:
Hmm - but do we need to return both `s1` and `t1=|s1|` from the hint? Shouldn't it be sufficient if we return `|s1|` and sign bit and then in-circuit would compute `s1 = Select(signbit, t1, Neg(t1))`? I think this could reduce the number of constraints a little.

ivokub:
Hmm, the tests fail now, but it is not consistent. I also tried commit bc1c711f50db4560d637681ae7a4d8bcdea4e770 and with bigger test count `-count=100` it still occasionally fails.

ivokub:
Looking at the implementation now -- wouldn't it work better if the second point would be random? We could even compute it deterministically from a hash and make `R` and `[2**b] R` part of the curve parameters?

yelhousni:
> It looks like alternative approach works -- we have instead two hint functions where one returns non-native subscalars and the other one bits. When we call `Select` for the native bits, then they are range checked to be 0-1. This allows to avoid a few multiplications and range checks.
> 
> But I guess before merging this PR it would be better to create a new one which introduces the variants to `emulated.Field.NewHint` and `emulated.UnwrapHint` which allow native outputs.
> 
> But still, the test `TestJointScalarMul6` occasionally fails now, but it is not consistent. I also tried commit [bc1c711](https://github.com/Consensys/gnark/commit/bc1c711f50db4560d637681ae7a4d8bcdea4e770) and even there it still occasionally fails.
> 
> So I guess adding conditionally `G` or `phi^2(G)` doesn't work? Inside division hint we get `ModInverse(k*p, p)` which is undefined.

I think `G` and `phi^2(G)` still work fine but they need to be inverted the Select logic. Pushed the fix which seems to solce the `TestJointScalarMul6` failing test.

yelhousni:
> Looking at the implementation now -- wouldn't it work better if the second point would be random? We could even compute it deterministically from a hash and make `R` and `[2**b] R` part of the curve parameters?

randomly as in at each call of `JointScalarMul` we sample a random point `R` and compute `[2**b]R`? That would work in the sense that it's unfeasible for an attacker to come up with two input points `P` and `Q` such that `P+Q = phi(R)`. But if the random point is part of the curve parameters I don't think it different that having the base point `G`. If `R` is publicly available one can still craft the failing input point `P` and `Q`.

ivokub:
I separated the hint calling with native outputs in a separate PR (#1080). That PR also adds non-native hint calling with native inputs and implements tests. This means that we could revert the commit https://github.com/Consensys/gnark/pull/1061/commits/cdedecaff1cca5feb048aacbc3b5f4af04d7c271 from this PR as it would be merged in from #1080. But that means that #1080 needs to be merged first and then merged into PR.

Otherwise, your changes look good. I think the comments now are very nice, they make quite an understandable implementation.

ivokub (approved):
LGTM!

ivokub:
> > Looking at the implementation now -- wouldn't it work better if the second point would be random? We could even compute it deterministically from a hash and make `R` and `[2**b] R` part of the curve parameters?
> 
> randomly as in at each call of `JointScalarMul` we sample a random point `R` and compute `[2**b]R`? That would work in the sense that it's unfeasible for an attacker to come up with two input points `P` and `Q` such that `P+Q = phi(R)`. But if the random point is part of the curve parameters I don't think it's different from having the base point `G`. If `R` is publicly available one can still craft the failing input points `P` and `Q`.

I see. In this case I do agree it is better to stay with `G` and `phi2(G)` as we already have them.

ivokub:
Seemed that master applied well. I would say good to merge from my side!

PR #1063, opened 2024-02-19
feat: Groth16 Solidity contract with commitments

ahmetyalp:
# Description

This PR adds commitment verification to groth16 solidity verifier contract. Contract does not do foldings so uses first commitment for pairing check and uses sha256 hash as hash to field function. I added warning logs about these assumptions 

Resolves: #524 

## Type of change

# How has this been tested?

- [x] Test case added under `internal/backend/circuits/commit.go`. Tests depends on changes at https://github.com/Consensys/gnark-solidity-checker/pull/2

# Checklist:


backend/groth16/bn254/solidity.go
```diff
@@ -490,18 +632,77 @@ contract Verifier {
...
+            mstore(add(f, 0x60), PEDERSEN_G_X_0)
+            mstore(add(f, 0x80), PEDERSEN_G_Y_1)
+            mstore(add(f, 0xa0), PEDERSEN_G_Y_0)
+            calldatacopy(add(f, 0xc0), commitmentPok, 0x40)
+            mstore(add(f, 0x100), PEDERSEN_GROOTSIGMANEG_X_1)
+            mstore(add(f, 0x120), PEDERSEN_GROOTSIGMANEG_X_0)
+            mstore(add(f, 0x140), PEDERSEN_GROOTSIGMANEG_Y_1)
+            mstore(add(f, 0x160), PEDERSEN_GROOTSIGMANEG_Y_0)
+
+            success := staticcall(gas(), PRECOMPILE_VERIFY, f, 0x180, f, 0x20)
```
hussein-aitlahcen, line 686: is it possible to merge this with the next batch of pairings?

hussein-aitlahcen, line 686: is it possible to merge this with the next batch of pairings?

ahmetyalp: I am not sure the soundness if we combine two pairings. If pairings are correct then single pairing also works but I am not sure the other part

test/assert_solidity.go
```diff
@@ -53,29 +54,38 @@ func (assert *Assert) solidityVerification(b backend.ID, vk verifyingKey,
...
+		checkerOpts = append(checkerOpts, "--groth16")
 		var buf bytes.Buffer
 		_proof := proof.(*groth16_bn254.Proof)
 		_, err = _proof.WriteRawTo(&buf)
 		assert.NoError(err)
 		proofBytes := buf.Bytes()
-		// keep only fpSize * 8 bytes; for now solidity contract doesn't handle the commitment part.
-		proofBytes = proofBytes[:32*8]
+		if numOfCommitments == 0 {
+			proofBytes = proofBytes[:32*8] // exclude commitments count if there are no commitments
```
ivokub, line 73: I think this is a bit hacky - I think it would be better to have a similar interfaces as for plonk.Proof i.e. to have `Proof.MarshalSolidity` method which encodes only `Ar,Bs,Krs` or `Ar,Bs,Krs,commitments,commitmentsPok`.

Another approachs is that we encode the length of the commitments and use this to instead of having to provide the `--commitment` argument to `gnark-solidity-checker`.

ivokub (commented):
Looks good, thanks!

I'll try to think maybe we can omit the `--commitment` argument for `gnark-solidity-checker`. For now it works, but I'm quite certain it would haunt us in the future trying.

And also one part in the Solidity template -- I think it can be simplified a little to be more readable. Otherwise imo looks perfect. I tested locally and works end to end.

backend/groth16/bn254/solidity.go
```diff
@@ -381,19 +418,28 @@ contract Verifier {
             let s
             mstore(f, CONSTANT_X)
             mstore(add(f, 0x20), CONSTANT_Y)
+            {{- if gt $numCommitments 0 }}
+            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, commitments, {{mul 0x40 $numCommitments}}, g, 0x40))
```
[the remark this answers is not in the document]
ahmetyalp:
```go
kSum = v.curve.Add(kSum, &vk.G1.K[0])

for i := range proof.Commitments {
kSum = v.curve.Add(kSum, &proof.Commitments[i].G1El)
}
```
I assumed that `curve.Add` here corresponds to `ecAdd` in solidity and these `PRECOMPILE_ADD`s calculates sum of `K[0]` and commitments (in two steps)

Yes, we can merge `PRECOMPILE_ADD`s by moving commitments to memory especially for one commitment

backend/groth16/bn254/solidity.go
```diff
@@ -490,18 +632,77 @@ contract Verifier {
...
+        {{- if eq $numCommitments 0 }}
         (uint256 x, uint256 y) = publicInputMSM(input);
+        {{- else }}
+        // HashToField
+        uint256[{{$numCommitments}}] memory publicCommitments;
+        {{- range $i := intRange $numCommitments }}
+            publicCommitments[{{$i}}] = uint256(
+                sha256(
+                    abi.encodePacked(
+                        commitments[{{mul $i 2}}],
```
bernard-wagner, line 658: On larger witness data, this will result in `Stack too deep` or `memoryguard was present` with `via-ir`. 
Should try and pack the data using assembly to prevent this. Will see if I can contribute the necessary code.

ivokub: I'm not too knowledgeable about Solidity to comment on that. But I think in general it doesn't make sense to have too much of public witness as it increases MSM size, making the verifier more expensive.

Rather, what the circuit designer could do is that for expected public witness `P1, ..., Pn` it provides as a public witness only `P* = Hash(P1, ..., Pn)` and then as a private witness `P1, ..., Pn`. Then, in circuit it asserts that `P* = Hash(P1, ..., Pn)` holds. Now, when we want to actually use the values `P1, .., Pn` in Solidity, then we would have to perform same hashing in Solidity to verify that public witness `P*` and values `P1, ..., Pn` correspond to each other.

NB! Keep in mind that we would have to use ZK friendly hash function for that in order for the in-circuit assertion to be efficient.

backend/groth16/bn254/solidity.go
kustosz: I wonder if this could be done with a random linear combination? Would be a huge gas saving.

ivokub: I haven't written it out, but could be doable. And indeed, it would be a huge saving both for the in-circuit recursive verifier and Solidity verifier. I'll try and ask.

hussein-aitlahcen: > I haven't written it out, but could be doable. And indeed, it would be a huge saving both for the in-circuit recursive verifier and Solidity verifier. I'll try and ask.

Can confirm merging indeed work as I did it, but no clue security wise if this is sound...

hussein-aitlahcen: For 1 commitment, this saves approximately 70K gas

ivokub: Hmm, I'd assume if the coefficient for random linear combination comes from Fiat-Shamir, then I would by default assume it is sound. And we would need it anyway as in Solidity and in circuit we do not have access to RNG.

test/assert_solidity.go
ahmetyalp: I added `MarshalSolidity` for groth16 and simply tests using `interface{ MarshalSolidity() []byte }`

backend/groth16/bn254/solidity.go
```diff
@@ -381,19 +418,28 @@ contract Verifier {
             let s
             mstore(f, CONSTANT_X)
             mstore(add(f, 0x20), CONSTANT_Y)
+            {{- if gt $numCommitments 0 }}
+            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, commitments, {{mul 0x40 $numCommitments}}, g, 0x40))
+            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
```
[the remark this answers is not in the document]
ahmetyalp: Added special case for `nbCommitments == 1` that moves commitment to memory than calculates `ECADD(Constant, Commitment)` to use as a first point for the next loop

backend/groth16/bn254/solidity.go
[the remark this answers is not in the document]
ahmetyalp: I added a comment to explain commitment `ECADD` part, also simplify logic if there is only one commitment. I don't know if copying commitment array to `g` would be cheaper than calling `ECADD` twice

backend/groth16/bn254/solidity.go
```diff
@@ -409,13 +455,40 @@ contract Verifier {
...
     /// verifyCompressedProof. I.e. points (A, B, C) in compressed format.
+    {{- if gt $numCommitments 0 }}
+    /// @return compressedCommitments compressed Pedersen commitments from the proof.
+    /// @return compressedCommitmentPok compressed proof of knowledge for the Pedersen commitments.
+    {{- end }}
+    {{- if eq $numCommitments 0 }}
     function compressProof(uint256[8] calldata proof)
     public view returns (uint256[4] memory compressed) {
+    {{- else }}
+    function compressProof(
```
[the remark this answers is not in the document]
ahmetyalp: It seems like this function only called in test suite to calculate compressed proof

backend/groth16/bn254/solidity.go
```diff
@@ -425,61 +498,130 @@ contract Verifier {
...
-
-        // Check pairing equation.
-        bool success;
-        uint256[1] memory output;
-        assembly ("memory-safe") {
-            success := staticcall(gas(), PRECOMPILE_VERIFY, pairings, 0x300, output, 0x20)
+
+        {{- if gt $numCommitments 0 }}
+        {
+            (commitments[0], commitments[1]) = decompress_g1(compressedCommitments[0]);
```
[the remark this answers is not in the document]
ahmetyalp: I just added a for loop if `nbCommitment > 1` and `TODO`

ahmetyalp:
@ivokub I tried to address some comments you left. Thanks for the review

backend/groth16/bn254/solidity.go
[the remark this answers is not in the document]
kustosz: Context: this was meant as a utility function, so that the compression algorithm is conveniently accessible without having to understand it.

backend/groth16/bn254/solidity.go
hussein-aitlahcen: > Hmm, I'd assume if the coefficient for random linear combination comes from Fiat-Shamir, then I would by default assume it is sound. And we would need it anyway as in Solidity and in circuit we do not have access to RNG.

Doing the shamir trick on the proof commitment G1 point and proof commitment proof of knowledge G1 point is fine (or also we also need to do it with A/C and H^alpha from the groth16 proof)?

I'm currently doing:
$`r = hash(PC, PCPPOK)`$
$`e(A, B) . e(L, -H^{\gamma}) . e(C, -H^{\delta}) . e(PC^r, g) . e(PCPOK^r, g_{-1/\sigma})) = 1_GT`$

Every extra multiplication for the random linear combination is costing slightly more than 6K gas

hussein-aitlahcen: For reference https://xn--2-umb.com/23/groth16-batch/#poor-man-s-snarkpack

backend/groth16/bn254/solidity.go
ivokub: One approach would be to try to use some other hash function or try some continues hashing with several calls, but dunno if Solidity allows this.

But I think that in this PR we can keep as is.

backend/groth16/bn254/solidity.go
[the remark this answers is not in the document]
ivokub: Alright, makes sense!

internal/backend/circuits/circuits.go
```diff
@@ -39,5 +65,5 @@ func addNewEntry(name string, circuit frontend.Circuit, proverGood, proverBad []
 	}
 	solver.RegisterHint(hintFunctions...)
 
-	Circuits[name] = TestCircuit{circuit, proverGood, proverBad, hintFunctions, curves}
+	Circuits[name] = TestCircuit{Circuit: circuit, ValidAssignments: proverGood, InvalidAssignments: proverBad, HintFunctions: nil, Curves: curves}
```
ahmetyalp, line 68: I think I forgot to pass `hintFunctions` here

ivokub: And due to that, I think we can also remove the prover/verifier options for the integration tests.

ivokub (approved):
Lets wait for the CI pipeline to finish and I'll also create two new issues for tracking improvement effors, but I think the PR is now good to merge. We really appreciate the contribution and the implementation has been very good! Good work!

ivokub:
I created new issues for tracking improvements: https://github.com/Consensys/gnark/issues/1094 and https://github.com/Consensys/gnark/issues/1095.

PR #1066, opened 2024-02-28
feat: add range check selector retrieval

ivokub:
# Description

This PR adds an option to retrieve the IDs of the wire which were range checked through `frontend.Checker` interface.

cc @AlexandreBelling 

## Type of change

# How has this been tested?

Tested in Wizard.

# Checklist:

### DIFF
--- a/frontend/cs/scs/builder.go
+++ b/frontend/cs/scs/builder.go
@@ -688,19 +688,21 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
 	}
 }
 
-// GetWireGates returns the pairs (gateId, wireLocation) for the given wires in
-// the compiled constraint system:
-//   - gateId is the index of the gate in the constraint system.
-//   - wireLocation is the location of the wire in the gate (0=xA or 1=xB).
+// GetWireConstraints returns the pairs (constraintID, wireLocation) for the
+// given wires in the compiled constraint system:
+//   - constraintID is the index of the constraint in the constraint system.
+//   - wireLocation is the location of the wire in the constraint (0=xA or 1=xB).
 //
 // If the argument addMissing is true, then the function will add a new
-// instruction for each wire that is not found in the constraint system. This
-// may happen when getting the gate for a witness which is not used in gates.
-// Otherwise, when addMissing is false, the function returns an error if a wire
-// is not found in the constraint system.
+// constraint for each wire that is not found in the constraint system. This may
+// happen when getting the constraint for a witness which is not used in
+// constraints. Otherwise, when addMissing is false, the function returns an
+// error if a wire is not found in the constraint system.
 //
-// The order of the returned pairs is not the same as for the given arguments.
-func (builder *builder) GetWireGates(wires []frontend.Variable, addMissing bool) ([][2]int, error) {
+// The method only returns a single pair (constraintID, wireLocation) for every
+// unique wire (removing duplicates). The order of the returned pairs is not the
+// same as for the given arguments.
+func (builder *builder) GetWireConstraints(wires []frontend.Variable, addMissing bool) ([][2]int, error) {
 	// construct a lookup table table for later quick access when iterating over instructions
 	lookup := make(map[int]struct{})
 	for _, w := range wires {
@@ -712,32 +714,32 @@ func (builder *builder) GetWireGates(wires []frontend.Variable, addMissing bool)
 	}
 	res := make([][2]int, 0, len(wires))
 	iterator := builder.cs.GetSparseR1CIterator()
-	for c, gate := iterator.Next(), 0; c != nil; c, gate = iterator.Next(), gate+1 {
+	for c, constraintIdx := iterator.Next(), 0; c != nil; c, constraintIdx = iterator.Next(), constraintIdx+1 {
 		if _, ok := lookup[int(c.XA)]; ok {
-			res = append(res, [2]int{gate, 0})
+			res = append(res, [2]int{constraintIdx, 0})
 			delete(lookup, int(c.XA))
 			continue
 		}
 		if _, ok := lookup[int(c.XB)]; ok {
-			res = append(res, [2]int{gate, 1})
+			res = append(res, [2]int{constraintIdx, 1})
 			delete(lookup, int(c.XB))
 			continue
 		}
 	}
 	if addMissing {
 		for k := range lookup {
-			gate := builder.cs.AddSparseR1C(constraint.SparseR1C{
+			constraintIdx := builder.cs.AddSparseR1C(constraint.SparseR1C{
 				XA: uint32(k),
 				XC: uint32(k),
 				QL: constraint.CoeffIdOne,
 				QO: constraint.CoeffIdMinusOne,
 			}, builder.genericGate)
-			res = append(res, [2]int{gate, 0})
+			res = append(res, [2]int{constraintIdx, 0})
 			delete(lookup, k)
 		}
 	}
 	if len(lookup) > 0 {
-		return nil, fmt.Errorf("gate with wire not found in circuit")
+		return nil, fmt.Errorf("constraint with wire not found in circuit")
 	}
 	return res, nil
 }
--- a/frontend/cs/scs/builder.go
+++ b/frontend/cs/scs/builder.go
@@ -727,7 +727,11 @@ func (builder *builder) GetWireConstraints(wires []frontend.Variable, addMissing
 		}
 	}
 	if addMissing {
+		nbWitnessWires := builder.cs.GetNbPublicVariables() + builder.cs.GetNbSecretVariables()
 		for k := range lookup {
+			if k >= nbWitnessWires {
+				return nil, fmt.Errorf("addMissing is true, but wire %d is not a witness", k)
+			}
 			constraintIdx := builder.cs.AddSparseR1C(constraint.SparseR1C{
 				XA: uint32(k),
 				XC: uint32(k),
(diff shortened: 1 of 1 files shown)


frontend/cs/scs/builder.go
```diff
@@ -686,3 +697,39 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
...
+
+func (builder *builder) storeWireLocation(a, b int) {
+	for i, v := range []int{a, b} {
+		if _, ok := builder.mRangecheckInterleave[i][v]; !ok {
+			builder.mRangecheckInterleave[i][v] = builder.cs.GetNbConstraints()
+		}
+	}
+}
+
+func (builder *builder) GetWireGates(wires []frontend.Variable) [2][]int {
```
AlexandreBelling, line 709: I believe one alternative way to implement this would be to not store additional tracking data for the wires, but instead do **ALL** the extra computation within that function.

This would be done by scraping the wire IDs of `wires` within a set : `map[int]struct{}`. Then, we could try to iterate on all the `xA` and then all the `xB` of the compiled circuit to try and find all the wires in the set. The advantage of that solution is that we only "pay the cost" of `GetWireGates` when we use it; meaning that the compilation performances for the other circuits are independent.

Another point, is that if we try to find all the wires within xA (and luckily succeed) then it simplifies the range-checking mechanism later on the wizard because, then, we only have to look at a one column instead of 2. (I genuinely don't know if that would work though)

ivokub: I checked once if we could only get away with being able to check only a single column, but this doesn't always work - it depends on the order of arguments when calling `api.Mul`. Considering that the complexity may be quite high (ECDSA -> scalarmul -> field emulation -> defer -> api.Mul), then it is difficult to reliably ensure we only assign range checkable wire to the first or second position.

ivokub: But I guess one more thing what we can do is to implement `frontend.Checker` on gnark side -- maybe then we have more information? Currently we implement it on zkevm side (by appending all the variables we want to range check) and then use `GetWireGates` to get the corresponding gate locations. But if we implement it on gnark side and finding the first gate location then it is easier (as the number of gates may be smaller). But this makes the compilation slower (but only when we have the builder which implements `frontend.Checker`).

frontend/cs/scs/builder.go
```diff
@@ -82,9 +86,13 @@ func newBuilder(field *big.Int, config frontend.CompileConfig) *builder {
 		mtBooleans:       make(map[expr.Term]struct{}),
 		mMulInstructions: make(map[uint64]int, config.Capacity/2),
 		mAddInstructions: make(map[uint64]int, config.Capacity/2),
-		config:           config,
-		Store:            kvstore.New(),
-		bufL:             make(expr.LinearExpression, 20),
+		mRangecheckInterleave: map[int]map[int]int{
+			0: make(map[int]int, config.Capacity/2),
```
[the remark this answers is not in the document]
ivokub: I'm not sure if double, but it does increase it a lot, yes. But usually we compile once, so it wouldn't be too much of an issue.

frontend/cs/scs/builder.go
AlexandreBelling: You mean, moving the `externalRangeChecker` into `gnark`?

AlexandreBelling: > I checked once if we could only get away with being able to check only a single column, but this doesn't always work - it depends on the order of arguments when calling api.Mul. Considering that the complexity may be quite high (ECDSA -> scalarmul -> field emulation -> defer -> api.Mul), then it is difficult to reliably ensure we only assign range checkable wire to the first or second position.

I think it is fine if we go with an optimistic approach. e.g. first try to find the wire in xA. If not, then too bad : try to find it in xB.

ivokub: Yes, and then allowing to output something what the user needs. For example, which gates should be range checking gates etc.

ivokub: Yep. But after thinking a few minutes about it, I'm not sure it can be nicely done. I guess it would be better to try the memoryless approach where we perform searches only at call to `GetWireGate` and do not store the first-gate-per-wire map.

frontend/cs/scs/builder.go
```diff
@@ -686,3 +697,39 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
 		return term
 	}
 }
+
+func (builder *builder) storeWireLocation(a, b int) {
+	for i, v := range []int{a, b} {
+		if _, ok := builder.mRangecheckInterleave[i][v]; !ok {
+			builder.mRangecheckInterleave[i][v] = builder.cs.GetNbConstraints()
```
AlexandreBelling, line 704: Some notes: here we never capture the witness wires.

frontend/cs/scs/builder.go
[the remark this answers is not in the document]
ivokub: New implementation does not use memory in case we do not call `GetWireGates`

frontend/cs/scs/builder.go
ivokub: Now performs all computation when calling `GetWireGates`

frontend/cs/scs/builder.go
ivokub: Added optional argument to add dummy gates for witness wires.

frontend/cs/scs/builder.go
```diff
@@ -698,38 +688,56 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
...
-			existing[ww.WireID()] = struct{}{}
-			break
+		if _, ok := lookup[int(c.XB)]; ok {
+			res = append(res, [2]int{gate, 1})
+			delete(lookup, int(c.XB))
+			continue
 		}
 	}
-	return res
+	if addMissing {
```
AlexandreBelling, line 727: wouldn't that be called post-compilation?

ivokub:
Requesting additional review from @gbotrel to ensure I haven't missed anything.

AlexandreBelling (approved):
Approving to signify that I reviewed, but still should be reviewed by @gbotrel

frontend/cs/scs/builder.go
```diff
@@ -686,3 +687,57 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
...
+			delete(lookup, int(c.XA))
+			continue
+		}
+		if _, ok := lookup[int(c.XB)]; ok {
+			res = append(res, [2]int{gate, 1})
+			delete(lookup, int(c.XB))
+			continue
+		}
+	}
+	if addMissing {
```
gbotrel, line 727: if this can happen iff the wires are the witness, shouldn't we sanity check the wireIDs against the range of wireIDs for the witness ?

frontend/cs/scs/builder.go
```diff
@@ -686,3 +687,57 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
...
+//   - wireLocation is the location of the wire in the gate (0=xA or 1=xB).
+//
+// If the argument addMissing is true, then the function will add a new
+// instruction for each wire that is not found in the constraint system. This
+// may happen when getting the gate for a witness which is not used in gates.
+// Otherwise, when addMissing is false, the function returns an error if a wire
+// is not found in the constraint system.
+//
+// The order of the returned pairs is not the same as for the given arguments.
+func (builder *builder) GetWireGates(wires []frontend.Variable, addMissing bool) ([][2]int, error) {
```
gbotrel, line 703: noticed that if we have duplicate wireIDs in the inputs, then the length of the return variable will remove duplicates; might be worth documenting, don't know about current caller usage

frontend/cs/scs/builder.go
```diff
@@ -686,3 +687,57 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
 		return term
 	}
 }
+
+// GetWireGates returns the pairs (gateId, wireLocation) for the given wires in
```
gbotrel, line 691: naming wise, we use "constraint" everywhere, not gate, I'ld prefer to keep it that way for consistency even though it means the same thing here :)

gbotrel:
also, more general question on this PR, to understand the context; what happens if the caller queries say, wireID 42, and it appears in multiple constraints (i.e. gates) ? 

What is the expected behavior --> here from the current PR it returns the first constraint it appears into right ?

ivokub:
> also, more general question on this PR, to understand the context; what happens if the caller queries say, wireID 42, and it appears in multiple constraints (i.e. gates) ?
> 
> What is the expected behavior --> here from the current PR it returns the first constraint it appears into right ?

Yes, in this PR it returns the first constraint.

frontend/cs/scs/builder.go
ivokub: I think it makes sense to check wireIDs agains wireIDs of the witness. I think otherwise it is a bug in the builder itself when we have created a non-witness wire which doesn't appear in any constraint. I think the best approach would be to panic.

frontend/cs/scs/builder.go
ivokub: Yes, it would be better to document it

frontend/cs/scs/builder.go
ivokub: 👍  will change.

frontend/cs/scs/builder.go
ivokub: Added sanity check.

frontend/cs/scs/builder.go
ivokub: Documented.

frontend/cs/scs/builder.go
ivokub: Renamed.

PR #1072, opened 2024-03-06
feat: add MulNoReduce and Sum methods in field emulation

ivokub:
# Description

Non-reducing multiplication can be useful in contexts where we craft the non-native operations in a order where we know that we're going to use addition chains after. Doing this allows us to call the modular reduction later, saving constraints due to not having to reduce twice (before and after addition chains). For example, the approach can be used for computing inner products, where we first compute `a_i * b_i`, then compute `c = \sum_i a_i * b_i` and finally only reduce `c` once.

Depends on #1064 being merged.

## Type of change

# How has this been tested?

- [x] TestMulNoReduce

New methods, no impact on existing performance.

# Checklist:

### DIFF
--- a/std/math/emulated/field_ops.go
+++ b/std/math/emulated/field_ops.go
@@ -3,6 +3,7 @@ package emulated
 import (
 	"errors"
 	"fmt"
+	"math/bits"
 
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/std/selector"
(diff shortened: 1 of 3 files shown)


std/math/emulated/field_ops.go
```diff
@@ -132,6 +133,34 @@ func (f *Field[T]) add(a, b *Element[T], nextOverflow uint) *Element[T] {
...
+	limbs := make([]frontend.Variable, nbLimbs)
+	for i := range limbs {
+		limbs[i] = 0
+	}
+	for i := range inputs {
+		for j := range inputs[i].Limbs {
+			limbs[j] = f.api.Add(limbs[j], inputs[i].Limbs[j])
+		}
+	}
+	return f.newInternalElement(limbs, overflow+uint(addOverflow))
```
gbotrel, line 161: so, the final overflow is the max input overflow + `addOverflow` which is .. `bits.Len(nbInput)` ? 
if nbInput == 1 --> `addOverflow` should be 0 right? 
why is the `addOverflow` not `nbInputs - 1` ?

gbotrel:
> changed the base branch from master  to perf/emulated-assertisequal  
to have simpler diff

std/math/emulated/field_ops.go
```diff
@@ -132,6 +133,34 @@ func (f *Field[T]) add(a, b *Element[T], nextOverflow uint) *Element[T] {
 	return f.newInternalElement(limbs, nextOverflow)
 }
 
+func (f *Field[T]) AddMany(inputs ...*Element[T]) *Element[T] {
```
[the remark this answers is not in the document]
ivokub: Heh, of course :) Renamed

std/math/emulated/field_ops.go
ivokub: So, in case when we add two elements, we increase the overflow by one to account for additional bit required.

Now, when we add many elements, then the bound on the sum is `nbInputs * 2^(maxOverflow + nbBitsPerLimb)`. Its length is `log2(nbInputs * 2^(maxOverflow + nbBitsPerLimb)) = log2(nbInputs) + maxOverflow + nbBitsPerLimb`. When counting only the overflow (so can omit `nbBitsPerLimb` part), then it is `log2(nbInputs) + maxOverflow` and we need to round `log2(nbInputs)` up.

ThomasPiellard (approved):

PR #1091, opened 2024-03-27
Refac/compressor nodictref2

Tabaie:
# Description

This PR mirrors changes in https://github.com/consensys/compress removing the idea of compression level, and turning dictionary references into back-references to "negative indices."

Also includes some constraint level optimizations.

## Type of change

Performance improvement

# How has this been tested?

All unit tests pass

Compiled by `lzss/large-tests` on EIP4844 blob-sized input

# Checklist:


go.mod
```diff
@@ -1,12 +1,14 @@
 module github.com/consensys/gnark
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
gbotrel, line 5: heh didn't know this one. Why did you need to add that? :)

std/compress/lzss/snark.go
```diff
@@ -10,43 +10,44 @@ import (
...
 	// check that the input is in range and convert into small words
 	rangeChecker := internal.NewRangeChecker(api)
 
 	bytes := make([]frontend.Variable, len(c)-sizeHeader+1)
 	copy(bytes, c[sizeHeader:])
-	bytes[len(bytes)-1] = 0                                             // pad with a zero to avoid out of range errors
-	c, bytes = rangeChecker.BreakUpBytesIntoWords(wordNbBits, bytes...) // from this point on c is in words
-	cLength = api.Mul(api.Sub(cLength, sizeHeader), 8/wordNbBits)       // one constraint; insignificant impact anyway
+	bytes[len(bytes)-1] = 0                                    // pad with a zero to avoid out of range errors
+	c, bytes = rangeChecker.BreakUpBytesIntoWords(1, bytes...) // from this point on c is in bits
```
gbotrel, line 49: can the code linked to this be simplified ? (i.e now that we assume 8 bits per word all the time; )

std/compress/lzss/snark.go
```diff
@@ -133,31 +139,39 @@ func sliceToTable(api frontend.API, slice []frontend.Variable) *logderivlookup.T
...
 	}
 
 	res := logderivlookup.New(api)
 
-	for i := range c {
-		entry := frontend.Variable(0)
-		for j := range backrefs {
-			isSymb := api.IsZero(api.Sub(bytes[i], backrefs[j].Delimiter))
+	for i := range _bits {
+		/*entry := frontend.Variable(0)
```
gbotrel, line 168: remove deadcode

gbotrel (approved):

std/compress/lzss/snark.go
Tabaie: I think so. We can just use `api.ToBinary` but I'd like to wait for the next version because I'm planning some input granularity shenanigans where this may come in handy.

go.mod
Tabaie: I think `go get` adds it automatically.

Issue #1094, opened 2024-04-05
refactor: Groth16 Solidity verifier improvements


ivokub:
With https://github.com/Consensys/gnark/pull/1063 being merged, there are still a few open issues to solve. This issue lists all the related sub-issues
* [ ] the method [`NbPublicWitness`](https://github.com/Consensys/gnark/blob/master/backend/groth16/bn254/setup.go#L653-L655) returns the number of `K` elements in the key, but this may mismatch the actual number of assignable number of public witness elements in case commitments are used. But for PLONK backend, the method returns correct output. This leads to a mismatch in [Solidity test](https://github.com/Consensys/gnark/pull/1063/files#diff-f42eb1f1c6ad5c46af0a8f7f9301fc25f3e802336777e9113eef0d3d254641c9R55) where for PLONK we compute wrong number of commitments used, but fortunately it doesn't affect `gnark-solidity-checker` as it doesn't actually use `--commitments` parameter.
* [ ] add method `NbCommitments` to the proving/verification key for Groth16 and PLONK.
* [ ] currently there is API mismatch between Groth16 and PLONK solidity verifier. PLONK solidity verifier takes as [inputs the serialized proof as byte array](https://github.com/Consensys/gnark/blob/master/backend/plonk/bn254/solidity.go#L172), but Groth16 verifier expects [an array of `uint256`](https://github.com/Consensys/gnark/blob/master/backend/groth16/bn254/solidity.go#L496). This means that we have to manually convert between serialized gnark proof and what Solidity verifier expects (and this may lead to invalid inputs). I think it would be better to also takes as input the serialized proof and additionally remove the conditional method signature difference depending if the circuit uses commitments or not (as it can be deduced from the proof itself). Now, the problem may be that the serialized proof is not the most gas-optimal as it includes as array header the number of items in the array, which may have a lot of zeroes. But we can change the `ExportSolidity` method on `groth16.Proof` to use smaller types for representing lengths (I think uint8 could actually be sufficient)
* [ ] with the previous modification, it is possible to drop the `--commitments` argument from `gnark-solidity-checker` tool as the Solidity contract should parse sufficient number of commitment elements from the proof directly. For me it seems a little error-prone that we can provide mismatching arguments (e.g. proof has N/X commitment elements and public witness elements, but arguments define M/Y).
[...]

PR #1097, opened 2024-04-05
feat: groth16 solidity use calldatacopy for commitments

bernard-wagner:
# Description

Addresses some causes for `Stack too deep` or `memoryguard was present` compiler errors by using assembly to copy calldata to an array. Tracked as part of issue #1094 .


ivokub (approved):
Thanks for the contribution! Looks good. I additionally re-enabled a few checks which failed previously with Solidity verifier as it was commiting to many public inputs.

PR #1101, opened 2024-04-12
feat: handle invalid signature failures in ECRecover precompile

ivokub:
# Description

In zkEVM, the signature values for the ECRecover precompile can be invalid in some cases. The failing cases are:
* `r^3 + ax+b` is not a quadratic residue, preventing computing the y-coordinate of commitment R
* the resulting public key is 0

We need to be able to check such cases to ensure that we can prove reverts. This PR adds a path to also check invalid inputs when `isFailure` input is given.

## Type of change

Breaking change due to changing the signature of `ECRecover` function.

# How has this been tested?

- [x] TestECRecoverQNR - tests that check for QNR failure succeeds
- [x] TestECRecoverQNRWoFailure - tests that QNR input cannot be provided when `isFailure` argument is false
- [x] TestECRecoverInfinity - tests that check for zero public key succeeds
- [x] TestECRecoverInfinityWoFailure - tests that zero PK input cannot be provided when `isFailure` is false
- [x] TestInvalidFailureTag - tests that valid signature cannot be proven when `isFailure` is true

ECRecover master 459082, now 465156

# Checklist:


ivokub:
Implemented for now, but the increase in the number of constraints is huge. I recon this is due to needing to use `WithCompleteArithmetic` when computing the public key in-circuit. I'm trying to figure out maybe we can still use incomplete arithmetic (by adding-subtracting G for example) for scalar multiplication.

ivokub:
> Implemented for now, but the increase in the number of constraints is huge. I recon this is due to needing to use `WithCompleteArithmetic` when computing the public key in-circuit. I'm trying to figure out maybe we can still use incomplete arithmetic (by adding-subtracting G for example) for scalar multiplication.

This approach works. Now can avoid complete arithmetic in `JointScalarMul`, but I still need to do one complete addition. It is now 477275 constraints (vs master 459082)

ivokub:
And, before we can merge this one, we need to merge https://github.com/Consensys/gnark-crypto/pull/497 and update go.mod here to point to gnark-crypto master.

PR #1110, opened 2024-04-23
Perf: revisiting field extensions in `std/`

yelhousni:
# Description

Revisiting quadratic and cubic field extensions using Karatsuba, Toom-Cook-3 and Chung-Hasan (sq). Basically experiments show that Karatsuba is better but this PR just saves some (emulated) subtractions here and there. Direct extensions using Toom-Cook work nice;  for example an (unoptimized) version Toom-Cook-6 saves ~2k scs per multiplication (theoretically it saves 7 Fp-mul but uses more adds/subs) . However they are removed from this PR for now because the pairing implementation should be rewritten in the direct extension (Karabina, Granger-Scott, lines...), which will be in a separate PR. 

## Type of change

# How has this been tested?

Existing tests work. 

- PLONK verifier circuit on bw6-761:
  - old: 32,145,457 scs
  - new: 32,094,676 scs

- Pairing check circuit `e(a,b)e(c,d)==1` on BN254:
  - old: 6,674,047 scs
  - new: 6,620,773 scs
  
- Pairing check circuit `e(a,b)e(c,d)==1` on BLS12-381:
  - old: 8,928,354 scs
  - new: 8,792,606 scs
  

# Checklist:


std/algebra/emulated/fields_bw6761/e6.go
```diff
@@ -85,13 +86,18 @@ func (e Ext6) Mul(x, y *E6) *E6 {
...
 	c0 := e.Ext3.Sub(&x.B0, &x.B1)
-	c3 := e.Ext3.MulByNonResidue(&x.B1)
-	c3 = e.Ext3.Neg(c3)
-	c3 = e.Ext3.Add(&x.B0, c3)
+	c3 := &E3{
+		A0: x.B1.A2,
+		A1: x.B1.A0,
+		A2: x.B1.A1,
+	}
+	c3.A0 = *e.fp.MulConst(&c3.A0, big.NewInt(4))
```
ivokub, line 97: Here it is ok to do `c3.A0 = *e.fp.MulConst(&c3.A0, big.NewInt(4))` (overwrite `c3.A0`) as `c3.A0` is not part of a multiplication check right now, but long-term may be error prone. It would be better to do
```go
d0 := e.fp.MulConst(&x.B1.A2, big.NewInt(4))
d0 = e.fp.Add(&X.B0.A0, d0)
d1 := e.fp.Sub(&x.B0.A1, &x.B1.A0)
d2 := e.fp.Sub(&x.B0.A2, &x.B1.A1)
c3 = &E3{
    A0: *d0,
    A1: *d1,
    A2: *d2,
} 
```
instead to avoid bugs in the future.

ivokub (approved):
Looks perfect. Left a small suggestion, which is not an issue now, but may lead to a hard-to-trace bug in the future when we add multiplication to Ext6.Square after creating c3

std/algebra/emulated/fields_bw6761/e6.go
yelhousni: Got it! anyway this will be superseded by https://github.com/Consensys/gnark/pull/1126

yelhousni (commented):
I guess we can merge this to incorporate bn254 and bls12-381 optims in master and then merge master into https://github.com/Consensys/gnark/pull/1126 and resolve conflicts related to bw6.

ivokub:
> @ivokub I guess we can merge this to incorporate bn254 and bls12-381 optims in master and then merge master into #1126 and resolve conflicts related to bw6.

Yep, I think it is a good approach 👍

PR #1119, opened 2024-04-29
perf/feature: new serialization format for constraint systems

gbotrel:
# Description

This PR uses [integer compression techniques](https://github.com/ronanh/intcomp)  to serialize some constraint systems parts containing ids, offsets, etc and bypass our current cbor choice. (cbor is still used for the rest). 

As a bench, using `lzss.DecompressionTestCircuit` with params landing at 90M constraints (plonk), we go from 9225Mb to 2238Mb size, plus, we serialize (and most importantly deserialize) way faster:
```
benchmark                                      old ns/op        new ns/op      delta
BenchmarkSerialization/serialize_cbor-96       54962357901      3086254450     -94.38%
BenchmarkSerialization/serialize_cbor-96       41877637150      3117870148     -92.55%
BenchmarkSerialization/deserialize_cbor-96     100910397720     4267472009     -95.77%
BenchmarkSerialization/deserialize_cbor-96     117815108278     3776636214     -96.79%
```


constraint/marshal.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+
+	// header
+	h := header{
```
ivokub, line 47: I see why using `[]byte` now instead of `io.Reader`, this allows to create correct header right now. Maybe instead we can still use `io.Writer` in calls to `..ToBytes` above by writing header first with padded zeros, then seeking back to start and overwriting with actual lengths?

ivokub: But I think we do not need the actual lengths at all? If we use some end-of-struct separator in the serialization, then during Reading in method `System.FromBytes` the methods `levelsFromBytes` etc. can just read until the separator?

ivokub: Or we can prepend the number of elements (not the number of bytes). And in `...FromBytes` we read as many elements expected.

internal/backend/ioutils/intcomp_test.go
```diff
@@ -0,0 +1,38 @@
+package ioutils
+
+import (
+	"bytes"
+	"testing"
+)
+
+func FuzzIntcomp32(f *testing.F) {
```
ivokub, line 8: Fuzzing becomes more efficient with some seeded corpus. I ran it for a few minutes and built one. Should I included it? It also becomes regression test in the sense that when `FuzzIntcomp32` is run as a unit test not a fuzzing test then checks all the cases defined in the corpus.

ivokub: And as a general note, Go integrated fuzzing seems very powerful, I think we should aim to include it with more in serialization tests in general.

ivokub (approved):
Very nice optimizations! I think in general is good, but in a perfect scenario I would try to figure out if we can work with `io.Writer` and `io.Reader` instead of direct byte arrays as imo otherwise we have many copies of very similar structures in memory (constraint system itself, bytes when serializing them and also byte buffer for internal steps a la serializing calldata). But due to applying compression to uints we still cannot get away with all.

And I completely agree that it doesn't make sense for us to include additional compression in the implementation. It should be straightforward to add `gzip` etc. when the user is interested in it. Otherwise there is too many options (fast or good compression ...)

internal/generator/backend/template/representations/coeff.go.tmpl
```diff
@@ -27,6 +29,42 @@ func newCoeffTable(capacity int) CoeffTable {
...
+	for _, c := range ct.Coefficients {
+		for _, w := range c {
+			buf = binary.LittleEndian.AppendUint64(buf, w)
+		}
+	}
+
+	return buf
+}
+
+func (ct *CoeffTable) fromBytes(buf []byte) error {
```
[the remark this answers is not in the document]
gbotrel: yes so, bottom line is empirically, on large circuits, perf suffers by a significant factor (30-40% if I recall correctly without accounting for parallelism) using `io` `Reader` and `Writer`. It does a bit of reflection (or interface method access) behind the scene for example for `Uvarint` which is annoying ; plus doing []byte allows for parallel section processing. Wanted to maximize reading speed :) .

Second counter argument; yes it uses more memory than a streaming solution, but, 1GB will be for a 100M~ constraints circuits, I think if you are working with such large circuits, you can afford 1GB since you'll need 40times that anyway for the prover 😁

constraint/marshal.go
gbotrel: having the lengths enable the reader to process in parallel the sections. But yes we could avoid it in few ways, it's just more convenient and is quite cheap (4 uint64 vs the weight of the circuit is nothing).

internal/backend/ioutils/intcomp_test.go
gbotrel: oh yes feel free to include it !

ivokub: Added test corpus

ivokub:
> Very nice optimizations! I think in general is good, but in a perfect scenario I would try to figure out if we can work with `io.Writer` and `io.Reader` instead of direct byte arrays as imo otherwise we have many copies of very similar structures in memory (constraint system itself, bytes when serializing them and also byte buffer for internal steps a la serializing calldata). But due to applying compression to uints we still cannot get away with all.
> 
> And I completely agree that it doesn't make sense for us to include additional compression in the implementation. It should be straightforward to add `gzip` etc. when the user is interested in it. Otherwise there is too many options (fast or good compression ...)

Yep, it makes sense that if we work with `[]byte` then can have parallel processing which is more important than saving a few GB in settings where memory usage is in tens of GBs anyway. Added the test corpus.

Good to merge from my side.

PR #1139, opened 2024-05-22
fix(uints): constrain valueOf

bernard-wagner:
# Description

As stated in the TODO, `uints8.ValueOf` is not constrained. This commit constrains the output of ValueOf. 

I am jumping in on the deep end with circuit development, so would appreciate as much feedback as possible. Especially would appreciate guidance on testing constraints for intermediate values, the overflow was the only technique I could think of that works in a unit test.

## Type of change

# How has this been tested?

Added a unit test that if added to the current master branch will fail. The third test case should return an error as the input exceeds the output size.

# Checklist:

### DIFF
--- a/std/math/uints/uint8.go
+++ b/std/math/uints/uint8.go
@@ -173,10 +173,18 @@ func (bf *BinaryField[T]) ValueOf(a frontend.Variable) T {
 	if err != nil {
 		panic(err)
 	}
-	// TODO: add constraint which ensures that map back to
-	for i := range bts {
+
+	var checker frontend.Variable = 0x00
+	for i := len(bts) - 1; i >= 0; i-- {
 		r[i] = bf.ByteValueOf(bts[i])
+
+		// Reconstruct the value of a
+		checker = bf.api.Add(bf.api.Mul(checker, 256), r[i].Val)
+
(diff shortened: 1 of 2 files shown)


ivokub:
Thanks for the PR. I'm trying to have a look at it this week to give feedback.

As of why we yet haven't added the constraining of the values always, is that in the context of some operations (XOR, AND), the `uints.U8.Value` are used as keys in looking up from the table and as such if the would not be in range, then the lookup would fail. So there is implicit constraining of the values to be 8-bit. But for example when we're doing some other operations which is not based on lookups, then it could lead to underconstrained circuits indeed.

I haven't looked at your implementation yet, but my general idea for building it better was to have an additional value or DB which says if U8 value is used in lookup table operations and is implicitly constrained.

ivokub (approved):
Alright - I finally got to reviewing it. Your initial PR was spot on, but it revealed another issue in `Add` method where we didn't check that the carry we omitted was correct. And fixing that made me realize that I could clean up some parts.

I pushed the changes directly to your branch, I hope you wont mind. Maybe you can have a look before I merge?

Otherwise, thanks for the contribution! Great work!

PR #1143, opened 2024-05-24
perf: replace BN254 final exp by a class equivalence check

yelhousni:
# Description

For pairing checks of the form `∏ e(Pi, Qi) == 1` we replace the final exponentiation by class equivalence check as described in https://eprint.iacr.org/2024/640.pdf (Section 4). We compute in a hint the residue witness `c` and check in-circuit that `f * w  == c^λ` where `f = ∏ MillerLoop(Pi, Qi)`, `w` some hinted scaler to ensure `f*w` is a cubic residue and `λ=6u+2+q^3-q^2+q` the optimal exponent (instead of `r`) with `u` the curve seed.  Exponentiation by `6u+2` is done using an optimized addition chain. 

The paper suggests to include `c^(6u+2)` in the multi-Miller loop computation so that the mutualized squarings in the loop would catch the squarings needed for `c^(6u+2)` too but I don't see how this can be done since we need `f` to compute `c`. One thing we can maybe do later is to push `f` and all the hint outputs to the torus or at least the cyclotomic subgroup by doing the easy part of the final exponentiation so that the FE elimination trick should be conducted with torus-based arithmetic over Fp6 (or at least with cyclotomic squarings over Fp12). 

_Edit: the squaring mutualization is possible (see https://github.com/Consensys/gnark/pull/1143#issuecomment-2148632498). Cyclotomic subgroup / torus push becomes inefficient now that squarings are for "free"._

_Small typo in Alg.4 of the paper for the modified Tonelli-Shanks: p^k-1 = 3^n * s instead of p-1 = 3^r * s_

## Type of change

# How has this been tested?

The existing `TestPairingCheckTestSolve` works for this.

Compared to the previous torus-based final exponentiation, this PR saves  in the ECPair precompile:
- ~807,034~ 1,378,371 scs

# Checklist:


zilong-dai:
cool

yelhousni:
> The paper suggests to include c^(6u+2) in the multi-Miller loop computation so that the mutualized squarings in the loop would catch the squarings needed for c^(6u+2) too but I don't see how this can be done since we need f to compute c.

Actually we can feed the hint with the points instead of the Miller function to compute the witness residue. This means we compute the Miller loop out-circuit (in the hint) and based on that we compute the witness residue `c` and then, in-circuit, we re-compute the Miller loop alongside `c^{6u+2}`. We can do this by initilizing the Miller loop accumulator in-circuit by `1/c` instead of 1 and when bit is we multiply it by `1/c` and resp. `c` when bit is -1. Both `c` and `1/c` are provided in the hint. Finally we check that `w * f * (1/c)^(q^3-q^2+q) == 1` only this time `f` is `f = ∏ MillerLoop(Pi, Qi) * (1/c)^(6u+2)`.

However, we need the Miller loop out-circuit to match exactly the in-circuit version to have the same witness residue. Previously in gnark-crypto we used `doubleStep` and `addStep` for affine pairing while here we use `doubleAndAddStep` this resulted in a mismatch of lines but is now fixed in https://github.com/Consensys/gnark-crypto/pull/506.

yelhousni:
Some remarks:
- The efficiency of the technique requires `d` to be small. For BN254 `d=3` (requiring a cube root) and for BW6-761 it's even better `d=1` but for BLS families `d=u-1` where `u` is the curve seed.
- The technique is efficient only when pairings are equal i.e. `∏ MillerLoop(Pi, Qi) == 1`. When they are not we end up with some residue `w` we need to check in-circuit that `w^r == 1`. Ofc most applications need the `==1` check but maybe some future applications would requite the `!=1` check e.g. some kind of fraud proofs.

std/algebra/emulated/sw_bn254/pairing.go
```diff
@@ -644,12 +642,150 @@ func (pr Pairing) MillerLoopAndMul(P *G1Affine, Q *G2Affine, previous *GTEl) (*G
...
+	t1 := pr.FrobeniusCube(&residueWitnessInv)
+	t0 := pr.FrobeniusSquare(&residueWitnessInv)
+	t1 = pr.DivUnchecked(t1, t0)
+	t0 = pr.Frobenius(&residueWitnessInv)
+	t1 = pr.Mul(t1, t0)
+
+	t2 = pr.Mul(t2, t1)
+
+	pr.AssertIsEqual(t2, pr.One())
+
```
feltroidprime, line 789: Doesn't it lack a check that mul(residue_witness, residue_witness_inv) == 1 ?

feltroidprime:
> Some remarks:
> 
>     * The efficiency of the technique requires `d` to be small. For BN254 `d=3` (requiring a cube root) and for BW6-761 it's even better `d=1` ([perf: replace BW6-761 final exp by a class equivalence check #1155](https://github.com/Consensys/gnark/pull/1155)) but for BLS families `d=u-1` where `u` is the curve seed (64 bits for BLS12).


Moreover 1/m' mod h doesn't exist for BLS12-381 ...

std/algebra/emulated/sw_bn254/pairing.go
yelhousni: yeah you're right. Either this or to compute the inverse in-circuit.

yelhousni:
> > Some remarks:
> > ```
> > * The efficiency of the technique requires `d` to be small. For BN254 `d=3` (requiring a cube root) and for BW6-761 it's even better `d=1` ([perf: replace BW6-761 final exp by a class equivalence check #1155](https://github.com/Consensys/gnark/pull/1155)) but for BLS families `d=u-1` where `u` is the curve seed (64 bits for BLS12).
> > ```
> 
> Moreover 1/m' mod h doesn't exist for BLS12-381 ...

yes `gcd(m',h)=u-1`

ivokub (approved):
Looks good. I got the idea of having last ML and FinalExp in the same call.

yelhousni:
> Looks good. I got the idea of having last ML and FinalExp in the same call.

we can do this actually for the whole multi-pairing, but for the precompile we need to have fixed circuits regardless of the number of pairs so I did the trick only for the last Miller loop.

std/algebra/emulated/fields_bn254/e12_pairing.go
```diff
@@ -351,3 +421,127 @@ func (e Ext12) FrobeniusCubeTorus(y *E6) *E6 {
...
+		C0: E6{
+			B0: E2{A0: *res[12], A1: *res[13]},
+			B1: E2{A0: *res[14], A1: *res[15]},
+			B2: E2{A0: *res[16], A1: *res[17]},
+		},
+		C1: E6{
+			B0: E2{A0: *res[18], A1: *res[19]},
+			B1: E2{A0: *res[20], A1: *res[21]},
+			B2: E2{A0: *res[22], A1: *res[23]},
+		},
```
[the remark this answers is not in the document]
yelhousni: Thanks for reporting!
#1213 #1214

PR #1160, opened 2024-06-10
fix: fixes #1157 ensures calls to AttachDebugInfo are surrounded with…

### LINKED ISSUE #1157: bug: AttachDebugInfo() contributes to most memory usage regardless of debug.Debug flag value
## Description
While profiling gnark during compilation and proof phaze I noticed that `AttachDebugInfo` and in turn `NewDebugInfo` contribute to large (if not the largest) portion of memory. In my case it was 500mb out of 800.

## Expected Behavior
Do not create anything in `NewDebugInfo` if `debug.Debug` is set to `false`

## Actual Behavior
NewDebugInfo() contributes to memory growth regardless of `debug.Debug` const value

## Possible Fix

smth like 
```go
NewDebugInfo(errName string, i ...interface{}) DebugInfo {
	var l LogEntry

	if !debug.Debug{
		return DebugInfo(l)
	}
	...
```
would do the job

## Context
We are planning to run gnark on mobile devices where memory consumption turns to be critical

## Your Environment
* gnark version used (e.g. v0.10.0):

gbotrel:
Fixes #1157

### DIFF
--- a/debug_test.go
+++ b/debug_test.go
@@ -78,6 +78,9 @@ func (circuit *divBy0Trace) Define(api frontend.API) error {
 }
 
 func TestTraceDivBy0(t *testing.T) {
+	if !debug.Debug {
+		t.Skip("skipping test in non debug mode")
+	}
 	assert := require.New(t)
 
 	var circuit, witness divBy0Trace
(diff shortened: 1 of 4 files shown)


ivokub (approved):
Looks good to me. Have a look at alternative way to avoid initializing debuginfo when not in debug mode and having single `builder.AddPlonkConstraint` call regardless of debug tag.

PR #1155, opened 2024-06-03
perf: replace BW6-761 final exp by a class equivalence check

yelhousni:
# Description

Similarly to #1143 we adapt https://eprint.iacr.org/2024/640.pdf to the BW6-761 case with the following parameters:
- $h = \frac{p^6-1}{r}=3^2\cdot l$ with $\text{gcd}(l,3)=1$
- $\lambda=x^3-x^2+1-(x+1)q$
```magma
QQ := Rationals();
QQx<x> := PolynomialRing(QQ);
rx := (x^6 - 2*x^5 + 2*x^3 + x + 1)/3;
qx := (103*x^12 - 379*x^11 + 250*x^10 + 691*x^9 - 911*x^8 - 79*x^7 + 623*x^6 - 640*x^5 + 274*x^4 + 763*x^3 + 73*x^2 + 254*x + 229)/9;
M := Matrix(QQx, 2, 2, [rx, 0, -qx mod rx, 1]);
R := LLL(M);
print R

[        x + 1 x^3 - x^2 - x]
[x^3 - x^2 + 1        -x - 1]

assert ((R[2][1] + qx*R[2][2]) mod rx) eq 0; // better Hamming weight
```
- $m=\lambda/r$
- $d=\text{gcd}(m,h)=1$
- $m'=m/d = m$

First, we find the residue witness in a hint:
1. Compute r-th root: Raising the miller function to $1/r \pmod h$
2. Compute m′-th root: Raising the result to $1/m' \pmod h$

Then, we check in-circuit that:
```
MillerLoop == Witness ^ (u^3-u^2+1-(u+1)q)
```
with two optimized addition chains, a Frobenius power and a hinted division in Fp6.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

`TestPairingCheckTestSolve` test passes.

# How has this been benchmarked?

This PR saves **1,269,173** scs in the emulated PLONK verifier of BW6-761 in a BN254-PLONK.

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1173, opened 2024-06-14
perf(bls12-381): eliminate finalexp ~naively

yelhousni:
# Description

Applying https://eprint.iacr.org/2024/640.pdf as in #1143 and #1155 for BLS12-381 doesn't work (https://github.com/Consensys/gnark/pull/1143#issuecomment-2162921125). Alternatively we can apply theorem 1 of that paper to check `f == c^r` instead. @feltroidprime suggested to check instead that `f^k == c^(q-u)` where `k=(u-1)^2/3` using a Frobenius map and 2 optimized addition chains. This PR implements their idea after pushing `f` to the cyclotomic subgroup. The exponentiation are done then using Granger--Scott algorithm.

TODO: 
- [ ] use `Karabina12345` instead of Granger--Scott. 
- [ ] include `c^(-u)` in the the Miller loop computation.
- [ ] do both

## Type of change

# How has this been tested?

The existing `TestPairingCheckTestSolve` works for this.

This PR saves **681,769 scs** or **199,105 R1CS** in a BLS12-381 pairing.

# Checklist:


ivokub (approved):
I don't get the math, but the implementation looks good :)

PR #1181, opened 2024-06-27
fix: fix OR computation in case one input is constant and other variable

ivokub:
# Description

See the added regression tests. If we had
```go
api.Or(x, 1)
```
then the result was incorrectly computed for both `x=0` and `x=1`. This was because we used `a*(1-b)` for computation not `a+b-ab` for OR. Furthermore, if we know that one input is constant then we do not have to add constraints as the result is always constant `1` if constant input is `1` and `x` if constant input is `0`.

## Type of change

# How has this been tested?

- [x] TestRegressionOr

# Checklist:

### DIFF
--- a/frontend/cs/scs/api.go
+++ b/frontend/cs/scs/api.go
@@ -351,9 +351,6 @@ func (builder *builder) Or(a, b frontend.Variable) frontend.Variable {
 		return 0
 	}
 
-	res := builder.newInternalVariable()
-	builder.MarkBoolean(res)
-
 	// if one input is constant, ensure we put it in b
 	if aConstant {
 		a, b = b, a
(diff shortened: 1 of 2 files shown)


frontend/cs/scs/api.go
```diff
@@ -362,20 +359,14 @@ func (builder *builder) Or(a, b frontend.Variable) frontend.Variable {
...
-			xa: xa.VID,
-			xc: res.VID,
-			qL: qL,
-			qO: builder.tOne,
-		})
-		return res
+		if builder.cs.IsOne(_b) {
+			return 1
+		} else {
+			return a
```
gbotrel, line 365: quick read out of context; are we sure `a` is actually a boolean here?

gbotrel (approved):
lgtm

PR #1191, opened 2024-07-09
Audit/final commit

ThomasPiellard:
# Description

Final commit fixing Openzeppelin findings
<waiting for report link>

## Type of change

# Checklist:


backend/plonk/bls12-377/prove.go
```diff
@@ -696,13 +725,6 @@ func (s *instance) computeLinearizedPolynomial() error {
 
 func (s *instance) batchOpening() error {
 
-	// wait for LRO to be committed (or ctx.Done())
```
ivokub, line 699: Just to confirm - have you checked that this synchronization point was not necessary? Were you able to remove it because we have the next check for the linearizedPolynomial do be done?

backend/plonk/bn254/solidity.go
```diff
@@ -222,6 +223,17 @@ contract PlonkVerifier {
         revert(ptError, 0x64)
       }
 
+      /// Called when an operation on Bn254 fails
+      /// @dev for instance when calling EcMul on a point not on Bn254.
```
ivokub, line 227: Comment invalid? The error is for mod exp, but comment is about EcMul?

backend/plonk/bn254/solidity.go
```diff
@@ -403,30 +415,28 @@ contract PlonkVerifier {
...
+        mstore(add(mPtr, {{ hex $offset }}), VK_QM_COM_X) {{ $offset = add $offset 0x20}}
+        mstore(add(mPtr, {{ hex $offset }}), VK_QM_COM_Y) {{ $offset = add $offset 0x20}}
+        mstore(add(mPtr, {{ hex $offset }}), VK_QO_COM_X) {{ $offset = add $offset 0x20}}
+        mstore(add(mPtr, {{ hex $offset }}), VK_QO_COM_Y) {{ $offset = add $offset 0x20}}
+        mstore(add(mPtr, {{ hex $offset }}), VK_QK_COM_X) {{ $offset = add $offset 0x20}}
+        mstore(add(mPtr, {{ hex $offset }}), VK_QK_COM_Y) {{ $offset = add $offset 0x20}}
         {{ range $index, $element := .CommitmentConstraintIndexes}}
-        mstore(add(mPtr, {{ hex (add 544 (mul $index 64)) }}), VK_QCP_{{ $index }}_X)
-        mstore(add(mPtr, {{ hex (add 576 (mul $index 64)) }}), VK_QCP_{{ $index }}_Y)
+        mstore(add(mPtr, {{ hex (add $offset (mul $index 64)) }}), VK_QCP_{{ $index }}_X)
```
ivokub, line 438: I would do similarly as before `{{ $offset = add $offset 0x20 }}` and then use `$offset` in the next line.

ivokub (approved):
Looks good to me. I have two suggestions essentially, one is just a something nice to have but the second one imo would introduce an invalid comment. But I'm not sure if we can fix it as you changes are already approved.

However, I don't know what to do with the merge conflict - it there seem to have already modifications in the PLONK template and maybe it would apply nicely. In any case I think it needs to be resolved manually because otherwise the merge would mess up the PLONK Solidity template.

backend/plonk/bls12-377/prove.go
ThomasPiellard: yes so in batchOpening we wait for `chLinearizedPolynomial` to be closed,  `computeLinearizedPolynomial` waits for `chH` to be closed, and `computeQuotient` (which closes chH) waits for `s.chLRO` to be closed

backend/plonk/bn254/solidity.go
ThomasPiellard: right it's a mistake

backend/plonk/bn254/unmarshal.go
```diff
@@ -73,7 +71,7 @@ func UnmarshalSolidity(s []byte, nbCommits int) Proof {
 	// uint256[] selector_commit_api_at_zeta;
 	// uint256[] wire_committed_commitments;
 	for i := 0; i < nbCommits; i++ {
-		proof.BatchedProof.ClaimedValues[7+i].SetBytes(s[offset : offset+fr_size])
+		proof.BatchedProof.ClaimedValues[6+i].SetBytes(s[offset : offset+fr_size])
```
gbotrel, line 74: if I'm not mistaken, `UnmarshalSolidity` is untested in our codebase --> can you add a simple round trip test with the `MarshalSolidity` func to ensure non regression / offset errors ?

gbotrel (changes requested):

backend/plonk/bn254/unmarshal.go
ThomasPiellard: so I discussed a bit with OZ about that, in fact UnmarshalSolidity was used for debugging only (at the time there was an issue on testnet on I had to check the proof that was given by linea). At the time MarshalSolidity serialised all the elements of the proof. Now there has been an optimisation such that we don't serialise the opening of the linearised polynomial, it is recomputed in the verifier. To recompute this value we need the public inputs. So UnmarshalSolidity would have a weird signature (we would need to pass the public inputs). I told OZ that this function should be used internally for debugging only, I don't think we should keep it in the repo

gbotrel (approved):

PR #1197, opened 2024-07-11
feat: use blake2 for variable hashcode

### LINKED ISSUE #1146: bug: use cryptographically strong hash function for checking if non-native element has to be width-constrained
## Description

Currently, we use `HashCode` method to store the identifiers of the non-native limbs which are already width-constrained. The method `HashCode` is not cryptographically strong and the circuit designer may be able to reorder the constraints in a way there is a hash collision which would lead to a non-native element limbs not being width constrained. This is particularly relevant for R1CS where the variables are linear expressions and the circuit designer has a lot of leeway to choose inputs.

ivokub:
# Description

Fixes #1146.

Instead of a fast hash non-cryptographic hash use cryptographically secure hash. Previously it was possible for a malicious circuit designer to define a circuit in a way which would omit either boolean marking or non-native element range checking by creating collisions between variables. This PR avoids this by using a strong hash function instead.

NB! It uses as hash outputs only 16 bytes instead of full 32 bytes for saving some memory.

## Type of change

# How has this been tested?

Existing tests pass

For example, I compiled the BW6-671 PLONK verifier in BN254 as one of the biggest circuits to get estimates in possible performance bottlenecks. Results:
```
# PRE patch

SCS:
* time 37.39810504s
* mem 32498MB

R1CS:
* time 46.508562287s
* mem 65606MB

# POST patch

SCS:
* time 37.215583182s
* mem 32430MB

R1CS:
* time 47.082183625s
* mem 66455MB
```

So it seems that for SCS it actually improves performance (but only very little). For R1CS we have approx 1% more memory allocations and 1% slower compiler. I think we could also try to optimize the hash code computation a little by using fixed buffers and hasher as the compiler is sequential anyway. And we can provide the destination to store the hash output instead of allocating new.

This makes the code a little less readable though, but may be worth it?

# Checklist:

### DIFF
--- a/frontend/cs/r1cs/builder.go
+++ b/frontend/cs/r1cs/builder.go
@@ -58,7 +58,7 @@ type builder struct {
 	kvstore.Store
 
 	// map for recording boolean constrained variables (to not constrain them twice)
-	mtBooleans map[uint64][]expr.LinearExpression
+	mtBooleans map[[16]byte][]expr.LinearExpression
 
 	tOne        constraint.Element
 	eZero, eOne expr.LinearExpression
@@ -82,7 +82,7 @@ func newBuilder(field *big.Int, config frontend.CompileConfig) *builder {
 		macCapacity = config.CompressThreshold
 	}
 	builder := builder{
-		mtBooleans: make(map[uint64][]expr.LinearExpression, config.Capacity/10),
+		mtBooleans: make(map[[16]byte][]expr.LinearExpression, config.Capacity/10),
 		config:     config,
 		heap:       make(minHeap, 0, 100),
 		mbuf1:      make(expr.LinearExpression, 0, macCapacity),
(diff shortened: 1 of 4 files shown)


gbotrel (commented):
lgtm 🚀 

edit: for the benchmark, did you benchmark on your machine or on a large one? i.e. do you have more than 64GB of ram? If not can you try on a machine with enough ram just to be sure the CPU measurement is not fake (i.e. memswaps) (I recall trying couple of hash functions for the hash code and it having a significant perf impact at compile time)

ivokub:
> the impact of doing a collision resistant hashcode was if I recall, quite high in term of perf for the compile phase of a large circuit. 
> Can you do a quick benchmark (frontend.Compile) of a large circuit before / after the PR? 
> 
> edit: sorry missed the benchmark in the description

Yup, I also recall that it was supposed to be slower, and tested runtime and profiled memory usage. I guess with Blake and 16 byte output its still reasonable.

I could test binary heavy stuff though separately (with a lot of MarkBoolean) to confirm

gbotrel (approved):

PR #1200, opened 2024-07-15
fix: variable modulus subtraction padding

### LINKED ISSUE #1152: bug: subtraction padding in variable-modulus non-native arithmetic does not enforce underflow protection
## Description

The subtraction padding should ensure that we do not underflow when subtracting limbs. For fixed-modulus case we can compute the padding as a constant such that it always is bigger than the value we subtract, but for variable-modulus case we compute the padding on the go. We currently use the same function for the padding computation and add a range check on the limbs of the padding. However, we do not add assertion that it is actually sufficiently big for avoiding underflows.

There are several approaches we can take. We either try to fix it or omit the method and dependent methods.

For the second approach, we do not actually really use it in practice - we need subtraction for var-mod equality assertion which we only use in the tests for checking the results. So there shouldn't be a problem in making the methods private.

For the first approach, we can fix it by checking that `pad[i]` is greater than `1 << (fp.BitsBerLimb()+overflow+1)`. And in this case it would really make sense to cache the padding values for different `overflow` values so that we wouldn't add new constraints every time (padding depends on the overflow).

------

There is additional bug in the subtraction padding hint. There is
```go
padding := f.newInternalElement(res, fp.BitsPerLimb()+overflow+1)
```
but should be
```go
padding := f.newInternalElement(res, overflow+1)
```

------

Another bug is that we do not check that `nextOverflow` [here](https://github.com/Consensys/gnark/blob/master/std/math/emulated/custommod.go#L50) doesn't overflow the native field

ivokub:
# Description

This PR fixes some of the methods used in testing the variable modulus case in field emulation:
* correctly set the next overflow
* assert that the subtraction padding is big enough
* check that the next overflow wouldn't the native field

We currently only panic when the next overflow would overflow the native field, in contrast to the behaviour with fixed emulated modulus where we then would mod-reduce the value automatically. As `modSub` is not directly exposed and used in equality assertion (which is then only used in tests), then it should be sufficient for now.

Fixes #1152

## Type of change

# How has this been tested?

Existing tests pass.

# Checklist:

### DIFF
--- a/std/math/emulated/hints.go
+++ b/std/math/emulated/hints.go
@@ -188,16 +188,33 @@ func subPaddingHint(mod *big.Int, inputs, outputs []*big.Int) error {
 }
 
 func (f *Field[T]) computeSubPaddingHint(overflow uint, nbLimbs uint, modulus *Element[T]) *Element[T] {
+	// we compute the subtraction padding hint in-circuit. The padding has satisfy:
+	// 1. padding % modulus = 0
+	// 2. padding[i] >= (1 << (bits+overflow))
+	// 3. padding[i] + a[i] < native_field for all valid a[i] (defined by overflow)
 	var fp T
 	inputs := []frontend.Variable{fp.NbLimbs(), fp.BitsPerLimb(), overflow, nbLimbs}
 	inputs = append(inputs, modulus.Limbs...)
+	// compute the actual padding value
 	res, err := f.api.NewHint(subPaddingHint, int(nbLimbs), inputs...)
 	if err != nil {
 		panic(fmt.Sprintf("sub padding hint: %v", err))
 	}
+	maxLimb := (1 << (int(fp.BitsPerLimb()) + int(overflow))) - 1
 	for i := range res {
+		// ensure that condition 3 always holds. As we compute the padding hint
+		// depending on the overflow of the inputs, then can use the maximum
+		// possible value
 		f.checker.Check(res[i], int(fp.BitsPerLimb()+overflow+1))
+		// to ensure that condition 2 holds we subtract the maximum possible
+		// value from the padding and check that the result is small (fits into
+		// expected range). If we would have underflow then the result would
+		// underflow the native field otherwise and the range check wouldn't
+		// hold.
+		f.checker.Check(f.api.Sub(res[i], maxLimb), int(fp.BitsPerLimb()+overflow+1))
 	}
+
+	// ensure that condition 1 holds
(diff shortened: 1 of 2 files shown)


std/math/emulated/hints.go
```diff
@@ -188,17 +188,35 @@ func subPaddingHint(mod *big.Int, inputs, outputs []*big.Int) error {
...
 	if err != nil {
 		panic(fmt.Sprintf("sub padding hint: %v", err))
 	}
+	maxLimb := new(big.Int).Lsh(big.NewInt(1), fp.BitsPerLimb()+overflow)
+	maxLimb.Sub(maxLimb, big.NewInt(1))
 	for i := range res {
+		// ensure that condition 3 always holds. As we compute the padding hint
+		// depending on the overflow of the inputs, then can use the maximum
+		// possible value
 		f.checker.Check(res[i], int(fp.BitsPerLimb()+overflow+1))
```
gbotrel, line 209: @ivokub  so, I read, using range checks, ensure:
1. res[i] < ( 1 << (bits_with_overflow + 1) )
2. res[i] -  ( 1 << bits_with_overflow) + 1 < ( 1 << (bits_with_overflow + 1) )

Doesn't 2. suffice (i.e. isn't 1 a sub constraint of 2?)

ivokub: I think you are right, I'll fix it.

gbotrel: there might be some edge cases with very small emulated and scalar field and or very large, but, didn't find any since I assume overflow is correctly computed and won't allow for bogus values (don't take my word for it just spent couple minutes on it)

ivokub: So there are two things what we want to check:
1. that the width of the padding limb is not bigger than `nbBits+overflow+1` (i.e. `padding < 2^{nbBits+overflow+1}`) -- we need to ensure this so that when we add the padding to `a[i]` that we wouldn't overflow the native field.
2. that the padding limb is not smaller than maximum permitted value for non-native element limb i.e. `2^{nbBits+overflow}-1` -- this is to ensure that we wouldn't underflow the circuit when subtracting `b[i]` from padded `a[i] + s[i]`

The first range check checks condition 1 directly. For the second one we instead subtract the maximum permitted value and check that the difference is not big (it is big when we would have underflown).

But you are right, I think we can merge the two checks together

ivokub: Removed one range check and tried to clarify the comments.

gbotrel (approved):

PR #1204, opened 2024-07-16
feat: add IsOnG2 for BN254

ivokub:
# Description

This method allows to check for non-membership. In case we need to assert membership it is still more efficient to use `AssertIsOnG2` as then we don't have to call `IsZero` for non-native elements which is more expensive that `AssertIsEqual`.

@yelhousni - can you recommend better naming for the methods? I think this could generalize to other curves, but right now needed for Linea, so omitted the implementation.

## Type of change

# How has this been tested?

- [x] TestIsOnTwistSolve
- [x] TestIsOnG2Solve

# Checklist:

### DIFF
--- a/std/algebra/emulated/sw_bn254/pairing.go
+++ b/std/algebra/emulated/sw_bn254/pairing.go
@@ -306,7 +306,7 @@ func (pr Pairing) AssertIsOnG1(P *G1Affine) {
 	pr.AssertIsOnCurve(P)
 }
 
-func (pr Pairing) computeSubgroupEquation(Q *G2Affine) (_Q *G2Affine) {
+func (pr Pairing) computeG2ShortVector(Q *G2Affine) (_Q *G2Affine) {
 	// [x₀]Q
 	xQ := pr.g2.scalarMulBySeed(Q)
 	// ψ([x₀]Q)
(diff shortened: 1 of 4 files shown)


std/algebra/emulated/sw_bn254/pairing.go
```diff
@@ -273,33 +273,40 @@ func (pr Pairing) AssertIsOnCurve(P *G1Affine) {
...
 	pr.AssertIsOnCurve(P)
 }
 
-func (pr Pairing) AssertIsOnG2(Q *G2Affine) {
-	// 1- Check Q is on the curve
-	pr.AssertIsOnTwist(Q)
-
-	// 2- Check Q has the right subgroup order
-
+func (pr Pairing) computeSubgroupEquation(Q *G2Affine) (_Q *G2Affine) {
```
yelhousni, line 309: For naming, I would call this method `computeG2ShortVector` as in last column of table 2 in https://eprint.iacr.org/2022/348.pdf

yelhousni (approved):
LGTM

std/algebra/emulated/sw_bn254/pairing.go
ivokub: Thanks! Sounds better and is more accurate.

PR #1207, opened 2024-07-17
perf: optimize class equivalence check for BLS12 final exp 

yelhousni:
# Description

Instead of applying Th.1 of https://eprint.iacr.org/2024/640.pdf naively as in #1173 ans #1202 we can scale the miller loop result by some factors:
- for BLS12-381 by 27th root and `p`-th root where `p=(1-u)/3`
- for BLS12-377 by `p'`-th root where `p'=12(u-1)`

 and use the optimal exponent `q-u` instead of `r`. This is based on a personal communication (and a lot of help) from Andrija Novakovic @akinovak: https://gist.github.com/akinovak/0db531d350b95ccec682666b2257db77

## Type of change

# How has this been tested?

pairing tests pass.

- BLS12-381:
This PR saves **3,369,246 scs** for a `PairingCheck`.
- BLS12-377:
This PR saves **21,989 scs** in the `PairingCheck` and in the PLONK native aggregation circuit.

# Checklist:


ivokub (approved):
LGTM

std/algebra/emulated/fields_bls12381/e12_pairing.go
```diff
@@ -387,11 +387,12 @@ func (e Ext12) FrobeniusSquareTorus(y *E6) *E6 {
 // FinalExponentiationCheck checks that a Miller function output x lies in the
 // same equivalence class as the reduced pairing. This replaces the final
 // exponentiation step in-circuit.
-// The method follows Section 4 of [On Proving Pairings] paper by A. Novakovic and L. Eagen.
+// The method is inspired from [On Proving Pairings] paper by A. Novakovic and
+// L. Eagen, and is based on a personal communication with A. Novakovic.
 //
 // [On Proving Pairings]: https://eprint.iacr.org/2024/640.pdf
 func (e Ext12) FinalExponentiationCheck(x *E12) *E12 {
```
ivokub, line 394: We have denoted return type, but don't return anything. Remove returned value.

std/algebra/native/fields_bls12377/e12_pairing.go
```diff
@@ -176,3 +178,69 @@ func (e *E12) ExpX0Minus1Square(api frontend.API, e1 E12) *E12 {
...
+	return e
+}
+
+// FinalExponentiationCheck checks that a Miller function output x lies in the
+// same equivalence class as the reduced pairing. This replaces the final
+// exponentiation step in-circuit.
+// The method follows Section 4 of [On Proving Pairings] paper by A. Novakovic and L. Eagen.
+//
+// [On Proving Pairings]: https://eprint.iacr.org/2024/640.pdf
+func (x *E12) FinalExponentiationCheck(api frontend.API) *E12 {
```
ivokub, line 215: Remove returned value.

ivokub (changes requested):
Changes good. I only now noticed that in `FinalExpCheck` we have indicated that we return `*E12`, but always return nil. I think we can change the the function signature to indicate that.

And, now after https://github.com/Consensys/gnark/pull/1209 is merged, maybe it would make sense to also have methods for other curves for returning the FinalExpCheck result? Long-term this allows us to amend the `std/algebra/Pairing` interface so that we can also prove that final exponentiation doesn't hold.

std/algebra/emulated/fields_bls12381/e12_pairing.go
yelhousni: done. also for other curves.

std/algebra/native/fields_bls12377/e12_pairing.go
yelhousni: done

ivokub (approved):
Thanks!

PR #1214, opened 2024-07-19
fix: avoid malicious hint in BN254 final exp check

### LINKED ISSUE #1213: bug: malicious hint in BN254 final exp check
## Description
reported in https://github.com/Consensys/gnark/pull/1143#pullrequestreview-2150219385 by [feltroidprime](https://github.com/feltroidprime).
TL;DR: we check in-circuit that `f * w  == c^λ` where `c` and `lambda` come from a hint. A malicious prover can provide the incorrect hints `w = c^λ/f` and `c` and the check will pass. The original [paper](https://eprint.iacr.org/2024/640.pdf) suggests in 7.1 to check in circuit that `(1-w)(f-w)(f^2-w) == 0` where `f^i` is the non-trivial 3rd root of unity. The paper then proves that `w` can only be in Fp3 and suggest to constraint it so. 

## Possible Fix

Since `w = w0+w1*u` is in Fp3, then `w^(p^3) = w` and `w^(p^6) = w`. On the other hand `w^(p^6) = conj(w) = w0 - w1*u` and thus `w = w0` in Fp6. We can force in-circuit `w1` to be 0 and retrieve `w0` from the hint:
```go
cubicNonResiduePower := E12{
    C0: E6{
                B0: E2{A0: *res[12], A1: *res[13]},
                B1: E2{A0: *res[14], A1: *res[15]},
                B2: E2{A0: *res[16], A1: *res[17]},

    },
    C1: (*pr.Ext6.Zero()),
}
```

yelhousni:
# Description

Fixes #1213 

## Type of change

# How has this been tested?

Same tests.

no change.

# Checklist:

### DIFF
--- a/std/algebra/emulated/fields_bn254/e12_pairing.go
+++ b/std/algebra/emulated/fields_bn254/e12_pairing.go
@@ -447,17 +447,14 @@ func (e Ext12) FinalExponentiationCheck(x *E12) *E12 {
 			B2: E2{A0: *res[10], A1: *res[11]},
 		},
 	}
+	// constrain cubicNonResiduePower to be in Fp6
 	cubicNonResiduePower := E12{
 		C0: E6{
 			B0: E2{A0: *res[12], A1: *res[13]},
 			B1: E2{A0: *res[14], A1: *res[15]},
 			B2: E2{A0: *res[16], A1: *res[17]},
 		},
-		C1: E6{
-			B0: E2{A0: *res[18], A1: *res[19]},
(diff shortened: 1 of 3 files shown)


ivokub (approved):
LGTM

ivokub (approved):
I think it is sufficient that the hint returns error when previous Miller Loop result is zero. Imo it cannot really happen in any valid cases and would probably mean that the input wasn't previous Miller loop result or the accumulator was not correctly initialized.

PR #1218, opened 2024-07-23
fix: remove unconstrained and unused variables

ivokub:
# Description

PR #1214 removed usage of possible malicious hint value from the circuit. However, we still output the variables in circuit which are then enforced to be range checked. In gnark it leads to unnecessary constraints and in zkevm it creates problem with the external range checker as we're looking for constraints where this variables are used.

## Type of change

# How has this been tested?

Current CI. Only cosmetic change.

# Checklist:


yelhousni (approved):
makes sense

Issue #1147, opened 2024-05-29
bug: describe in emulated `Reduce` method that the result may be either `x` or `x+p`


ivokub:
The behaviour is described in `AssertIsInRange` method documentation but not in `Reduce` method documentation. We also have to add a part to `field.ToBits` method to describe how to obtain binary decomposition of the canonical value of non-native element.

PR #1224, opened 2024-07-24
fix: strict ModReduce in emulated fields

ivokub:
# Description

Fixes https://github.com/Consensys/gnark/issues/1147, but more thoroughly. Added separate method which performs modular reduction and asserts that the result is in range. Updated the documentation accordingly.

Additionally, added an option to algebra methods to obtain the canonical bit representation if necessary (when comparing bitwise some values).

The current implementation are not modified - if the prover would provide `x+p` from the hint, then the recursion implementations fail to solve as we have incorrect FS challenges.

We now have reduction in the 2-chain PLONK verifier as previously when performing binary decomposition of BLS12 points we used canonical decomposition, but now by default we don't. 

## Type of change

# How has this been tested?

- [x] TestReduceStrict
- [x] TestToBitsCanonical
- [x] TestIsZeroEdgeCases

- [x] PLONK verify BLS12-377 in BW6-761. Was 358153 now 322251
- [x] PLONK verify BW6-761 in BN254. Was 28933345 now 28215408.

# Checklist:

### DIFF
--- a/std/math/emulated/field_assert.go
+++ b/std/math/emulated/field_assert.go
@@ -50,8 +50,7 @@ func (f *Field[T]) AssertIsEqual(a, b *Element[T]) {
 }
 
 // AssertIsLessOrEqual ensures that e is less or equal than a. For proper
-// bitwise comparison first reduce the element using [Reduce] and then assert
-// that its value is less than the modulus using [AssertIsInRange].
+// bitwise comparison first reduce the element using [Field.ReduceStrict].
 func (f *Field[T]) AssertIsLessOrEqual(e, a *Element[T]) {
 	// we omit conditional width assertion as is done in ToBits below
 	if e.overflow+a.overflow > 0 {
--- a/std/math/emulated/field_assert.go
+++ b/std/math/emulated/field_assert.go
@@ -91,8 +91,13 @@ func (f *Field[T]) AssertIsLessOrEqual(e, a *Element[T]) {
 // it is not. For binary comparison the values have both to be below the
 // modulus.
 func (f *Field[T]) AssertIsInRange(a *Element[T]) {
+	// short path - this element is already enforced to be less than the modulus
+	if a.modReduced {
+		return
+	}
 	// we omit conditional width assertion as is done in ToBits down the calling stack
 	f.AssertIsLessOrEqual(a, f.modulusPrev())
+	a.modReduced = true
 }
 
 // IsZero returns a boolean indicating if the element is strictly zero. The
--- a/std/math/emulated/field_assert.go
+++ b/std/math/emulated/field_assert.go
@@ -104,8 +104,15 @@ func (f *Field[T]) AssertIsInRange(a *Element[T]) {
 // method internally reduces the element and asserts that the value is less than
 // the modulus.
 func (f *Field[T]) IsZero(a *Element[T]) frontend.Variable {
+	// to avoid using strict reduction (which is expensive as requires binary
+	// assertion that value is less than modulus), we use ordinary reduction but
+	// in this case the result can be either 0 or p (if it is zero).
+	//
+	// so we check that the reduced value limbs are either all zeros or
+	// corrspond to the modulus limbs.
 	ca := f.Reduce(a)
-	f.AssertIsInRange(ca)
+	p := f.Modulus()
+
 	// we use two approaches for checking if the element is exactly zero. The
 	// first approach is to check that every limb individually is zero. The
 	// second approach is to check if the sum of all limbs is zero. Usually, we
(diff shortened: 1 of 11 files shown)


ivokub:
Converting to draft until bugs fixed.

std/math/emulated/field_assert.go
```diff
@@ -109,23 +120,32 @@ func (f *Field[T]) IsZero(a *Element[T]) frontend.Variable {
 	// then we can ensure in most cases that no overflows happen.
 
 	// as ca is already reduced, then every limb overflow is already 0. Only
-	// every addition adds a bit to the overflow
+	// every addition adds a bit to the overflow.
+	var res0 frontend.Variable
 	totalOverflow := len(ca.Limbs) - 1
 	if totalOverflow > int(f.maxOverflow()) {
```
gbotrel, line 126: it seems to me this branch will never be taken; didn't we just reduce `ca` couple of lines above? why would we have an overflow larger than the field?
I added a panic in this branch, ran some unit tests (not sure I ran the right ones) and the panic wasn't triggered.

std/math/emulated/field_assert.go
```diff
@@ -109,23 +120,32 @@ func (f *Field[T]) IsZero(a *Element[T]) frontend.Variable {
...
+		res0 = f.api.IsZero(ca.Limbs[0])
+		for i := 1; i < len(ca.Limbs); i++ {
+			res0 = f.api.Mul(res0, f.api.IsZero(ca.Limbs[i]))
+		}
+	} else {
+		// default case, limbs sum does not overflow the native field
+		limbSum := ca.Limbs[0]
 		for i := 1; i < len(ca.Limbs); i++ {
-			res = f.api.Mul(res, f.api.IsZero(ca.Limbs[i]))
+			limbSum = f.api.Add(limbSum, ca.Limbs[i])
```
gbotrel, line 137: doesn't this work only if `len(ca.limbs) * max limb value <= max snark scalar field value` ?

std/math/emulated/field_binary.go
```diff
@@ -34,6 +33,24 @@ func (f *Field[T]) ToBits(a *Element[T]) []frontend.Variable {
...
+	// TODO: implement a inline version of this function. We perform binary
+	// decomposition both in the `ReduceStrict` and `ToBits` methods, but we can
+	// essentially do them at the same time.
+	//
+	// If we do this, then also check in places where we use `Reduce` and
+	// `ToBits` after that manually (e.g. in point and scalar marshaling) and
+	// replace them with this method.
+
+	var fp T
+	nbBits := fp.Modulus().BitLen()
```
gbotrel, line 48: (random thought; I think most modulus we use have the same number of bits for q and q-1, but if not, we could use one less bit here)

std/math/emulated/field_binary.go
```diff
@@ -34,6 +33,24 @@ func (f *Field[T]) ToBits(a *Element[T]) []frontend.Variable {
...
+	ca := f.ReduceStrict(a)
+	bts := f.ToBits(ca)
```
gbotrel, line 50: don't we have an option in `ToBits` to give the number of bits we want and reduce the number of constraints generated by `ToBits`?

std/math/emulated/field_reduce.go
```diff
@@ -0,0 +1,53 @@
...
+// In cases where the canonical representation of the element is required, use
+// [Field.ReduceStrict].
+func (f *Field[T]) Reduce(a *Element[T]) *Element[T] {
+	ret := f.reduce(a, false)
+	return ret
+}
+
+func (f *Field[T]) reduce(a *Element[T], strict bool) *Element[T] {
+	f.enforceWidthConditional(a)
+	if strict && a.modReduced {
```
gbotrel, line 18: why do we check if the `strict` flag is provided? 
if `modReduced` is set, both in Reduce and Reduce strict, we should return `a`

gbotrel (changes requested):

std/math/emulated/field_reduce.go
ivokub: Fixed.

std/math/emulated/field_assert.go
ivokub: I tried to future-proof a little. Currently the native fields (and thus `f.maxOverflow()`) are quite big, so this case normally wouldn't be taken. However, when we would be using smaller fields e.g. Goldilocks on 64 bits, then for example when we would emulate `Mod1e4096` then this case could potentially be taken.

But if it seems to be early then can remove.

std/math/emulated/field_assert.go
ivokub: Exactly - this is the guard `totalOverflow > int(f.maxOverflow())` checks against. Every limb in non-native Element is up to `nbBits+overflow` bits wide, where `nbBits` is constant and `overflow` is tracked through arithmetic operations.

Every time we do an addition between `nbBits+overflow_1` and `nbBits+overflow_2` wide limbs, the resulting limbs are up to `nbBits+max(overflow_1, overflow_2)+1` bits wide. In this method, as the overflow is same for all limbs, then after we add `n` limbs the sum is up to `nbBits+overflow+n-1` bits wide. This is exactly the computed value `totalOverflow` which we compare against maximum permitted overflow computed by `f.maxOverflow()`. So `limbSum` should never overflow native field in this branch.

std/math/emulated/field_binary.go
ivokub: Added.

std/math/emulated/field_binary.go
ivokub: Currently for `Field.ToBits` we don't and I think that it would be a good addition.

But I think I would do it in a separate issue/PR to additionally also handle the TODO in this method. Imo in places we currently use the `ToBits` method we use all bits.

ivokub:
Fixes pushed.

PR #1228, opened 2024-07-25
fixes #1227: api.AssertIsLessOrEqual incorrect behavior on R1CS with constant variable

### LINKED ISSUE #1227: bug: `api.AssertIsLessOrEqual` incorrect behavior on R1CS with constant variable
```

import (
	"testing"

	"github.com/consensys/gnark/test"
)

type Circuit struct {
	X frontend.Variable
}

func (circuit *Circuit) Define(api frontend.API) error {
	api.AssertIsLessOrEqual(1, circuit.X)
	return nil
}

func TestConstantPath(t *testing.T) {
	assert := test.NewAssert(t)
	assert.CheckCircuit(&Circuit{},
		test.WithValidAssignment(&Circuit{X: 1}),   // 1 <= 1 --> true
		test.WithInvalidAssignment(&Circuit{X: 0})) // 1 <= 0 --> false
}

```

thanks to @wuestholz for reporting 

Fails because of a bug in the R1CS compiler when the first parameter to `api.AssertIsLessOrEqual` is a constant.

gbotrel:


frontend/cs/r1cs/api_assertions.go
```diff
@@ -152,28 +156,18 @@ func (builder *builder) mustBeLessOrEqVar(a, bound frontend.Variable) {
...
-			// aBits[i] is a constant;
-			l = builder.Mul(l, aBits[i])
-			// TODO @gbotrel this constraint seems useless.
-			added = append(added, builder.cs.AddR1C(builder.newR1C(l, zero, zero), builder.genericGate))
-		} else {
-			added = append(added, builder.cs.AddR1C(builder.newR1C(l, aBits[i], zero), builder.genericGate))
-		}
+		// (1 - t - ai) * ai == 0
+		l := builder.Sub(builder.cstOne(), t, aBits[i])
+		added = append(added, builder.cs.AddR1C(builder.newR1C(l, builder.Mul(aBits[i], builder.cstOne()), zero), builder.genericGate))
```
ivokub, line 170: Why here `builder.Mul(aBits[i], builder.cstOne())`? Isn't it noop?

ivokub (approved):
Looks good. See the comment, maybe makes sense.

frontend/cs/r1cs/api_assertions.go
gbotrel: so, when `a` is a constant, and is on one bit, there is now a fast path in `ToBinary` that just returns the input; and then, giving a constant directly here (not an expr.LinearExpression) to `newR1C` screw things up :)

ivokub: Of course, I see it now

ivokub:
All comments resolved, good to merge 👍 . Thanks @wuestholz for reporting!

PR #1234, opened 2024-08-01
fix: branch with unchecked cast could panic at compile time

gbotrel:
branch with unchecked cast could panic at compile time, thanks to @wuestholz  for reporting.


ivokub (approved):
LGTM

PR #1245, opened 2024-08-13
feat add random mask to groth16 commitment

Tabaie:
This PR adds a random mask to the BSB22 commitments in order to make them hiding.

It is achieved simply by randomizing a new variable and adding it to the commitment.

## Type of change

# Checklist:


Tabaie:
@ivokub `TestNoCommitmentVariable` is failing because no commitment is truly empty anymore. Is that a desired feature or can we just remove the test?

frontend/cs/r1cs/api.go
```diff
@@ -687,6 +688,19 @@ func (builder *builder) Compiler() frontend.Compiler {
 
 func (builder *builder) Commit(v ...frontend.Variable) (frontend.Variable, error) {
 
+	// add a random mask to v
+	{
```
ivokub, line 692: As far as I understand, doing it this way adds a randomized value every time we call a commit? But in case we call it many times it could add a lot of redundant constraints/vars.

ivokub: Aah, I realized that `Commit` always adds a commitment indeed and we have to use multicommit anyway. So no issue here.

ivokub:
> @ivokub `TestNoCommitmentVariable` is failing because no commitment is truly empty anymore. Is that a desired feature or can we just remove the test?

I made the test to expect provers success instead. It can be useful in the future to catch regressions. I think it is useful that we can have prover assigned random commitment if we don't commit to anything.

ivokub (approved):
I think it makes all sense! I added a regression test also.

PR #1269, opened 2024-09-04
feat constant 'randomness' - domain-size-independent vk

Tabaie:
This PR fixes (i.e. always the same value) the randomness used in unsafe KZG SRS production. This results in getting the same SRS for a given domain every time, and even the same verifying key for different domain sizes over the same curve.
The intended application of this are tests involving batch verification of commitments - for example when verifying multiple PlonK proofs.

## Type of change

# Checklist:

### DIFF
--- a/test/unsafekzg/kzgsrs.go
+++ b/test/unsafekzg/kzgsrs.go
@@ -6,7 +6,7 @@ package unsafekzg
 
 import (
 	"bufio"
-	"crypto/rand"
+	"crypto/sha256"
 	"fmt"
 	"math/big"
 	"os"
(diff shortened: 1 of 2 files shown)


ivokub:
I would instead use an option for this `a la WithAlpha` etc. and keep the default behaviour to be randomized. I think if we by default use fixed alpha then it may obscure some other errors (e.g. some deserialized values are static when they are not supposed to be).

gbotrel:
+1 with @ivokub suggestion, just also on the off chance that some cavalier users, by mistake, do use this SRS in prod.

ivokub:
I added an option to set the toxic value. @gbotrel can you review?

gbotrel (approved):

Issue #1159, opened 2024-06-07
feat: make `std/algebra/native/twistededwards` to implement `std/algebra.Curve` interface.


ivokub:
We currently do not use directly inside gnark, but would be better and safer if we allow using only through the interface. Particularly, it enforces that the scalar arithmetic is done using non-native arithmetic.

PR #1271, opened 2024-09-06
Fix: Bandersnatch GLV scalar multiplication

### LINKED ISSUE #268: Hinted scalar decomposition in a gadget
In SW and tEd gadgets, when using the hinted scalar decomposition for GLV, we only check that `s1 + λ * s2 == s + k*o` but this is done `mod r` (where `o` is the order of SW/tEd curve prime subgroup and `r` the order of the SNARK curve prime subgroup).

It means that in Z: `s1 + λ * s2 - s - k*o = c * r` but with no guarantee that `c` is always `0`.  We should look for edge cases if any and constrain them or look for a better constrained decomposition.

yelhousni:
# Description

Fixes #268

## Type of change

# How has this been tested?

`TestCurve` for Bandersnatch pass when using `scalarMulGLV`.

`Bandersnatch` in R1CS:
2-bit windowed double-and-add: **3,314**
GLV: **2,735** 
`Bandersnatch` in SCS:
2-bit windowed double-and-add: **5,991**
GLV: **6,077** 

So the GLV saves 579 r1cs but adds 86 scs, which is because of the non-native scalar decomposition check. In this PR we use GLV only if endomorphism + R1CS.

# Checklist:


gbotrel (approved):
A remark, that probably applies to other neighbor pacakges;

I checked the call graph to see how it is tested, and it seems we just test with `testData` --> with random scalars. 

Should we add some edge cases tests ? (i.e. scalar at 0, p-1, 1, ...),

ivokub (approved):
Looks perfect! I only added some comments what my local linter mentioned (unused arguments in methods).

ivokub (approved):
Looks good again. Maybe we should add a comment to the `Curve` interface that the methods ScalarMul, Add etc. do not check that point is on the curve.

For the future, it would be nice to also implement this - https://github.com/Consensys/gnark/issues/1159. There was actually a discussion related to it just recently - https://github.com/Consensys/gnark/discussions/1292

yelhousni:
Closing since superseded by the fake GLV

PR #1276, opened 2024-09-13
feat: replace stats gob format with csv for easier diffs

gbotrel:
# Description

Stores snippets stats in CSV format to get readable diffs in CI / tests.


internal/stats/stats.go
```diff
@@ -49,16 +31,45 @@ func NewGlobalStats() *globalStats {
 	}
 }
 
-func (s *globalStats) Save(path string) error {
-	fStats, err := os.Create(path) //#nosec G304 -- ignoring internal package s
-	if err != nil {
-		return err
+func (s *globalStats) WriteTo(w io.Writer) (int64, error) {
```
ivokub, line 34: This approach works. Imo for me it would make more sense for `snippetStats` to be `type struct {Name string, Backend backend.ID, Curve curve.ID, NbConstraints int, NbVariables int}`. And then we would have to implement comparison on it to get deterministic output. And then imo everything gets easier (dont have the `map[][][]snippetStats` etc.)

ivokub: But I'm also fully OK to keep it as is. I think this is very static and independent component, so won't have to think about too much.

ivokub (approved):
LGTM. One nice to have comment, but I think lets leave it as is.

PR #1288, opened 2024-10-04
fix: fix slice init length

cuishuang:
# Description

The intention here should be to initialize a slice with a capacity of  `len(prof.Location)`  rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/profile/internal/graph/graph.go
+++ b/profile/internal/graph/graph.go
@@ -438,7 +438,7 @@ func newTree(prof *profile.Profile, o *Options) (g *Graph) {
 		}
 	}
 
-	nodes := make(Nodes, len(prof.Location))
+	nodes := make(Nodes, 0, len(prof.Location))
 	for _, nm := range parentNodeMap {
 		nodes = append(nodes, nm.nodes()...)
 	}


ivokub:
Thanks for the PR. Right now it indeed seems that we initialize the nodes differently. I'll further check if we don't somehow assume that first `len(prof.Location)` values are there for inserting something, but currently it looks like not.

cuishuang:
> Thanks for the PR. Right now it indeed seems that we initialize the nodes differently. I'll further check if we don't somehow assume that first `len(prof.Location)` values are there for inserting something, but currently it looks like not.

I have searched many Go repositories, and this is a common issue. I am trying to solve it at the Go toolchain level

ivokub:
> > Thanks for the PR. Right now it indeed seems that we initialize the nodes differently. I'll further check if we don't somehow assume that first `len(prof.Location)` values are there for inserting something, but currently it looks like not.
> 
> I have searched many Go repositories, and this is a common issue. I am trying to solve it at the Go toolchain level

Interesting! Imo it is a very good find. Indeed we have based our profiling implementation on the Go standard one, but capturing different metrics.

ivokub:
I checked - in gnark this code path is never called and cannot be called without modifying the internal implementation. And even if it is initalized incorrectly, then later in `selectNodesForGraph` the nil entries are filtered out. But the issue is valid in any case.

Thanks for reporting!

ivokub (approved):
LGTM!

PR #1291, opened 2024-10-07
docs: update api doc following audit suggestions

gbotrel:
# Description

- Minor API doc edits
- Remove scs/api.go api comments; godoc will use the one defined in the interface by default which is better.


ivokub (approved):
LGTM. Covers all the reported issues.

PR #1297, opened 2024-10-11
Groth 16 MPC setup refactor

Tabaie:
This PR updates the Groth16 MPC setup to support BSB22 commitments.


ivokub:
It seems the CI check fails due to https://github.com/Consensys/gnark-crypto/pull/585 - here now in both cases encode and encodeRaw we encode the G1/G2 points as compressed, but Solidity checker assumes uncompressed it seems. When I reverted gnark-crypto upgrade then the failing test started to pass. But I think it is better to fix the other PR.

Tabaie:
> It seems the CI check fails due to [Consensys/gnark-crypto#585](https://github.com/Consensys/gnark-crypto/pull/585) - here now in both cases encode and encodeRaw we encode the G1/G2 points as compressed, but Solidity checker assumes uncompressed it seems. When I reverted gnark-crypto upgrade then the failing test started to pass. But I think it is better to fix the other PR.

Yes that's it. Thank you! Just pushed a fix to the other PR.

Tabaie:
Closed in favor of https://github.com/Consensys/gnark/pull/1372

PR #1305, opened 2024-10-25
chore: Replace fmt.Printf calls with warning logs

wwared:
# Description

As far as I can tell, there's only two places (outside of tests) that still are using `fmt.Printf`/`fmt.Println` to print warnings to stdout, in the `frontend/schema` directory. This small PR replaces them with warning logs using the `logger` module.

If the severity of the message should be different from a warning, let me know so I can change it quickly.

## Type of change

# How has this been tested?

I've tested this locally against a test case that triggers this message (specifically https://github.com/argumentcomputer/sphinx doing a Plonk BN254 proof verification, via an FFI wrapper at https://github.com/argumentcomputer/lurk-hs), and with this patch the messages are now logged at the proper level, as follows:

Before:
```
Testing fixture: "/home/w/dev/w/lurk-hs/assets/epoch_change.json"
ignoring uninitialized slice: Vars []frontend.Variable
ignoring uninitialized slice: Vars []frontend.Variable
ignoring uninitialized slice: Vars []frontend.Variable
Test completed for fixture: "/home/w/dev/w/lurk-hs/assets/epoch_change.json"
```
After:
```
Testing fixture: "/home/w/dev/w/lurk-hs/assets/epoch_change.json"
17:15:32 WRN ignoring uninitialized slice slice name=Vars slice type=[]frontend.Variable
17:15:32 WRN ignoring uninitialized slice slice name=Vars slice type=[]frontend.Variable
17:15:32 WRN ignoring uninitialized slice slice name=Vars slice type=[]frontend.Variable
Test completed for fixture: "/home/w/dev/w/lurk-hs/assets/epoch_change.json"
```

This change is minor and has not been benchmarked.

# Checklist:

### DIFF
--- a/frontend/schema/schema.go
+++ b/frontend/schema/schema.go
@@ -22,6 +22,8 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+
+	"github.com/consensys/gnark/logger"
 )
 
 // Schema represents the structure of a gnark circuit (/ witness)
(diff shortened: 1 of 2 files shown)


ivokub (approved):
Looks great! Thanks for the contribution!

ivokub:
> Looks great! Thanks for the contribution!

And I think that a warning level is good - this may indicate that the slices were indeed uninitialized. We can have false positives in some cases, particularly with non-native arithmetic when in some cases where the resulting variable is zero, then we would have uninitialized slice, but we should fix these false positives in any case.

PR #1307, opened 2024-10-30
fix: sanitize groth16 verification key reading

ivokub:
# Description

See the related security advisory https://github.com/Consensys/gnark/security/advisories/GHSA-cph5-3pgr-c82g (will publish once PR has been reviewed and tested) for full description and POC.

The main issue is that serialized Groth16 verification key includes in a header number of expected Pedersen verification keys which is used for initializing a slice. When modifying the serialized file, we may allocate huge slice, leading to OOM.

In this PR instead of allocating the slice beforehand, we instead try to read all the existing commitment keys in the file and then compare that the number of read commitment keys corresponds to the expected. This prevents using heuristics on the number of commitment keys and bounds the size of the in-memory full verification key to the serialized file size.

## Type of change

# How has this been tested?

Tested against the POC in the advisory. Does not OOM anymore.

# Checklist:

### DIFF
--- a/internal/generator/backend/template/zkpschemes/groth16/groth16.marshal.go.tmpl
+++ b/internal/generator/backend/template/zkpschemes/groth16/groth16.marshal.go.tmpl
@@ -180,9 +180,9 @@ func (vk *VerifyingKey) readFrom(r io.Reader, raw bool) (int64, error) {
 		&nbCommitments,
 	}
 
-	for _, v := range toDecode {
+	for i, v := range toDecode {
 		if err := dec.Decode(v); err != nil {
-			return dec.BytesRead(), err
+			return dec.BytesRead(), fmt.Errorf("read field %d: %w", i, err)
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


ivokub:
@gbotrel - it is done now :) I didn't anticipate making that much bugs :/

internal/generator/backend/template/zkpschemes/groth16/groth16.marshal.go.tmpl
```diff
@@ -178,35 +180,39 @@ func (vk *VerifyingKey) readFrom(r io.Reader, raw bool) (int64, error) {
...
+	for i, v := range toDecode {
 		if err := dec.Decode(v); err != nil {
-			return dec.BytesRead(), err
+			return dec.BytesRead(), fmt.Errorf("read field %d: %w", i, err)
 		}
 	}
 
 	vk.PublicAndCommitmentCommitted = utils.Uint64SliceSliceToIntSliceSlice(publicCommitted)
 	
-	vk.CommitmentKeys = make([]pedersen.VerifyingKey, nbCommitments)
```
gbotrel, line 189: what's the "typical" nbCommitments value? 
for perf, we could add
```
if nbCommitments < N {
vk.CommitmentKeys = make(..., 0, nbCommitments)
}
```

gbotrel (approved):
lgtm, suggestion for perf to avoid many re-allocations in some cases

internal/generator/backend/template/zkpschemes/groth16/groth16.marshal.go.tmpl
ivokub: It is usually very little a la 0 or 1. So I assumed that doing `append` would be OK as the slices are allocated with bigger cap usually anyway.

There was actually another issue - we failed the roundtrip tests when we had `vk.CommitmentKeys == nil` or `vk.CommitmentKeys = []pedersen.VerificationKey`. By default when we initialize the Groth16 key in the roundtrip tests we only do `new(VerificationKey)` and get failing tests. We could have a separate case for `nbCommitments == 0`, but dunno if it is worth it.

PR #1309, opened 2024-10-31
perf: avoid commiting to table and queries in log-derivative argument

ivokub:
# Description

In the log-derivative argument check `F \in T` the prover provides the histogram of the occurrences `M` of the elements `F` in `T`. We then check `\sum_i m_i/(X - t_i) == \sum_j 1/(X - f_j)`.

Previously we always committed to `F, T, M` to obtain the random challenge `r` which we used for checking the equality. But as the left-hand side is always known and the equality doesn't hold for invalid `M` except for negligible probability, then we can avoid committing to `F` and `T`.

However, when the table entries and queries are vectors, then we need to challenge the random coefficients for the random linear combination, so we still commit to the queries then.

Doesn't affect Groth16 as there we compute the commitment as a Pedersen vector commitment. But for PLONK as we need to align the committed values to the commitment column and it saves the number of constraints significantly. Depending on the circuit it may be 10-15%.

See the updated stats

# Checklist:


ivokub:
Not applicable

PR #1312, opened 2024-11-05
perf: BW6 pairing computation using non-native Eval

ivokub:
# Description

This PR refactors the methods in the fields_bw6761 package to use the `Field.Eval` methods introduced in #1299. Depends on it being merged first.

Currently we use simple schoolbook multiplication formulas, but can have better performance with specialized formulas a la Toom-Cook etc. Currently didn't find a good way to give a multivariate representation to them.

## Type of change

# How has this been tested?

- [x] TestSquareVariantsFp6
- [x] TestMulVariantsFp6
- [x] TestFp6Mul023By023Variants
- [x] TestFp6MulBy02345Variants
- [x] TestFp6CyclotomicSquareKarabina12345Variants

- Single Miller loop before: 6500708, now: 3841000
- Single Miller loop fixed G2 before: 5344302, now: 2680076
- FinalExp before: 5245872, now: 3362746
- Full pairing before: 11486969, now: 6947630
- Full pairing fixed G2 before: 10440385, now: 5888826

But keep in mind that I only measured single operations so they include range checks for the initial witness. 

# Checklist:


yelhousni (approved):
Looks great!
I just did some experiments with the new Eval method + other variants of cyclotomic squares. It turns out that Granger-Scott Eval is better for 1 and 2 consecutive squares compared to Karabina. I pushed a commit.

yelhousni:
Actually `squareDirect` is a re-implementation of `mulDirect`. Last commit specialises it to the evaluation of equal operands and saves some more scs.

yelhousni:
Also slight optimization of Karabina decompression Eval which saves an additional 11k scs. Full pairing is now 6,920,110 scs.

yelhousni:
Full pairing is now 6919174 scs.

PR #1324, opened 2024-11-22
fix bn254 solidity template

simplexity-ckcclc:
# Description

Fixes  bn254 solidity template

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/backend/groth16/bn254/solidity.go
+++ b/backend/groth16/bn254/solidity.go
@@ -427,12 +427,8 @@ contract Verifier {
             mstore(f, CONSTANT_X)
             mstore(add(f, 0x20), CONSTANT_Y)
             {{- if gt $numCommitments 0 }}
-            {{- if eq $numCommitments 1 }}
             mstore(g, mload(commitments))
             mstore(add(g, 0x20), mload(add(commitments, 0x20)))
-            {{- else }}
-            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, commitments, {{mul 0x40 $numCommitments}}, g, 0x40))
-            {{- end }}
             success := and(success,  staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
             {{- end }}
             {{- range $i := intRange $numPublic }}
--- a/backend/groth16/bn254/solidity.go
+++ b/backend/groth16/bn254/solidity.go
@@ -427,8 +427,12 @@ contract Verifier {
             mstore(f, CONSTANT_X)
             mstore(add(f, 0x20), CONSTANT_Y)
             {{- if gt $numCommitments 0 }}
+            {{- if eq $numCommitments 1 }}
             mstore(g, mload(commitments))
             mstore(add(g, 0x20), mload(add(commitments, 0x20)))
+            {{- else }}
+            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, commitments, {{mul 0x40 $numCommitments}}, g, 0x40))
+            {{- end }}
             success := and(success,  staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
             {{- end }}
             {{- range $i := intRange $numPublic }}
--- a/backend/groth16/bn254/solidity.go
+++ b/backend/groth16/bn254/solidity.go
@@ -427,7 +427,7 @@ contract Verifier {
             mstore(f, CONSTANT_X)
             mstore(add(f, 0x20), CONSTANT_Y)
             {{- if gt $numCommitments 0 }}
-            {{- if eq $numWitness 1 }}
+            {{- if eq $numCommitments 1 }}
             mstore(g, mload(commitments))
             mstore(add(g, 0x20), mload(add(commitments, 0x20)))
             {{- else }}


ivokub:
Hi @simplexity-ckcclc, thanks for the contribution. Your contribution makes sense when reading diff, but I'm wondering that we haven't encountered the bug before. Do you have a regression test or a very small minified example? Don't worry if not, I'll try to create it otherwise.

simplexity-ckcclc:
Hi, we write this minified circuit for test. When using the original solidity template, in the generated solidity contract below we find in line 401 the first staticcall `PRECOMPILE_ADD`(i.e. 0x06), the input is 64 bytes. But actually it needs two G1 points, which is 128 bytes. It fails in our evm-compatible virtual machine. We think what is doing here is adding the CONSTANT and the commitment (given that there is only one in this case). So we modify the template and it pass. So we wonder is this a bug or we do it wrong?
```go
type testCircuit struct {
	X [2]frontend.Variable
	Y [2]frontend.Variable `gnark:",public"`
}

func (circuit *testCircuit) Define(api frontend.API) error {
	u64, err := uints.New[uints.U64](api)
	if err != nil {
		return err
	}
	u64.ByteAssertEq(u64.ByteValueOf(circuit.X[0]), u64.ByteValueOf(circuit.Y[0]))
	u64.ByteAssertEq(u64.ByteValueOf(circuit.X[1]), u64.ByteValueOf(circuit.Y[1]))
	return nil
}

func TestCircuit(t *testing.T) {
	var c testCircuit
	ccs, _ := frontend.Compile(ecc.BN254.ScalarField(), r1cs.NewBuilder, &c)
	pk, vk, _ := groth16.Setup(ccs)
	f, _ := os.Create("test.sol")
	vk.ExportSolidity(f, solidity.WithHashToFieldFunction(sha256.New()))
	assignment := testCircuit{X: [2]frontend.Variable([]frontend.Variable{uint('a'), uint('b')}),
		Y: [2]frontend.Variable([]frontend.Variable{uint('a'), uint('b')})}
	witness, _ := frontend.NewWitness(&assignment, ecc.BN254.ScalarField())
	publicWitness, _ := witness.Public()
	proof, _ := groth16.Prove(ccs, pk, witness, backend.WithProverHashToFieldFunction(sha256.New()))
	groth16.Verify(proof, vk, publicWitness, backend.WithVerifierHashToFieldFunction(sha256.New()))
}
```
image

ivokub:
Yeah. I confirm it was a problem with the template. In principle our version with error still worked, but it only appeared in your EVM implementation. Previously we had something like this (with one commitment and two public inputs):
```
        assembly ("memory-safe") {
            let f := mload(0x40)
            let g := add(f, 0x40)
            let s
            mstore(f, CONSTANT_X)
            mstore(add(f, 0x20), CONSTANT_Y)
            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, commitments, 64, g, 0x40))
            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
            mstore(g, PUB_0_X)
            mstore(add(g, 0x20), PUB_0_Y)
            s :=  calldataload(input)
            mstore(add(g, 0x40), s)
            success := and(success, lt(s, R))
            success := and(success, staticcall(gas(), PRECOMPILE_MUL, g, 0x60, g, 0x40))
            success := and(success, staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
            mstore(g, PUB_1_X)
            mstore(add(g, 0x20), PUB_1_Y)
            s :=  calldataload(add(input, 32))
            mstore(add(g, 0x40), s)
            success := and(success, lt(s, R))
            success := and(success, staticcall(gas(), PRECOMPILE_MUL, g, 0x60, g, 0x40))
            success := and(success, staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
            mstore(g, PUB_2_X)
            mstore(add(g, 0x20), PUB_2_Y)
            s := mload(publicCommitments)
            mstore(add(g, 0x40), s)
            success := and(success, lt(s, R))
            success := and(success, staticcall(gas(), PRECOMPILE_MUL, g, 0x60, g, 0x40))
            success := and(success, staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))

            x := mload(f)
            y := mload(add(f, 0x20))
```
now we have
```
        assembly ("memory-safe") {
            let f := mload(0x40)
            let g := add(f, 0x40)
            let s
            mstore(f, CONSTANT_X)
            mstore(add(f, 0x20), CONSTANT_Y)
            mstore(g, mload(commitments))
            mstore(add(g, 0x20), mload(add(commitments, 0x20)))
            success := and(success,  staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
            mstore(g, PUB_0_X)
            mstore(add(g, 0x20), PUB_0_Y)
            s :=  calldataload(input)
            mstore(add(g, 0x40), s)
            success := and(success, lt(s, R))
            success := and(success, staticcall(gas(), PRECOMPILE_MUL, g, 0x60, g, 0x40))
            success := and(success, staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
            mstore(g, PUB_1_X)
            mstore(add(g, 0x20), PUB_1_Y)
            s :=  calldataload(add(input, 32))
            mstore(add(g, 0x40), s)
            success := and(success, lt(s, R))
            success := and(success, staticcall(gas(), PRECOMPILE_MUL, g, 0x60, g, 0x40))
            success := and(success, staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))
            mstore(g, PUB_2_X)
            mstore(add(g, 0x20), PUB_2_Y)
            s := mload(publicCommitments)
            mstore(add(g, 0x40), s)
            success := and(success, lt(s, R))
            success := and(success, staticcall(gas(), PRECOMPILE_MUL, g, 0x60, g, 0x40))
            success := and(success, staticcall(gas(), PRECOMPILE_ADD, f, 0x80, f, 0x40))

            x := mload(f)
            y := mload(add(f, 0x20))
```

Here we see that previously
```
success := and(success,  staticcall(gas(), PRECOMPILE_ADD, commitments, 64, g, 0x40))
```
which essentially copies the G1 points in array `commitments` to `g` (as we set input length to be 64 e.g. only a single point!). It is correct, but inefficient as we call BNADD precompile for a simple copy operation. Your fix fixed it and now we just copy
```
            mstore(g, mload(commitments))
            mstore(add(g, 0x20), mload(add(commitments, 0x20)))
```


Now - the issue with your EVM implementation is either:
* you don't handle single point input to BNADD precompile
* or maybe you have some internal edge case which fails due to some other reasons.
In any case, I think you would still have to fix the bug as afaik reading the EVM yellow paper I don't see that single-point BNADD is prohibited. Moreover - when I tested pre-fix version with geth I didn't get a revert.

simplexity-ckcclc:
Thanks, we will read the EVM yellow paper and test again. By the way, our EVM implementation is based on the go-ethereum repo, in the file go-ethereum/core/vm/contracts.go, where the BNADD implementation only supports two g1 points as input. ref: (https://github.com/ethereum/go-ethereum/blob/c7a8bcecbed7b371f0d6ec97ab1a1ae33aa3ca85/core/vm/contracts.go#L503).

simplexity-ckcclc:
Yes, you are right, adding the zero point (the "point at infinity") to any valid point P on the curve yields P itself.

ivokub:
> Yes, you are right, adding the zero point (the "point at infinity") to any valid point P on the curve yields P itself.

Indeed - and in EVM there is a particular way of representing zero point. Usually we use projective coordinates but EVM uses affine and represents `(0, 0)` as zero point. Then when we call BNADD with uninitalized second point, it should resolve to (0,0) and we get the edge case of `BNADD(P, (0,0)) = P`.

And thanks for signing the CLA and the patch! I'll merge it right away.

ivokub (approved):
LGTM! Thanks for the contribution.

PR #1326, opened 2024-11-22
perf: fast path operations for small non-native values

ivokub:
# Description

This PR enforces small non-native values to work on zero or one limb. I also added some fast-path operations when we can deduce the result without needing to perform any operations (mul by zero for example).

Also adds a few more emulation parameters useful for implementing small-fields.

## Type of change

# How has this been tested?

Existing tests succeed

Updated the stats.

# Checklist:


ivokub:
> lgtm. Shouldn't we also have fast path for Div / Inv / Sqrt ?

We should. I'll see how we could be able to implement them.

ivokub:
> > lgtm. Shouldn't we also have fast path for Div / Inv / Sqrt ?
> 
> We should. I'll see how we could be able to implement them.

Added fast paths to Div/Sqrt methods. For inverse we don't support inverse of zero (different from gnark-crypto imo where we take inverse of zero to be zero), so omitted it.

Also added test for checking that the fast paths are complete against short zero (on zero limbs) and large zero.

PR #1333, opened 2024-11-28
perf: normalize the random linear combination in logderivarg

kustosz:
# Description

Sets the first coefficient for random linear combinations in the logderiv argument to 1. This should not affect soundness  (the attacker still cannot abuse any relationship between the coefficients), but it shaves off a constraint from table lookups – in the most basic case we go from 3 constraints down to 2.

Idea by @ivokub, soundness arguments may need to be refined.

## Type of change

- [x] Perf improvement

# How has this been tested?

- [x] run `go test ./...`

Run this before and after:
```go
type TestLookup struct {
	In frontend.Variable
}

func (c *TestLookup) Define(api frontend.API) error {
	table := logderivlookup.New(api)
	for i := range 256 {
		table.Insert(bits.RotateLeft8(uint8(i), 3))
	}
	c0 := c.In
	for range 256 {
		c0 = table.Lookup(c0)[0]
	}
	api.AssertIsEqual(c0, c.In)
	return nil
}

func main() {
	ccs, _ := frontend.Compile(ecc.BN254.ScalarField(), r1cs.NewBuilder, &TestLookup{})
	fmt.Printf("constraints: %d\n", ccs.GetNbConstraints())
}
```

Output goes from 1687 to 1100

# Checklist:

- [X] I have made corresponding changes to the documentation
- [?] `golangci-lint` does not output errors locally - it output a bunch of errors, but none near my changes


ivokub (approved):
LGTM! Thanks for the contribution!

Seems sound and @ThomasPiellard has also confirmed.

PR #1338, opened 2024-12-08
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


ivokub (approved):
Thanks for the contribution! At first I was a bit hesitant, but now when reading discussions about it, seems to make sense as `errors.New` does not require input processing.

I changed the base to `fix/typos-v0.12` which we will merge before next scheduled release.

PR #1348, opened 2024-12-13
refactor: unify the error handling methods in the crypto package that are different from the project style

ChengenH:


RiceChuan:
agree

ivokub:
Thanks for the contribution. I'll keep the current behaviour. `errors.Is` performs recursive comparison, but we're interested for now if the top level error. Otherwise it may happen that we would skip some fields to eagerly.

PR #1349, opened 2024-12-13
feat(bls12-381): pairing using direct Fp12 + non-native `Eval()`

yelhousni:
# Description

Similar to https://github.com/Consensys/gnark/pull/1312 and #1339 but for BLS12-381.

## Type of change

# How has this been tested?

- New tests corresponding to new methods.
- PairingCheck and FinalExpCheck tests against gnark-crypto.

size-2 pairing check circuit, i.e. `e(a,b)e(c,d)==1`,  in a BN254-PLONK:

- old: 4,586,709 scs
- new: 2,686,878 scs
**- diff: 1,899,831 scs**

# Checklist:


std/algebra/emulated/fields_bls12381/e12_pairing.go
```diff
@@ -392,82 +166,59 @@ func (e Ext12) FrobeniusSquareTorus(y *E6) *E6 {
...
-			B0: E2{A0: *res[0], A1: *res[1]},
-			B1: E2{A0: *res[2], A1: *res[3]},
-			B2: E2{A0: *res[4], A1: *res[5]},
-		},
-		C1: E6{
-			B0: E2{A0: *res[6], A1: *res[7]},
-			B1: E2{A0: *res[8], A1: *res[9]},
-			B2: E2{A0: *res[10], A1: *res[11]},
-		},
+	if err != nil {
```
ivokub, line 177: Duplicate `if err != nil` condition.

std/algebra/emulated/fields_bls12381/e12_pairing.go
```diff
@@ -392,82 +166,59 @@ func (e Ext12) FrobeniusSquareTorus(y *E6) *E6 {
 //
 // [On Proving Pairings]: https://eprint.iacr.org/2024/640.pdf
 func (e Ext12) AssertFinalExponentiationIsOne(x *E12) {
```
ivokub, line 168: The method is already in this package, but wouldn't it make more sense to move it to `sw_bls12381` package? Imo rest of pairing related methods are exposed there (FinalExp, ML). I think that because how we have defined the `Pairing` type, then even if we call `Pairing.AssertFinalExponentiationIsOne` actually calls `Pairing.Ext12.AssertFinalExponentiationIsOne`.

But now looking at it, seems that we would have to refactor in all curve packages. Maybe it is not worth it and can keep as is (or do in separate PR).

ivokub (approved):
Looks perfect! Only a small comment about a duplicate code and potential refactor, but I think we can keep the refactor in a separate PR. And imo it is only small thing, dunno if it is worth it.

PR #1363, opened 2024-12-18
optim: avoid div in millerLoopAndFinalExpResult

shramee:
# Description

Super minor optimisation to replace a div with a mul, for exponentiation by `λ' = q^3 - q^2 + q`. We already have the inv so we can just exp the inverse by `q^2` and mul (instead of exp and div).

# How has this been tested?

- [x] Ran all the tests and benches


ivokub:
cc @yelhousni , see also https://github.com/Consensys/gnark/discussions/1347#discussioncomment-11599838

yelhousni (approved):
Thanks for spotting it 👍
Indeed it saves 1585 scs corresponding to the equality check in the hinted `DivUnchecked`.

PR #1365, opened 2024-12-20
perf: `PairingCheck` for BN254, BLS12-381, BLS12-377 and BW6-761

yelhousni:
# Description

This PR saves the squarings previously needed in `AssertFinalExpIsOne` for `PairingCheck` for BN254, BLS12-381, BLS12-377 and BW6-761 elliptic curves. 

See discussion https://github.com/Consensys/gnark/discussions/1347.

## Type of change

# How has this been tested?

`TestPairingCheckTestSolve` tests pass.

The PR saves:
- 371,175 scs for BLS12-381,
- 238,793 scs for BN254,
- 4,272 scs for BLS12-377 and
- 1,195,718 scs for BW6-761.
 
# Checklist:


shramee (approved):
Look great to me!
Thanks for the great work! ❤️

std/algebra/emulated/sw_bls12381/hints.go
```diff
@@ -14,7 +14,135 @@ func init() {
...
 // GetHints returns all hint functions used in the package.
 func GetHints() []solver.Hint {
-	return []solver.Hint{finalExpHint}
+	return []solver.Hint{
+		finalExpHint,
+		pairingCheckHint,
+	}
+}
+
+func pairingCheckHint(nativeMod *big.Int, nativeInputs, nativeOutputs []*big.Int) error {
```
ivokub, line 23: This method is essentially very similar to the finalExpHint? It seems the only difference is that in this method we compute the Milller loop result ourselves. I would refactor the common part into a separate function.

std/algebra/emulated/sw_bn254/hints.go
```diff
@@ -17,10 +17,142 @@ func init() {
 func GetHints() []solver.Hint {
 	return []solver.Hint{
 		millerLoopAndCheckFinalExpHint,
+		pairingCheckHint,
 		finalExpHint,
 	}
 }
 
+func pairingCheckHint(nativeMod *big.Int, nativeInputs, nativeOutputs []*big.Int) error {
```
ivokub, line 25: Similarly, can be refactored with finalExpHint

std/algebra/emulated/sw_bw6761/hints.go
```diff
@@ -0,0 +1,110 @@
...
+
+// GetHints returns all hint functions used in the package.
+func GetHints() []solver.Hint {
+	return []solver.Hint{
+		pairingCheckHint,
+		finalExpHint,
+	}
+}
+
+func pairingCheckHint(nativeMod *big.Int, nativeInputs, nativeOutputs []*big.Int) error {
```
ivokub, line 23: Refactor with finalExpHint

std/algebra/emulated/sw_bls12381/pairing.go
```diff
@@ -100,14 +100,122 @@ func (pr Pairing) Pair(P []*G1Affine, Q []*G2Affine) (*GTEl, error) {
 //
 // This function doesn't check that the inputs are in the correct subgroups.
 func (pr Pairing) PairingCheck(P []*G1Affine, Q []*G2Affine) error {
-	f, err := pr.MillerLoop(P, Q)
+	// check input size match
```
ivokub, line 103: The method seems to be almost fully inlined version of `millerLoopLInes` and `AssertFinalExpIsOne` (with some small differences). I would try to refer to those method directly. For example for `millerLoopLines` we can add boolean that we want to handle first loop iteration separately. Similarly for `AssertFinalExpIsOne`.

I think it is clearer and also more future-proof in case we have some other changes.

std/algebra/native/sw_bls12377/pairing2.go
```diff
@@ -313,21 +313,8 @@ func (p *Pairing) PairingCheck(P []*G1Affine, Q []*G2Affine) error {
...
-	// cyclotomic squaring (e.g. Karabina12345).
-	//
-	// res = res^(p⁶-1)(p²+1)
-	var buf GT
-	buf.Conjugate(p.api, res)
-	buf.DivUnchecked(p.api, buf, res)
-	res.FrobeniusSquare(p.api, buf).Mul(p.api, res, buf)
-
-	res.AssertFinalExponentiationIsOne(p.api)
+	PairingCheck(p.api, inP, inQ)
```
ivokub, line 316: return error

ivokub (changes requested):
Looks good to me. However I think currently there is too much copied implementation. It works, but in the future when we want to target more optimizations/refactors then I'm afraid we could have mismatch between different code paths.

I would try to refactor common parts into separate functions and then use function calls instead. I can have a try at it.

std/algebra/emulated/sw_bw6761/hints.go
yelhousni: this one is different (different `mInv`), but I'll what I can do

yelhousni:
> Looks good to me. However I think currently there is too much copied implementation. It works, but in the future when we want to target more optimizations/refactors then I'm afraid we could have mismatch between different code paths.
> 
> I would try to refactor common parts into separate functions and then use function calls instead. I can have a try at it.

Right, I'll try to refactor some bits.

ivokub (approved):
Thanks! Now imo is a lot more clearer!

PR #1372, opened 2024-12-25
Refac/groth16 mpc

Tabaie:
# Description

https://github.com/Consensys/gnark/pull/1297, but using tools from https://github.com/Consensys/gnark-crypto/pull/589.

Fixes # (issue)

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:


internal/generator/backend/template/zkpschemes/groth16/mpcsetup/utils.go.tmpl
```diff
@@ -52,18 +18,29 @@ func bitReverse[T any](a []T) {
...
-	for i := 1; i < n; i++ {
-		result[i].Mul(&result[i-1], &a)
+// Returns [1, a, a², ..., aᴺ⁻¹ ]
+func powers(a *fr.Element, N int) []fr.Element {
+
+	result := make([]fr.Element, N)
+	if N >= 1 {
+		result[0].SetOne()
+	}
+	if N >= 2 {
```
gbotrel, line 28: this looks useless; or the start bound of the for loop below should change

internal/generator/backend/template/zkpschemes/groth16/mpcsetup/utils.go.tmpl
```diff
@@ -84,70 +66,20 @@ func scaleG2InPlace(A []curve.G2Affine, a []fr.Element) {
...
-// L1 = ∑ rᵢAᵢ, L2 = ∑ rᵢAᵢ₊₁ in G2
-func linearCombinationG2(A []curve.G2Affine) (L1, L2 curve.G2Affine) {
-	nc := runtime.NumCPU()
-	n := len(A)
-	r := make([]fr.Element, n-1)
-	for i := 0; i < n-1; i++ {
-		r[i].SetRandom()
+func areInSubGroupG1(s []curve.G1Affine) bool {
+	for i := range s {
+		if !s[i].IsInSubGroup() {
```
gbotrel, line 71: that's not cheap, we don't have a batch subgroup check yet (@yelhousni wanted to add one) but shouldn't we parallelize since we assume the nominal case is everything is in subgroup ?

internal/utils/test_utils/test_utils.go
```diff
@@ -0,0 +1,54 @@
...
+	"bytes"
+	"github.com/stretchr/testify/require"
+	"io"
+	"log"
+	"testing"
+)
+
+var ConditionalLoggerEnabled bool
+
+func ConditionalLog(v ...any) {
```
gbotrel, line 13: I was going to suggest more idiomatic way to toggle logging levels, but it seems this is not used anyway, remove?

internal/utils/slices.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+// AppendRefs returns append(s, &v[0], &v[1], ...).
+func AppendRefs[T any](s []any, v []T) []any {
```
gbotrel, line 4: is this still needed with golang slices generic stuff ?

gbotrel (changes requested):
lgtm, minor changes maybe

internal/utils/test_utils/test_utils.go
Tabaie: What is that way? I think it's nice to have a tool lying around even if not used on main

internal/utils/slices.go
Tabaie: Just went through the package docs. Nothing there does it compactly.

backend/groth16/bls12-377/mpcsetup/phase2.go
```diff
@@ -129,125 +285,93 @@ func InitPhase2(r1cs *cs.R1CS, srs1 *Phase1) (Phase2, Phase2Evaluations) {
...
-	for i := 0; i < nWires; i++ {
-		var tmp curve.G1Affine
-		tmp.Add(&bA[i], &aB[i])
-		tmp.Add(&tmp, &C[i])
-		if i < public {
-			evals.G1.VKK[i].Set(&tmp)
-		} else {
-			c2.Parameters.G1.L[i-offset].Set(&tmp)
-		}
+	n := len(commons.G1.AlphaTau)
```
ivokub, line 293: We already initialized variable `size := len(commons.G1.AlphaTau)`. Could reuse here.

backend/groth16/bls12-377/mpcsetup/phase2.go
```diff
@@ -129,125 +285,93 @@ func InitPhase2(r1cs *cs.R1CS, srs1 *Phase1) (Phase2, Phase2Evaluations) {
...
+		prev = c[i]
 	}
 
-	return nil
+	pk, vk := prev.Seal(commons, &evals, beaconChallenge)
+	return pk, vk, nil
 }
 
-func (c *Phase2) hash() []byte {
+func (p *Phase2) hash() []byte {
```
ivokub, line 370: similarly as for Phase1, should be able to use provided hash function.

backend/groth16/bn254/mpcsetup/unit_test.go
```diff
@@ -0,0 +1,286 @@
+package mpcsetup
```
ivokub, line 1: File only generated for BN254? I would code-generate just in case.

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,254 @@
...
+	return mpcsetup.SameRatioMany(
+		p.parameters.G1.Tau,
+		p.parameters.G2.Tau,
+		p.parameters.G1.AlphaTau,
+		p.parameters.G1.BetaTau,
+	)
 }
 
-func (phase1 *Phase1) hash() []byte {
+func (p *Phase1) hash() []byte {
```
ivokub, line 233: And imo, the method could also be public and defined:
```func (p *Phase) Challenge(previous []byte) []byte```

File: io/roundtrip.go [hunk 9]
```diff
@@ -21,8 +22,8 @@ func RoundTripCheck(from any, to func() any) error {
 			if err != nil {
 				return err
 			}
-			if !reflect.DeepEqual(from, r) {
-				return errors.New("reconstructed object don't match original (ReadFrom)")
+			if err = equal(from, r); err != nil {
```
ivokub, line 25: I would keep it as is, imo there may be unwanted side-effects of using `Equal` if exists - there are already a few types in gnark which have the Equal method defined and they may be more strict than needed.

It seems like it is used to avoid comparing `Challenge` field in a test. But at least for me it would be clearer to manually empty the Challenge field manually when needed in the test. This current approach may break imo other roundtrip tests.

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,254 @@
...
-	// Check for valid updates using previous parameters
-	if !sameRatio(contribution.Parameters.G1.Tau[1], current.Parameters.G1.Tau[1], tauR, contribution.PublicKeys.Tau.XR) {
-		return errors.New("couldn't verify that [τ]₁ is based on previous contribution")
+	challenge := p.hash()
+	if len(next.Challenge) != 0 && !bytes.Equal(next.Challenge, challenge) {
+		return errors.New("the challenge does not match the previous contribution's hash")
 	}
-	if !sameRatio(contribution.Parameters.G1.AlphaTau[0], current.Parameters.G1.AlphaTau[0], alphaR, contribution.PublicKeys.Alpha.XR) {
-		return errors.New("couldn't verify that [α]₁ is based on previous contribution")
+	next.Challenge = challenge
```
ivokub, line 165: My assumption was that verification wouldn't change the inputs, but it does as we write `next.Challenge`. This imo may have practical side effects that the contributions are not static (a la when hashing serialized proofs). Instead, I would provide the challenge as argument to `Verify` and then `VerifyPhase1` automatically carries the continuation over to next round (with computing it using `newChallenge := Phase1.Hash(challenge)`)

backend/groth16/bls12-377/mpcsetup/phase2.go
```diff
@@ -6,48 +6,200 @@
...
+	// Challenge is the hash of the PREVIOUS contribution
+	Challenge []byte
+}
+
+func (p *Phase2) Verify(next *Phase2, options ...verificationOption) error {
+	challenge := p.hash()
+	if len(next.Challenge) != 0 && !bytes.Equal(next.Challenge, challenge) {
+		return errors.New("the challenge does not match the previous contribution's hash")
+	}
+	next.Challenge = challenge
```
ivokub, line 69: Similarly as for Phase1 imo Phase2.Verify shouldn't change the inputs. See the corresponding comment.

ivokub (changes requested):
Thanks Arya! It looks really nice. Imo I understood it mostly. There are a few comments though, see inline.

A few of the comments are nice to have and you don't need to address now. We could do it in separate PR:
- optional randomness source for contribution
- being able to change hash function for challenge computation for cool tricks a la providing SNARK of ceremony correctness

internal/utils/test_utils/test_utils.go
ivokub: I'd also remove. Imo having only a single toggle for conditional logger is too wide - imo it would be useful if could toggle individually, e.g. based on environment variable:
```
// ConditionalLog logs v... when environment variable envToggle is defined
func ConditionalLog(envToggle string, v ...any)
```

But I would mainly remove it because it is incompatible with `github.com/consensys/gnark/logger` package. I'd keep a single logging approach so that we would have unified message style etc.

backend/groth16/bls12-377/mpcsetup/setup_test.go
```diff
@@ -20,88 +25,92 @@ import (
...
-		nContributionsPhase2 = 3
-		power                = 9
-	)
+// TestAll a full integration test of the MPC setup
+func TestAll(t *testing.T) {
+	testAll(t, 3, 3)
+}
 
+func testAll(t *testing.T, nbContributionsPhase1, nbContributionsPhase2 int) {
 	assert := require.New(t)
```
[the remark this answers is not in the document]
Tabaie: Which methods do you have in mind? Since the setup has to be in a custom way (imitating how an actual multiparty ceremony would go with objects being written to and read from the network) there is only the proof and verification stages to get help with, and those are short enough imo.

backend/groth16/bn254/mpcsetup/unit_test.go
Tabaie: The idea was not to clog the CI with unit tests. The full test is still generated for all curves. Ignoring the unit tests saves 5 seconds per curve on my machine. Perhaps not worth it?

backend/groth16/bls12-377/mpcsetup/phase1.go
Tabaie: It's only changed if it is empty.

backend/groth16/bls12-377/mpcsetup/phase1.go
Tabaie: Would it perhaps be better to remove the `challenge` field entirely, and just feed it as input to the `Contribute` and `Verify` functions?

Tabaie: The downside of this approach is that when sending contributions on the wire the user has to read/write two objects rather than one.

ivokub: But in practice do we need to send challenge on the wire? When the user returns contribution to the sequencer, then it doesn't need to send the challenge, sequencer can compute it on its own (and it actually already does it, assuming that it verifies the contributions).

Now, the question imo is if the user is sent the tip of the ceremony for starting their contribution, then should it verify the full state of the current ceremony. Imo for me it would make sense it should, to ensure that it doesn't contribute to some fork etc. From practical point of view, it is probably too heavy for the users to verify and it could only contribute based on current trusted tip (i.e. ceremony tip and running challenge). And imo then it makes sense that challenge is separate, to indicate its optionality. And in this case imo we can provide it separately to `Contribute` and `Verify` 

But its only my gut feeling, maybe it is okay to keep it in, but add comment that it is "trusted" value and should be verfied in full.

backend/groth16/bn254/mpcsetup/unit_test.go
ivokub: In that case, I would still code generate, but maybe conditionally add `test.Skip()` in case the curve isn't BN254. Then atleast we ensure that everything is syntactically correct and the unit test is covered in code generation.

backend/groth16/bls12-377/mpcsetup/setup_test.go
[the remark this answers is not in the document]
ivokub: There is method below `proveVerifyCircuit`, but indeed it uses fixed proving and verifying key and `test.Assert` doesn't currently have a way to test it like this. But still, `test.Assert` is a very simple wrapper around `require.Assertions`, keeping all the methods what `require.Assertions` has. And imo this unifies the testing style if we use same instance everywhere.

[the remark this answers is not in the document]
Tabaie: Ah so we generally better not use the testify repo directly?

backend/groth16/bls12-377/mpcsetup/utils.go
```diff
@@ -58,18 +25,29 @@ func bitReverse[T any](a []T) {
...
 }
 
-// Returns [aᵢAᵢ, ...] in G1
+// Returns [aᵢAᵢ, ...]∈𝔾₁
+// it assumes len(A) ≤ len(a)
 func scaleG1InPlace(A []curve.G1Affine, a []fr.Element) {
+	/*if a[0].IsOne() {
+		A = A[1:]
+		a = a[1:]
+	}*/
```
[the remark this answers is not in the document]
ivokub: Indeed - but currently the functions `scaleG1InPlace` and `scaleG2InPlace` are used only in one method in `phase1.go` where we know anyway that `len(tauUpdates) == len(c.G1.Tau)` (etc for AlphaTau and BetaTau).

Even more, now having a second look - I think we could just inline `powers` and `scaleG1InPlace` without defining utility methods at all and avoid too early generefication. Then we can avoid allocating the slices `tauUpdates`, `betaUpdates` and `alphaUpdates`.

backend/groth16/bls12-377/mpcsetup/setup_test.go
```diff
@@ -189,12 +189,16 @@ func assignCircuit() frontend.Circuit {
...
-	})()
+var onceCircuit = sync.OnceValue(func() *cs.R1CS {
+	ccs, err := frontend.Compile(curve.ID.ScalarField(), r1cs.NewBuilder, &Circuit{})
+	if err != nil {
+		panic(err)
+	}
+	return ccs.(*cs.R1CS)
+})
+
+func getTestCircuit() *cs.R1CS {
```
ivokub, line 200: And now, we don't even need to have `getTestCircuit` function separately, we can just remove the current `getTestCircuit` function, rename `onceCircuit` variable to `getTestCircuit` and it works exactly the same.

backend/groth16/bls12-377/mpcsetup/utils.go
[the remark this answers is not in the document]
ivokub: And in order to make it parallel, we can use `utils.Parallelize` and then within compute `s = f^start` as the starting point and then at every loop multiply s by f and then do a scalar multiplication.

backend/groth16/bls12-377/mpcsetup/phase1.go
Tabaie: > But in practice do we need to send challenge on the wire?

Yes, any new contributor need to receive the challenge used by the last contribution in order to compute its own challenge (so that the channel reflects the entire protocol transcript)
You are right that the field can be omitted when a contribution is being sent back to the sequencer.

I don't think we should require contributors to verify the whole protocol transcript. That goes against two of the protocol's major design goals:
1. Low entry barrier for contributions: I should be able to contribute relatively easily on a small-ish machine in a short amount of time
2. Indefinite run: The protocol is never truly over. We should always be able to enhance our SRS with more contributions, so the number of contributions is not a constant. So especially in light of (1) the verification effort gets prohibitive.

> but add comment that it is "trusted" value

A trusted value from the contributor's perspective you mean? Because the verifier does reconstruct it, in the same way as you described in the first paragraph:
https://github.com/Consensys/gnark/blob/refac/groth16-mpc/backend/groth16/bn254/mpcsetup/phase1.go#L168

ivokub: Yep, I think from practical point of view it doesn't make sense for the contributor to verify the whole chain, only the previous contribution on top it builds its contribution. And I guess it is fine to keep `Challenge` field inside the structs, makes easier to serialize and transport.

Tabaie:
> Thanks Arya! It looks really nice. Imo I understood it mostly. There are a few comments though, see inline.
> 
> A few of the comments are nice to have and you don't need to address now. We could do it in separate PR:
> 
> * optional randomness source for contribution
> * being able to change hash function for challenge computation for cool tricks a la providing SNARK of ceremony correctness

These are great ideas! I made an issue for them.

io/roundtrip.go
gbotrel: ha, on that one I agree with @ivokub ; I previously had a similar snippet, and had some working round trip tests that actually didn't flag an error because the `Equal` function of an object was not correctly updated in a PR. i.e. `reflect.DeepEqual` is a bit more cumbersome but better for non-regression since it checks everything, including new fields, fields that we precompute, etc, that may not even be checked by a "canonical" `Equal`.

backend/groth16/bls12-377/mpcsetup/setup_test.go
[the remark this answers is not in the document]
gbotrel: I think it's fine to use testify directly here if there is no methods needed from our wrapper; unless we plan to add some extra logic to test or report in a "consistent" manner, but nothing so far

gbotrel (approved):

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
-
-	phase1.Parameters.G2.Beta.Set(&g2)
-
-	// Compute hash of Contribution
-	phase1.Hash = phase1.hash()
-
-	return
+// Contribute contributes randomness to the Phase1 object. This mutates Phase1.
+// p is trusted to be well-formed. The ReadFrom function performs such basic sanity checks.
+func (p *Phase1) Contribute() {
```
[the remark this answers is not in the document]
Tabaie: I agree providing randomness sources would be beneficial for debugging purposes, but are there other uses as well? Perhaps something stronger than whatever the go standard library does for production use?

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
-	scaleG1InPlace(phase1.Parameters.G1.AlphaTau, alphaTau)
-	scaleG1InPlace(phase1.Parameters.G1.BetaTau, betaTau)
-	var betaBI big.Int
-	beta.BigInt(&betaBI)
-	phase1.Parameters.G2.Beta.ScalarMultiplication(&phase1.Parameters.G2.Beta, &betaBI)
-
-	// Compute hash of Contribution
-	phase1.Hash = phase1.hash()
+// setZero instantiates the parameters, and sets all contributions to zero
+func (c *SrsCommons) setZero(N uint64) {
```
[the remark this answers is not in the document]
Tabaie: It's because `tau[i] = tau^i` so when we set `tau=0` we still get `tau[0] = 1`.
In the first case it's needed as well, since only `tau[1:]` are read.

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
-			BetaTau  []curve.G1Affine // {β[τ⁰]₁, β[τ¹]₁, β[τ²]₁, …, β[τⁿ⁻¹]₁}
-		}
-		G2 struct {
-			Tau  []curve.G2Affine // {[τ⁰]₂, [τ¹]₂, [τ²]₂, …, [τⁿ⁻¹]₂}
-			Beta curve.G2Affine   // [β]₂
-		}
-	}
-	PublicKeys struct {
-		Tau, Alpha, Beta PublicKey
+	proofs struct {
```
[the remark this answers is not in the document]
Tabaie: My thinking is that because the challenge sometimes doesn't need to be serialized (for example when a contribution is returning to the coordinator) I'd like to give the contributor the option to nullify it. Considering that the saving is only 32-64 bytes in an object of size ~GB maybe the extra complexity isn't worth it?

Generally I thought it'd be good to default to making things private especially since the parameters are the same object type as the final result of Phase 1, so it's best to make it hard for the user to make the mistake of just grabbing the parameters and calling it a day.

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
+// p is trusted to be well-formed. The ReadFrom function performs such basic sanity checks.
+func (p *Phase1) Contribute() {
+	p.Challenge = p.hash()
+
+	// Generate main value updates
+	var (
+		tauContrib, alphaContrib, betaContrib fr.Element
+	)
+
+	p.proofs.Tau = mpcsetup.UpdateValues(&tauContrib, p.Challenge, 1)
```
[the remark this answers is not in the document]
Tabaie: Great idea. Will implement.

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
+
+// Seal performs the final contribution and outputs the final parameters.
+// No randomization is performed at this step.
+// A verifier should simply re-run this and check
+// that it produces the same values.
+// The inner workings of the random beacon are out of scope.
+// WARNING: Seal modifies p, just as Contribute does.
+// The result will be an INVALID Phase1 object, since no proof of correctness is produced.
+func (p *Phase1) Seal(beaconChallenge []byte) SrsCommons {
+	newContribs := mpcsetup.BeaconContributions(p.hash(), []byte("Groth16 MPC Setup - Phase 1"), beaconChallenge, 3)
```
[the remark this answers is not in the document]
Tabaie: I think that's unnecessary since it's not used anywhere else

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
+		return fmt.Errorf("failed to verify contribution to β: %w", err)
 	}
 
-	// Check hash of the contribution
-	h := contribution.hash()
-	for i := 0; i < len(h); i++ {
-		if h[i] != contribution.Hash[i] {
-			return errors.New("couldn't verify hash of contribution")
-		}
+	if !areInSubGroupG1(next.parameters.G1.Tau[2:]) || !areInSubGroupG1(next.parameters.G1.BetaTau[1:]) || !areInSubGroupG1(next.parameters.G1.AlphaTau[1:]) {
```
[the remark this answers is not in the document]
Tabaie: I don't think the order matters here. If the subgroup check fails the function would still fail. For readability imo it's better to have the more essential logic further up.

backend/groth16/bls12-377/mpcsetup/phase1.go
```diff
@@ -6,187 +6,221 @@
...
+		p.parameters.G1.Tau,
+		p.parameters.G2.Tau,
+		p.parameters.G1.AlphaTau,
+		p.parameters.G1.BetaTau,
+	)
 }
 
-func (phase1 *Phase1) hash() []byte {
+func (p *Phase1) hash() []byte {
 	sha := sha256.New()
```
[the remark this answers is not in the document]
Tabaie: Cool idea indeed! Though I think implementation will be nontrivial - we will have to do some "alignment" work since some of the stuff we're hashing have different lengths. Mark it as a TODO?

backend/groth16/bls12-377/mpcsetup/phase2.go
```diff
@@ -6,48 +6,200 @@
...
+	}
+
+	return nil
+}
+
+// update modifies delta
+func (p *Phase2) update(delta *fr.Element, sigma []fr.Element) {
+	var I big.Int
+
+	scale := func(point any) {
```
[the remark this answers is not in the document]
Tabaie: I think the intention was to avoid repeating the -potentially long- reference to the point being scaled. Changing that now :D

backend/groth16/bls12-377/mpcsetup/marshal.go
```diff
@@ -6,165 +6,245 @@
...
-		&c.PublicKey.SG,
-		&c.PublicKey.SXG,
-		&c.PublicKey.XR,
-		&c.Parameters.G1.Delta,
-		&c.Parameters.G1.L,
-		&c.Parameters.G1.Z,
-		&c.Parameters.G2.Delta,
+func (c *Phase2Evaluations) ReadFrom(reader io.Reader) (int64, error) {
+	var N uint64
+	if err := binary.Read(reader, binary.BigEndian, &N); err != nil {
```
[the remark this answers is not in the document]
Tabaie: Yes I think that was an error. Deleted serialization functions for `Phase2Evals`.

backend/groth16/bls12-377/mpcsetup/phase2.go
```diff
@@ -6,48 +6,172 @@
...
+	}
+
+	// verify proof of knowledge of contributions to the σᵢ
+	// and the correctness of updates to Parameters.G2.Sigma[i] and the Parameters.G1.SigmaCKK[i]
+	for i := range p.Sigmas { // match the first commitment basis elem against the contribution commitment
+		if !areInSubGroupG1(next.Parameters.G1.SigmaCKK[i]) {
+			return errors.New("commitment proving key subgroup check failed")
+		}
+
+		if err := next.Sigmas[i].Verify(challenge, 2+byte(i),
```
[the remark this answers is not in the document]
Tabaie: Since each sigma gets its own DST I'd prefer to start from 1 to have 255 commitments possible (not that I see that many being used).

backend/groth16/bls12-377/mpcsetup/phase1.go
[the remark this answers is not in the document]
ivokub: I see you have already created an issue. I'm okay with postponing it (it is still exploratory imo and currently hardcoding hash is good).

Although I'm not a big believer in TODOs inside the implementation - imo usually they stay in the code indefinitely and then we just remove them (the TODO comments) at some point before audit/release etc. Imo GH issues are more explicit.

backend/groth16/bls12-377/mpcsetup/phase2.go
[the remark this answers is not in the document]
ivokub: Agree that it is good to have separate DST for every commitment. I'm proposing to have a constant for the first commitment and then index from this. A la:
```go
const (
    DST_SIGMA_IDX_0 int = 2
)

// ...
if err := next.Sigmas[i].Verify(challenge, DST_SIGMA_IDX_0+byte(sigma_index) // also rename i to `sigma_index`
```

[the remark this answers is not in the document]
Tabaie: Ah yes I think that's what we have now.

backend/groth16/bls12-377/mpcsetup/setup_test.go
```diff
@@ -157,32 +176,317 @@ func (circuit *Circuit) Define(api frontend.API) error {
...
+	}
 	return nil
 }
 
-func (phase1 *Phase1) clone() Phase1 {
-	r := Phase1{}
-	r.Parameters.G1.Tau = append(r.Parameters.G1.Tau, phase1.Parameters.G1.Tau...)
-	r.Parameters.G1.AlphaTau = append(r.Parameters.G1.AlphaTau, phase1.Parameters.G1.AlphaTau...)
-	r.Parameters.G1.BetaTau = append(r.Parameters.G1.BetaTau, phase1.Parameters.G1.BetaTau...)
+func (p *Phase2) Equal(o *Phase2) bool {
```
ivokub, line 435: Is this method needed anymore when the roundtripchecker doesn't call `Equal` method?

internal/utils/test_utils/test_utils.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"bytes"
+	"github.com/stretchr/testify/require"
+	"io"
+	"testing"
+)
+
+// Range (n, startingPoints...) = [startingPoints[0], startingPoints[0]+1, ..., startingPoints[0]+n-1, startingPoints[1], startingPoints[1]+1, ...,]
+// or [0, 1, ..., n-1] if startingPoints is empty
+func Range(n int, startingPoints ...int) []int {
```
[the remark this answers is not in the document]
Tabaie: It's generally a pretty handy thing. Doesn't hurt to keep it around should it come up again.

backend/groth16/bls12-377/mpcsetup/utils.go
```diff
@@ -90,70 +64,38 @@ func scaleG2InPlace(A []curve.G2Affine, a []fr.Element) {
...
-	res, err := curve.PairingCheck(
-		[]curve.G1Affine{a1, b1},
-		[]curve.G2Affine{na2, b2})
-	if err != nil {
-		panic(err)
-	}
-	return res
+// TODO @Tabaie replace with batch subgroup check, when available
+func areInSubGroupG1(wp *gcUtils.WorkerPool, s []curve.G1Affine, errorReporter func(int)) *sync.WaitGroup {
+	return wp.Submit(len(s), func(start, end int) {
```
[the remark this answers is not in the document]
Tabaie: Given what you pointed out (that `Decode` performs subgroup checks anyway) I guess we can remove the explicit subgroup checks and obviate this whole debate.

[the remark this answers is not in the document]
ivokub: I agree. Imo its the cleanest solution. But leave a comment that in place so that we could remember in the future that the subgroup checks are done elsewhere

backend/groth16/bls12-377/mpcsetup/phase2.go
```diff
@@ -89,26 +236,30 @@ func InitPhase2(r1cs *cs.R1CS, srs1 *Phase1) (Phase2, Phase2Evaluations) {
 	}
 
 	// Prepare Lagrange coefficients of [τ...]₁, [τ...]₂, [ατ...]₁, [βτ...]₁
-	coeffTau1 := lagrangeCoeffsG1(srs.G1.Tau, size)
-	coeffTau2 := lagrangeCoeffsG2(srs.G2.Tau, size)
-	coeffAlphaTau1 := lagrangeCoeffsG1(srs.G1.AlphaTau, size)
-	coeffBetaTau1 := lagrangeCoeffsG1(srs.G1.BetaTau, size)
+	coeffTau1 := lagrangeCoeffsG1(commons.G1.Tau, n)           // [L_{ω⁰}(τ)]₁, [L_{ω¹}(τ)]₁, ... where ω is a primitive sizeᵗʰ root of unity
```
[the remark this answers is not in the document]
Tabaie: From the looks of it they're all used in the same loop. Unless we want to iterate over the constraints multiple times?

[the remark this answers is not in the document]
ivokub: Indeed, this would require looping over the constraints several times. But I think this shouldn't be an issue, I'd recon the time-consuming part is the `accumulateG1` and `accumulateG2` parts, not looping the iterator.

But lets postpone, I'm creating a new issue for it.

ivokub (approved):
Thank you very much @Tabaie for addressing the issues. Sorry for the long back and forth. There are still some performance optimisations and additional features, but I have collected them in https://github.com/Consensys/gnark/issues/1428 for now. See if there is something missing there.

Good to merge from my side.

Tabaie:
Thank you for the detailed review @ivokub!

PR #1375, opened 2024-12-27
feat: enhance API with Printf functionality and examples

### LINKED ISSUE #1239: Improve the use of "api.Println()".
To be precise, I think you can add a func like api.Printf("this is my var %fv. ", circuit.x), and hope it have the same performance with fmt.Printf()

```go
func (circuit *Circuit) Define(api frontend.API) error{
    api.Printf("this is my var %fv. ", circuit.x)
}
go
cmd output: this is my var 2

crStiv:
## Description
This pull request introduces the following changes to the `frontend` package:
- **Added a `Printf` function**: Provides formatted output capabilities for debugging and better insight during computations.
- **Updated `API` and `Compiler` interfaces**: Added new methods to handle variable and coefficient ID retrieval.
- **Examples**: Added a new `PrintfExample` function demonstrating various use cases of the `Printf` functionality.
- **Tests**: Implemented unit tests in `api_test.go` to validate the new `Printf` function, including edge cases and format variations.

Fixes #1239 

## Type of Change

## How Has This Been Tested?
- [x] **Unit Tests**: Validated the correctness of `Printf` in `api_test.go`.
- [x] **Example Execution**: Ensured `PrintfExample` works as expected.

## Benchmarking
- [ ] Benchmark results will be provided in a subsequent update.

## Checklist:
- [x] Performed self-review of code.
- [x] Commented code where necessary.
- [x] Updated relevant documentation.
- [x] Verified unit tests pass locally.
- [x] Ensured changes do not introduce breaking changes.
- [x] Validated `golangci-lint` does not produce errors locally.

---

Allow edits by maintainers for further improvements.

### DIFF
--- a/examples/printf_example.go
+++ b/examples/printf_example.go
@@ -1,0 +1,17 @@
+package examples
+
+import "github.com/consensys/gnark/frontend"
+
+func PrintfExample(api frontend.API) {
+    x := api.Add(10, 20)
+    y := api.Mul(x, 2)
+    
+    // Basic usage
+    api.Printf("x = %v, y = %v\n", x, y)
+    
+    // Different formats for constants and variables
+    api.Printf("dec: %d hex: %x\n", x, x)
+    
+    // Debugging information
+    api.Printf("coeff: %c var: %i\n", x, y)
+}


gbotrel:
if we really want to do this (cc @ivokub ) we should probably not doing on the `API` object but only the `Compiler` one. (i.e. confusing with the `Println` that prints the actual value).

What are the typical scenarios where we want to print the internal frontend representation of a var ?

ivokub:
There is another similar PR open #1373 which imo is a bit more concise and has better tests. But I'll let it settle a bit more to figure out if there is a practical need for it. I'm afraid it could create overhead in the future when going to another arithmetization for example.

PR #1378, opened 2025-01-02
Fixes a typo in the error message phase1.go

sky-coderay:
Fixes a typo in the error message from “couldn't verify valid powers of α(τ) in G₁” to “couldn't verify valid powers of β(τ) in G₁”.


ivokub:
Thanks for the contribution. However we have decided against accepting typo fix contributions from Github users with closed history activity. Furthermore, we are in the process of rewriting MPC part and this PR would be irrelevant.

PR #1380, opened 2025-01-03
fix: typos

futreall:
Thanks for your work !


ivokub:
Thanks for the contribution. However we have decided against accepting contributions from users with locked Github activity history. Furthermore, we are rewriting MPC part and this PR would target code which will be removed.

PR #1381, opened 2025-01-04
Fix mismatched error message for Z verification

Guayaba221:
# Description
in the code, the error message for the Z check mistakenly references L — which can cause confusion during debugging and verification.


ivokub:
Thanks for the contribution. However we do not accept typo fixes from users without public Github history. And we're also reworking MPC packages so this fix would apply to package soon to be removed anyway.

PR #1385, opened 2025-01-07
fix: do not return unused sign

ivokub:
# Description

In the fake GLV method we return the sign, but it is not used in the point. Due to the hint calling abstraction for non-native computation we still range check the variables. It is unoptimal, but in Linea prover we explicitly check if the variables are used in sequential constraints. As we don't, then this leads to compile error.

However, we still use the fifth output in BLS12-377, so no change there.

@yelhousni - can you double check that we haven't previously added unsoundness due to unused variable?

## Type of change

# How has this been tested?

All existing tests succeed.

Updated stats - saves a bit more than 100 constraints for scalar multiplication

# Checklist:


yelhousni (approved):
looks good 👍 I missed that in some refactoring.

PR #1393, opened 2025-01-12
optimize: Remove unnecessary array allocation in polynomial interpola…

VolodymyrBg:
# Description

This PR optimizes the polynomial interpolation implementation by removing unnecessary memory allocation and computation in the computeDeltaAtNaive function.
The optimization removes an unnecessary array allocation and pre-computation loop that stored (at - i) values, instead computing these values directly when needed. This change maintains the same mathematical functionality while reducing memory usage and improving code clarity.
Fixes # (no specific issue number)

## Type of change

# How has this been tested?

The existing test suite should cover this change since it's an optimization that maintains the same mathematical functionality. The function's behavior remains identical - it still computes the same Lagrange basis polynomials, just more efficiently.
The changes affect only the internal implementation details:
- Removed the atMinus array allocation
- Removed the pre-computation loop
- Compute (at - j) values on-demand
- No changes to the mathematical logic or output

The optimization primarily affects memory allocation and loop efficiency. Here are the relevant benchmarks:
- [x] Memory allocation benchmark
- Before: Allocates O(n) extra memory for atMinus array
- After: No extra memory allocation beyond the required deltaAt array
- Impact: Reduced memory footprint for large polynomial interpolations
- [x] Computation efficiency
- Before: Two loops - one for pre-computation, one for actual computation
- After: Single loop with direct computation
- Impact: Eliminates redundant loop iteration while maintaining same mathematical operations
Note: Specific hardware benchmarks can be added after running the benchmarking suite on different platforms.

# Checklist:

### DIFF
--- a/std/polynomial/polynomial.go
+++ b/std/polynomial/polynomial.go
@@ -101,17 +101,13 @@ func negFactorial(n int) int {
 // computeDeltaAtNaive brute forces the computation of the δᵢ(at)
 func computeDeltaAtNaive(api frontend.API, at frontend.Variable, valuesLen int) []frontend.Variable {
 	deltaAt := make([]frontend.Variable, valuesLen)
-	atMinus := make([]frontend.Variable, valuesLen) //TODO: No need for this array and the following loop
-	for i := range atMinus {
-		atMinus[i] = api.Sub(at, i)
-	}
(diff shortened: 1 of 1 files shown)


ivokub (approved):
LGTM!

PR #1401, opened 2025-01-24
fix: several typos

DeVikingMark:


ivokub:
Thanks for the contribution. Unfortunately we don't accept typo fix contributions from GH users with activity history closed.

PR #1403, opened 2025-01-24
optimize: Remove unnecessary array in polynomial interpolation

crStiv:
Optimize computeDeltaAtNaive function by removing the unnecessary atMinus array and its initialization loop. Instead, compute the (at - j) values directly in the main loop. 

This change:
- Reduces memory usage by removing an unnecessary array allocation
- Simplifies the code by removing an initialization loop
- Maintains the same mathematical functionality


ivokub:
Your base branch out of date. Already fixed.

PR #1407, opened 2025-01-31
feat: merkle damgard and poseidon2

Tabaie:
Companion to https://github.com/Consensys/gnark-crypto/pull/623
This PR introduces a generic Merkle Damgard construction and uses it to instantiate Poseidon2 as a full hash function.
(Parameters only for BLS12-377)


std/hash/poseidon2/poseidon2.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// NewMerkleDamgardHasher returns a Poseidon2 hasher using the Merkle-Damgard
+// construction with the default parameters.
+func NewMerkleDamgardHasher(api frontend.API) (hash.FieldHasher, error) {
+	f, err := poseidon2.NewPoseidon2(api)
+	if err != nil {
+		return nil, fmt.Errorf("could not create poseidon2 hasher: %w", err)
+	}
+	return hash.NewMerkleDamgardHasher(api, f, 0), nil
```
ivokub, line 18: Is IV=0 temporary or it indeed doesn't have preimage?

test/quick.go
```diff
@@ -0,0 +1,51 @@
...
+		return errors.New("SingleFunction: unexpected number of output")
+	}
+	for i := range outs {
+		api.AssertIsEqual(outs[i], c.Outs[i])
+	}
+	return nil
+}
+
+// SingleFunction returns a test function that can run a simple circuit consisting of function f, and match its output with outs
+func SingleFunction(curve ecc.ID, f func(frontend.API) []frontend.Variable, outs ...frontend.Variable) func(*testing.T) {
```
ivokub, line 34: I think it is a cool idea, but currently imo it is still not ready. I have several issues with the implementation currently:
- you already have TODO that the registry is not thread safe. If adding a new feature then I would avoid adding implementation debt.
- this utility function prevents self-documenting tests. I'm referring (myself and also if anyone asks in dicusssions or through communication) quite a lot to test circuits to see how to implement. But this abstraction makes the tests a bit more confusing as would have to refer to the `snarkFunctionTestCircuit` definition and the provided function.
- it only uses `IsSolved`, but this tests only against test engine. It is good, but still not perfect and there have been cases where the actual solver/prover etc. differs from the test engine. And it also prevents running Solidity tests easily etc.

I would remove it and define the circuit directly in the `std/hash/poseidon2/poseidon2_test.go` file, it is only a few lines of code.

ivokub (changes requested):
I refactored a bit so that it would align with gnark-crypto (method names, documentation, referring to default parameters from gnark-crypto etc.).

Otherwise, looks good. I'm not so sure about the test utility for single function circuits.

test/quick.go
Tabaie: Addressed points 1 and 3. I think it's useful to remove a lot of bloat from our test files.

std/hash/poseidon2/poseidon2.go
Tabaie: I think it should not be known, but I'll check. For MiMC we also use 0.

test/quick.go
ivokub: In that case, I'll extract it into a separate PR. I would iterate over the design of the method signature and implementation a bit, but don't want to block this PR going forward.

ivokub: See https://github.com/Consensys/gnark/pull/1422 - I'll push changes to current PR to revert the corresponding changes and continue working on the utility there.

std/hash/poseidon2/poseidon2.go
ivokub: I think it should be good, https://www.poseidon-initiative.info/ has set finding preimage of 0 as one of the challenges.

ivokub (approved):
Thanks, looks good now!

PR #1413, opened 2025-02-06
fix: non-native zero IsZero edge case

### LINKED ISSUE #1412: bug: Field not working as expected after recent optimization
Reporting file: 
std/math/emulated/field.go

## Expected Behavior
constant field elements should behave like element variables via various APIs, but now even the simplest API trigger panic.

```
		// f, err := NewField[Secp256k1Fp](api)
		f.IsZero(f.Zero())
```

the above line was inserted to the function `func (c *ConstantCircuit) Define(api frontend.API)` of file `std/math/emulated/field_test.go`

## Actual Behavior
```
gnark/std/math/emulated/field_test.go:75
        	Error:      	Received unexpected error:
        	            	runtime error: index out of range [0] with length 0
        	            	runtime.goPanicIndex
        	            		panic.go:114
        	            	emulated.(*Field[...]).IsZero
        	            		field_assert.go:135
        	            	emulated.(*ConstantCircuit).Define
        	            		field_test.go:63
        	Test:       	TestConstantCircuit
```

## Possible Fix
Could it be due to the recent optimization (https://github.com/Consensys/gnark/pull/1326)?

## Steps to Reproduce
1. Add this line to the function `func (c *ConstantCircuit) Define(api frontend.API)` of file `std/math/emulated/field_test.go`
2.   `f.IsZero(f.Zero())`
3. Run the test `TestConstantCircuit`
4.

## Context
Our projects based on gnark would be impacted. We have paused the upgrade to gnark v0.12. Hope this is fixed asap. Or if you are running out of resource, give us some hints and we'd like to help.

## Your Environment

* gnark version used (e.g. v0.8.1, `HEAD@develop`): v0.12.0
* gnark-crypto version used: 0.15.0
* go version (e.g. 1.20.6): go1.22.6 darwin/arm64
* Operating System and version: macOS Sonoma 14.5

ivokub:
# Description

When the compiler can deduce that non-native element is constant zero, then we represent it on zero limbs. However in `IsZero` method we initialized the accumulator from first limb without range checking. This PR adds fast path for constant zero.

Fixes #1412 

## Type of change

# How has this been tested?

- [x] Improved `IsZeroCircuit` test to include the regression

# Checklist:

### DIFF
--- a/std/math/emulated/element_test.go
+++ b/std/math/emulated/element_test.go
@@ -917,6 +917,9 @@ func (c *IsZeroCircuit[T]) Define(api frontend.API) error {
 	}
 	R := f.Add(&c.X, &c.Y)
 	api.AssertIsEqual(c.Zero, f.IsZero(R))
+
+	isZero := f.IsZero(f.Zero())
+	api.AssertIsEqual(isZero, 1)
 	return nil
 }
 
--- a/std/math/emulated/field_assert.go
+++ b/std/math/emulated/field_assert.go
@@ -103,6 +103,11 @@ func (f *Field[T]) AssertIsInRange(a *Element[T]) {
 // method internally reduces the element and asserts that the value is less than
 // the modulus.
 func (f *Field[T]) IsZero(a *Element[T]) frontend.Variable {
+	// fast path - when the element is on zero limbs, then it is always zero
+	if len(a.Limbs) == 0 {
+		return 1
+	}
+
 	// to avoid using strict reduction (which is expensive as requires binary
 	// assertion that value is less than modulus), we use ordinary reduction but
 	// in this case the result can be either 0 or p (if it is zero).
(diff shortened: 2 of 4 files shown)


yelhousni (approved):
That makes sense. Fast paths for Div and Sqrt are already implemented in field_ops.go (though the comment [here](
https://github.com/consensys/gnark/tree/eb1876387732c2d8fe9ab9601e1fdb5df52f6311/std/math/emulated/field_ops.go#L14) should be `when multiplying by 0`?).

ivokub:
> That makes sense. Fast paths for Div and Sqrt are already implemented in field_ops.go (though the comment [here](https://github.com/consensys/gnark/tree/eb1876387732c2d8fe9ab9601e1fdb5df52f6311/std/math/emulated/field_ops.go#L14) should be `when multiplying by 0`?).

Thanks - changed the comment from "divide by 0" to "divide 0".

PR #1416, opened 2025-02-07
refactor: rename TestCommitUniquenessZerosScs to TestCommitUniquenessZerosSCS

crStiv:
- Fix test function name to use correct capitalization for SCS (Sparse Constraint System) abbreviation
- Move TODO comment to separate line for better readability


ivokub:
Thanks for the PR. However it seems to be a small stylistic change in a test function and doesn't affect readability

PR #1438, opened 2025-03-01
Implement big.Int pool in emulated element.go for improved performance

GarmashAlex:
# Description

This PR implements a big.Int pool in the std/math/emulated/element.go file to improve performance by reducing memory allocations. The implementation uses sync.Pool to reuse big.Int objects instead of creating new ones each time they are needed in the newConstElement function.

Fixes # TODO

## Type of change

# How has this been tested?

- [x] Ran existing tests in the std/math/emulated package to ensure functionality is preserved
= [x] Specifically tested TestIsZeroEdgeCases which passes successfully

- [ ] No specific benchmarks were run, but the implementation follows standard patterns for memory optimization using object pools

# Checklist:

### DIFF
--- a/std/math/emulated/element.go
+++ b/std/math/emulated/element.go
@@ -3,6 +3,7 @@ package emulated
 import (
 	"fmt"
 	"math/big"
+	"sync"
 
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/internal/utils"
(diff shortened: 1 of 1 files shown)


gbotrel:
Thanks for the contribution but this doesn't work as the pool objects are still in use in the frontend.Variable.

PR #1442, opened 2025-03-05
GKR Gate Registry

Tabaie:
Companion to https://github.com/Consensys/gnark-crypto/pull/652

* Implement a GKR Gate registry
* ... which includes automatic degree detection. However this is not perfect, as it always delegates the degree computation to a BN254 instantiation. If for whatever reason a gate polynomial has a leading coefficient multiple of the BN254 scalar field size, the estimated degree will be lower than the actual value. In the future we should probably use the  same field as the frontend api (but that will involve a lot of code generation.)
* Turn gkr.Gate into a struct, rather than an interface.
* Add "feed forward" to Poseidon2 GKR for it to compute a compression function, not just a permutation.


std/gkr/api.go
```diff
@@ -28,18 +28,18 @@ func (api *API) namedGate2PlusIn(gate string, in1, in2 constraint.GkrVariable, i
...
 }
 
 func (api *API) Neg(i1 constraint.GkrVariable) constraint.GkrVariable {
 	return api.NamedGate("neg", i1)
 }
 
-func (api *API) Sub(i1, i2 constraint.GkrVariable, in ...constraint.GkrVariable) constraint.GkrVariable {
-	return api.namedGate2PlusIn("sub", i1, i2, in...)
+func (api *API) Sub(i1, i2 constraint.GkrVariable) constraint.GkrVariable {
+	return api.namedGate2PlusIn("sub2", i1, i2)
```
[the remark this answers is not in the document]
Tabaie: Good catch!

std/gkr/registry.go
```diff
@@ -0,0 +1,148 @@
...
+	degree                    int
+}
+
+// TODO @Tabaie once GKR is moved to gnark, use the same options/settings type for all curves, obviating this
+
+type RegisterGateOption func(*registerGateSettings)
+
+// WithSolvableVar gives the index of a variable whose value can be uniquely determined from that of the other variables along with the gate's output.
+// RegisterGate will return an error if it cannot verify that this claim is correct.
+func WithSolvableVar(solvableVar int) RegisterGateOption {
```
ivokub, line 30: Same comment as for gnark-crypto side -- if you don't have the optimizations implemented, then I wouldn't define the options for configuring the solvedness of a variable. I'm not understanding the context.

std/gkr/registry.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	gatesLock.Lock()
+	defer gatesLock.Unlock()
+	gates[name] = &Gate{Evaluate: f, nbIn: nbIn, degree: s.degree, solvableVar: s.solvableVar}
+	return nil
+}
+
+func GetGate(name GateName) *Gate {
```
ivokub, line 118: Add documentation

std/gkr/registry.go
```diff
@@ -0,0 +1,148 @@
...
+func WithDegree(degree int) RegisterGateOption {
+	return func(settings *registerGateSettings) {
+		settings.degree = degree
+	}
+}
+
+// RegisterGate creates a gate object and stores it in the gates registry
+// name is a human-readable name for the gate
+// f is the polynomial function defining the gate
+// nbIn is the number of inputs to the gate
```
ivokub, line 72: Please follow the godoc style - lines are joined. If you want to define as a list, then use instead:
```go
// RegisterGate creates a gate object and stores it in the gates registry:
//   - name is a human-readable name for the gate
//   - f is the polynomial function defining the gate
//   - nbIn is the number of inputs to the gate
//
// NB! This package generally expects certain properties of the gate to be invariant across all curves.
// In particular the degree is computed and verified over BN254. If the leading coefficient is divided by
// the curve's order, the degree will be computed incorrectly.
```

std/gkr/testing.go
```diff
@@ -29,35 +30,13 @@ import (
 // This method only works under the test engine and should only be called to debug a GKR circuit, as the GKR prover's errors can be obscure.
 func (api *API) SolveInTestEngine(parentApi frontend.API) [][]frontend.Variable {
 	res := make([][]frontend.Variable, len(api.toStore.Circuit))
-	degreeTestedGates := make(map[string]struct{})
+	var degreeTestedGates sync.Map
```
ivokub, line 33: Why `sync.Map`? The variable is local to this method and it doesn't use it in parallel?

std/gkr/gkr.go
```diff
@@ -533,39 +534,8 @@ func DeserializeProof(sorted []*Wire, serializedProof []frontend.Variable) (Proo
...
-
-type AddGate struct{}
-
-func (a AddGate) Evaluate(api frontend.API, v ...frontend.Variable) frontend.Variable {
-	switch len(v) {
-	case 0:
-		return 0
-	case 1:
-		return v[0]
+func panicIfError(err error) {
```
ivokub, line 537: Imo unidiomatic Go. Better to handle explicitly

ivokub (approved):
See the comments. Otherwise LGTM

ivokub:
Also please run gofmt/goimports on the sources. I randomly ran it on `std/gkr/registry.go` and it changed import order. Otherwise there may be unrelated changes in other PR in the future which make reviews more complicated.

std/gkr/internal/bn254_wrapper_api.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr/gkr"
+	"github.com/consensys/gnark/constraint/solver"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/internal/utils"
+	"math/big"
+)
+
+// wrap BN254 scalar field arithmetic in a frontend.API
+// bn254WrapperApi uses *fr.Element as its variable type
+type bn254WrapperApi struct {
```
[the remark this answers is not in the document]
Tabaie: Replaced it with a more limited API

std/gkr/internal/bn254_wrapper_api.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func ToBn254GateFunction(f func(frontend.API, ...frontend.Variable) frontend.Variable) gkr.GateFunction {
+	var wrapper bn254WrapperApi
+
+	return func(x ...fr.Element) fr.Element {
+		if wrapper.err != nil {
+			return fr.Element{}
+		}
+		res := f(&wrapper, utils.Map(x, func(x fr.Element) frontend.Variable {
```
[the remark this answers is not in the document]
Tabaie: The referencing is not essential, but we need to map anyway to turn the `[]fr.Element` to `[]frontend.Variable`

Tabaie:
Merging for the Poseidon2 fix. More PR feedback addressed in future PR.

PR #1445, opened 2025-03-12
optimize FixedLengthSum of hash with minLen

ggq89:
# Description

Use minLen to reduce the times of selects and optimize the FixedLengthSum of hash.

In practice, the length of a piece of data to be hashed is variable, but its minimum value is rarely 0. In this case, if minLen is passed to FixedLengthSum, a lot of constraints can be saved. For example, in the project we are currently working on, one of the data lengths to be hashed ranges from 1680 to 1710. Using the new FixedLengthSum can reduce about 200,000 constraints, as shown in the results of Test_SHA3FixedLengthSum_WithMinLen_VS_Zero:

```
=== RUN   Test_SHA3FixedLengthSum_WithMinLen_VS_Zero
maxLen=1710, minLen=1680, hash=Keccak-512, nbConstraints: 6696856 vs 6555681(withMinLen)
maxLen=1710, minLen=1680, hash=SHA3-256, nbConstraints: 4040628 vs 3796970(withMinLen)
maxLen=1710, minLen=1680, hash=SHA3-384, nbConstraints: 4992172 vs 4800206(withMinLen)
maxLen=1710, minLen=1680, hash=SHA3-512, nbConstraints: 6696856 vs 6555681(withMinLen)
maxLen=1710, minLen=1680, hash=Keccak-256, nbConstraints: 4040628 vs 3796970(withMinLen)
--- PASS: Test_SHA3FixedLengthSum_WithMinLen_VS_Zero (47.33s)
```

## Type of change
- [X] New feature (non-breaking change which adds functionality)

# How has this been tested?

- [X] TestSHA2FixedLengthSum
- [X] TestSHA2FixedLengthSumWithMinLen
- [X] TestSHA3FixedLengthSum
- [X] TestSHA3FixedLengthSumWithMinLen
- [X] Test_SHA3FixedLengthSum_WithMinLen_VS_Zero

# Checklist:

- [X] I have made corresponding changes to the documentation
- [X] `golangci-lint` does not output errors locally

### DIFF
--- a/std/hash/hash.go
+++ b/std/hash/hash.go
@@ -89,6 +89,29 @@ type BinaryFixedLengthHasher interface {
 	FixedLengthSum(length frontend.Variable) []uints.U8
 }
 
+// HasherConfig allows to configure the behavior of the hash constructors. Do
+// not initialize the configuration directly but rather use the [Option]
+// functions which perform correct initializations. This configuration is
+// exported for importing in hash implementations.
+type HasherConfig struct {
+	MinimalLength int
+}
+
+// Option allows configuring the hash functions.
+type Option func(*HasherConfig) error
+
+// WithMinimalLength hints the minimal length of the input to the hash function.
+// This allows to optimize the constraint count when calling
+// [BinaryFixedLengthHasher.FixedLengthSum] as we can avoid selecting between
+// the dummy padding and actual padding. If this option is not provided, then we
+// assume the minimal length is 0.
+func WithMinimalLength(minimalLength int) Option {
+	return func(cfg *HasherConfig) error {
+		cfg.MinimalLength = minimalLength
+		return nil
+	}
+}
+
 // Compressor is a 2-1 one-way function. It takes two inputs and compresses
 // them into one output.
 //
--- a/std/hash/hash.go
+++ b/std/hash/hash.go
@@ -85,7 +85,7 @@ type BinaryHasher interface {
 type BinaryFixedLengthHasher interface {
 	BinaryHasher
 	// FixedLengthSum returns digest of the first length bytes. See the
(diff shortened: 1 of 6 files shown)


std/hash/hash.go
```diff
@@ -85,7 +85,8 @@ type BinaryHasher interface {
 type BinaryFixedLengthHasher interface {
 	BinaryHasher
 	// FixedLengthSum returns digest of the first length bytes.
-	FixedLengthSum(length frontend.Variable) []uints.U8
+	// the minLen is the minimum length of the input.
+	FixedLengthSum(minLen int, length frontend.Variable) []uints.U8
```
weijiguo, line 89: this change creates backward incompatibility

ivokub: Indeed - I'd keep the interfaces as is and change the constructor instead. I'll push the changes

ggq89: ok, updated to backward compatibility interface

ivokub:
Thanks for the contribution! I really like it and it indeed makes the implementation more efficient. I only change the interfaces to keep the current `FixedBinaryHasher` interface intact to ensure that anyone using it downstream wouldn't be broken. Instead, now the bound can be provided as an optional option when initializing the hash function. It is a bit more inconvenient to use when the circuits want to have different bounds, as we need to initialize the hasher for every different lower bound separately, but imo it is better to keep API backwards compatibility. I think sha2/sha3 is already used a lot.

@ggq89, @weijiguo have a look how it seems now. It is good to merge from my side but I'll accept review+merge when you confirm that my changes are good.

ggq89 (commented):
ok, it might be better to move the MinimalLength option to the hash constructor

weijiguo:
Looks good to me. And thank you for stepping in @ivokub

PR #1450, opened 2025-03-24
optimize paddingFixedWidth of sha3 using divmod hint

ggq89:
# Description

In this pr, a divmod function is implemented with hint, and it is used to optimize the paddingFixedWidth of sha3 to reduce the number of selects in the process of padding data and calculating numberOfBlocks.

## Type of change

# How has this been tested?

- [x] TestSHA2FixedLengthSum
- [x] TestSHA3FixedLengthSum

# Checklist:

### DIFF
--- a/std/hash/sha2/sha2.go
+++ b/std/hash/sha2/sha2.go
@@ -100,7 +100,13 @@ func (d *digest) FixedLengthSum(length frontend.Variable) []uints.U8 {
 	// use.
 
 	maxLen := len(d.in)
-	comparator := cmp.NewBoundedComparator(d.api, big.NewInt(int64(maxLen+64+8)), false)
+	maxPaddingCount := 64 - maxLen%64
+	if maxLen%64 >= 56 {
+		maxPaddingCount += 64
+	}
+	maxTotalLen := maxLen + maxPaddingCount
+
+	comparator := cmp.NewBoundedComparator(d.api, big.NewInt(int64(maxTotalLen)), false)
 	// when minimal length is 0 (i.e. not set), then we can skip the check as it holds naturally (all field elements are non-negative)
 	if d.minimalLength > 0 {
 		// we use comparator as [frontend.API] doesn't have a fast path for case API.AssertIsLessOrEqual(constant, variable)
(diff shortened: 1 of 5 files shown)


ivokub:
Thanks for the PR. I'll review it soon.

std/hash/sha2/sha2.go
```diff
@@ -100,7 +100,13 @@ func (d *digest) FixedLengthSum(length frontend.Variable) []uints.U8 {
...
 	maxLen := len(d.in)
-	comparator := cmp.NewBoundedComparator(d.api, big.NewInt(int64(maxLen+64+8)), false)
+	maxPaddingCount := 64 - maxLen%64
+	if maxPaddingCount <= 8 {
+		maxPaddingCount += 64
+	}
+	maxTotalLen := maxLen + maxPaddingCount
+
+	comparator := cmp.NewBoundedComparator(d.api, big.NewInt(int64(maxTotalLen)), false)
 	// when minimal length is 0 (i.e. not set), then we can skip the check as it holds naturally (all field elements are non-negative)
```
ggq89, line 110: the constraint `comparator.AssertIsLessEq(length, maxLen)` should be added

std/hash/sha3/sha3.go
```diff
@@ -75,33 +70,48 @@ func (d *digest) padding() []uints.U8 {
...
 	maxLen := len(d.in)
+	maxPaddingCount := d.rate - maxLen%d.rate
+	maxTotalLen := maxLen + maxPaddingCount
+
+	comparator := cmp.NewBoundedComparator(d.api, big.NewInt(int64(maxTotalLen)), false)
+	// in case the lower bound on the length of input is given, check that the input is long enough
+	if d.minimalLength > 0 {
+		comparator.AssertIsLessEq(d.minimalLength, length)
+	}
+
```
ggq89, line 82: the constraint `comparator.AssertIsLessEq(length, maxLen)` should be added

std/hash/sha2/sha2.go
```diff
@@ -109,7 +115,7 @@ func (d *digest) FixedLengthSum(length frontend.Variable) []uints.U8 {
```
ggq89, line 115: the constraint comparator.AssertIsLessEq(length, maxLen) should be added

ggq89 (commented):
The constraint comparator.AssertIsLessEq(length, maxLen) should be added to prevent the return value when length>maxLen.

std/hash/sha2/sha2.go
ivokub: Is it necessary though? Indeed it can return a wrong value, but if I would see it only breaks the computed padding. Imo this wouldn't allow to construct a hash for which a preimage is known. But for assurance we could indeed add it.

ivokub: In any case, I added the check. Better to err on the safe side if not sure :)

ivokub (approved):

ggq89:
@ivokub It looks fine after you make some changes，and thank you for stepping in!

PR #1457, opened 2025-03-27
fix: poseidon2 compression didn't include feed-forward

ivokub:
# Description

The compression function of Poseidon2 permutation differs between gnark and gnark-crypto. gnark-crypto includes right side feed-forward in the result but gnark not. This made the results differ.

Also added example test for non-default parameters to see that instance generation matches between gnark and gnark-crypto

Strangely - this was tested before but CI didn't fail? I don't see any obvious reasons, @gbotrel maybe you have ideas? I think it is somehow related which curves we test - currently the test was only for BLS12-377 as we don't have default parameters for other curves yet, but I think that for some test configuration it completely discarded testing.

## Type of change

# How has this been tested?

Added new test. Current test works locally. Strangely the test should have had failed in CI before, but didn't?

# Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,9 +7,9 @@ toolchain go1.22.6
 require (
 	github.com/bits-and-blooms/bitset v1.20.0
 	github.com/blang/semver/v4 v4.0.0
-	github.com/consensys/bavard v0.1.29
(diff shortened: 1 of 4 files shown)


Tabaie:
https://github.com/Consensys/gnark/pull/1442 covers this.

ivokub:
Fixed in #1442

PR #1458, opened 2025-03-27
Docs/gkr example

Tabaie:
An example for the GKR package computing point doubling on BLS12-377 G1


Tabaie:
@gbotrel Does this look okay? Without the commitment part the example is not sound.

gbotrel (approved):
lgtm, I guess will have an update after #1490

PR #1462, opened 2025-03-31
feat(wires): implements GetWiresConstraintExact as a draft proposal

AlexandreBelling:
# Description

Fixes # (issue)

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/frontend/cs/scs/builder.go
+++ b/frontend/cs/scs/builder.go
@@ -684,7 +684,7 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
 	}
 }
 
-// GetWireConstraintsNoDuplicate returns the pairs (constraintID, wireLocation) for the
+// GetWireConstraints returns the pairs (constraintID, wireLocation) for the
 // given wires in the compiled constraint system:
 //   - constraintID is the index of the constraint in the constraint system.
 //   - wireLocation is the location of the wire in the constraint (0=xA or 1=xB).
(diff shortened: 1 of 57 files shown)


frontend/cs/scs/builder.go
```diff
@@ -697,8 +697,8 @@ func (builder *builder) ToCanonicalVariable(v frontend.Variable) frontend.Canoni
 //
 // The method only returns a single pair (constraintID, wireLocation) for every
 // unique wire (removing duplicates). The order of the returned pairs is not the
-// same as for the given arguments.
-func (builder *builder) GetWireConstraints(wires []frontend.Variable, addMissing bool) ([][2]int, error) {
+// same as for the given arguments. It is however, deterministic order.
+func (builder *builder) GetWireConstraintsN(wires []frontend.Variable, addMissing bool) ([][2]int, error) {
```
ivokub, line 701: Should keep the method name? Also in documentation.

ivokub (approved):
Looks good to me. Indeed I think we could optimize that we wouldn't add constraints for every constant (if they are duplicate), but I guess it shouldn't happen too much?

Only thing I would change is the name of the existing method - the documentation says `GetWireConstraintsNoDuplicate` but method name is `GetWireConstraintsN`. But I would keep it as is, because we use it using implicit interfaces and then its easier to keep existing implementation working.

AlexandreBelling:
Fair enough. I have applied your suggestion. For the constant, it seems easy enough to use a map to read ten times the same constants. For MiMC, the constant is essentially the "0" that we use to initialize the state of a MiMC hasher.

AlexandreBelling:
Lemme test this on my end before merging. Will merge after. Happy to get a reapproval :)

frontend/cs/scs/builder.go
```diff
@@ -750,3 +750,130 @@ func (builder *builder) GetWireConstraints(wires []frontend.Variable, addMissing
...
+
+// GetWiresConstraintExact works as [GetWireConstraints], but returns an
+// exact wire for each constraint. That is if the caller passes the same wire
+// several times at different positions in [vars], it will not deduplicate
+// unlike [GetWireConstraints]. The function has also a different way to deal
+// with constants and missing wires. If the same variabes is passed, then the
+// same wire ID is returned. The function returns the first occurrence of the
+// wire in the constraint system, by order of the constraints.
+//
+//   - If a variable is a constant. It will introduct an adhoc term and it will
```
ivokub, line 762: Not valid anymore.

ivokub (approved):
Yep - now it seems better. Now only the method documentation doesn't match the implementation.

PR #1466, opened 2025-04-03
GKR wrapper APIs

Tabaie:
Adds wrapper APIs for all curves. Thus, a gate's degree can be computed and verified over any curve the user demands.

`utils.FromInterface` returns an error now. To have a version that panics on error, `utils.ForceFromInterface` can be used.


gbotrel:
@ivokub didn't you mention you were looking into a "convert" package or something like that? may be a good home for the `utils.ForceFromInterface` (don't like too much the naming, we sometime use "Must", sometime "Force" we should be consistent but that's me nitpicking ^^)

ivokub:
> @ivokub didn't you mention you were looking into a "convert" package or something like that? may be a good home for the `utils.ForceFromInterface` (don't like too much the naming, we sometime use "Must", sometime "Force" we should be consistent but that's me nitpicking ^^)

The convert package I was thinking is a gadget which would allow:
- convert between different emulated fields in-circuit (for example in ECDSA we need to cast from p256 base field to scalar field)
- convert from native field to emulated field (useful for proof recursion, because otherwise when we have emulated element in emulated element recursion, then we're really wasteful as first we need 4*64 limbs to represent non-native element and later in recursion 4*4*64, as we represent every limb as a non-native element etc.)
- pack many []uints.U8 into a single frontend.Variable
- maybe some more usecases, would have to have a look where are the different places we use already. One place is for example in recursive PLONK where for FS challenges we use binary decomposition to be able to use native mimc.

So I guess not applicable in this PR?

gbotrel:
Indeed 😄

Tabaie:
Made obsolete by https://github.com/Consensys/gnark/pull/1490

PR #1467, opened 2025-04-04
Migrate GKR from gnark-crypto

Tabaie:
Companion to https://github.com/Consensys/gnark-crypto/pull/670


Tabaie:
> shouldn't internal/* be in std/gkr/internal/* ?
Tried that first, but it is also accessed from the `constraint` package.

Tabaie:
> * same thing for test_vector_utils (don't really like the package name / place where it lives) --> if it's meant to be use in gkr test internally, then just being defined in the same place under _test.go is fine, if it's meant to be use across the gkr/internal/native prover and the grk/ circuit prover, then I would just defined it in the internal package along side the gkr prover

Addressed in https://github.com/Consensys/gnark/pull/1476

gbotrel (approved):

PR #1469, opened 2025-04-05
refactor: incremental instance addition for GKR API

### LINKED ISSUE #1424: Enable the GKR API to add instances
Currently all the instances of a GKR sub-circuit have to be defined at once. This can be cumbersome since it gives the user access to the output variables only after they are all defined. For example, if the user is defining hash permutations, the results of which are needed right away, they would have to independently define a hint that computes the output, and later equate that with the `Export`ed values from the gkr API.

Instead it is desirable for the user to be able to define the whole circuit (by allowing the `Import` function to take no arguments, or better yet renaming it to `NewVariable`). And then having a function called `NewInstance/NewInstances` that takes assignments or dependencies for the input values and returns assignments for output variables.

The complexity in this will be that the GKR `Solve` hint would have to be called multiple times (for every invocation of `Export` or `NewInstance`).

DeVikingMark:
# Description

Added incremental instance addition to GKR API to enable users to add variables and instances on-demand instead of having to define all instances at once through the `Import()` method.

Fixes #1424

## Type of change

# How has this been tested?

- [x] Created a new test `TestNewVariableAndInstance` that verifies the functionality of `NewVariable()` and `NewInstance()` methods
- [x] Modified existing tests to ensure compatibility with the changes
- [x] Added additional validation to prevent potential errors and edge cases

# Checklist:

### DIFF
--- a/std/gkr/README.md
+++ b/std/gkr/README.md
@@ -1,0 +1,141 @@
+# Generic Knowledge of Computation (GKR) API
+
+This package implements the GKR protocol in gnark. It allows proving the correct execution of a circuit on multiple instances at once.
+
+## Overview
+
+GKR is a protocol that allows efficient proving and verification of computations. The GKR API in gnark allows defining a circuit once and then proving its correct execution on multiple inputs.
+
+## Usage
+
+### Basic Usage
+
+```go
+// Initialize GKR API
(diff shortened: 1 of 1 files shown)


ivokub:
Duplicate of #1504

PR #1040, opened 2024-02-07
feat: added HashToG2 and BLS G2 signature verification circuit for BLS12-381

weijiguo:
# Description

This PR implemented circuit for `HashToG2` and BLS G2 signature verification for the BLS12-381 curve. The implementation is based on affine coordinations. 

Along with the said functionalities, it also added `G2.addUnified(p, q)` function which can handle the case that `p == q`. And as an optimization, it also adopted a new hint to calculate the `sqrtRatio` function with the gnark-crypto library to save constraints. Therefore this PR depends on [an update to gnark-crypto](https://github.com/Consensys/gnark-crypto/pull/481)

Fixes # (issue)
https://github.com/Consensys/gnark/issues/648

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

Added unit tests to cover:

- [x] G2.addUnified for BLS12-381
- [x] HashToG2 for BLS12-381

# How has this been benchmarked?

Added test cases for HashToG2. Results: 2761896 constraints with SCS and 779198 with R1CS for simple message ("abcd").

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1473, opened 2025-04-07
feat: implement map to G2 on BLS12-381

ivokub:
# Description

This PR takes #1447 and #1040 and https://github.com/Consensys/gnark-crypto/pull/674, implementing Map to G2 over BLS12-381.

It is still work in progress, i'm refactoring to make it compatible with the non-native group algebra we have elsewhere. I'll also generify mapping bytes to non-native elements more efficiently (instead of doing it byte-by-byte).

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:


std/signature/bls/bls_g1.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// public key cannot be infinity. Thus either coordinate has to be non-zero.
+	xtest := fp.IsZero(&pk.X)
+	ytest := fp.IsZero(&pk.Y)
+	pubTest := api.And(xtest, ytest)
+	api.AssertIsEqual(pubTest, 0)
+
+	// prime order subgroup checks
+	pairing.AssertIsOnG1((*sw_bls12381.G1Affine)(&pk))
+	pairing.AssertIsOnG2((*sw_bls12381.G2Affine)(sig))
```
yelhousni: I think `PairingCheck()` already checks for G2 membership in `computeLines()` method 
https://github.com/consensys/gnark/tree/44ab4051ac0868f05ef1dfe6b2fd4119ec5eb588/std/algebra/emulated/sw_bls12381/precomputations.go#L50
So this check here is redundant

ivokub: Thanks! I removed the redundant check.

PR #1479, opened 2025-04-16
feat: emulated AssertIsDifferent

ivokub:
# Description

Even though we were able to assert emulated element difference previously quite easily, then add a dedicated method so that in case we have some optimizations in the future it could be applied automatically.

## Type of change

# How has this been tested?

- [x] TestAssertIsDifferent

# Checklist:


yelhousni (approved):
LGTM!

PR #1490, opened 2025-05-05
Refactor/gkr/unigate

Tabaie:
This PR removes the need for independently defining GKR gates for each curve. The syntax for defining a gate now can be as simple as `gkrgates.Register(gateFunction, nbInput)` with options available for names, if the function is anonymous. A gate function has the signature `func (gkr.GateAPI, in ... frontend.Variable) frontend.Variable`. Any `frontend.API` is a `gkr.GateAPI` but not vice versa.

Behind the scenes, the unification has enabled removal of a lot of boilerplate code, aided by the replacement of pointers to specific types (Wires) with their index.

Lastly, a `Println` function has been provided to aid circuit debugging. Its only difference with the `Println` convention is that its first argument determines the instance index of the variables to be printed.

I'm still open to a reorganization/consolidation of packages that would reduce their number.

* `std/gkr`: Basic definitions of types exposed to the uses like `gkr.Variable`, `gkr.GateFunction`, as well as common gate names like `gkr.Identity`.
The Verifier SNARK gadget. Circuit and Wire types.
* `std/gkrapi`: Containing the API type, instantiated by `gkrapi.New`
* `solver/gkrgates`: The Gate registry.

* `internal/gkr/{curve}`: Per curve implementation of prover and verifier, along with hints for SNARK solver integration.
* `internal/gkr/gkrinfo`: Serializable information capable of being saved in a SNARK circuit CS object.
* `internal/gkr/gkrtypes`: Definition of the Gate, Wire, and Circuit types. Also includes the topological sorting function.
* `internal/gkr`: The SNARK verifier gadget, along with the Circuit, Wire, and WireAssignment types.

### Dependency Diagram
**Note:** The diagram is not complete. A few transitively implied dependencies have been removed for readability. Removed dependencies are available as commented lines in this markdown file.
Also included in the diagram are non-GKR packages `frontend`, and `constraint/solver`.

```mermaid

graph TD
    std/gkr --> frontend
    %% gkrapi --> std/gkr
    gkrapi --> gkrinfo
    %% gkrapi --> solver
    gkrapi --> gkrgates
    gkrapi --> internal/gkr
    gkrgates --> curve
    %% gkrgates --> gkrtypes
    %% gkrgates --> std/gkr
    %% curve --> std/gkr
    curve --> gkrtypes
    %% curve --> frontend
    curve --> solver
    %% gkrtypes --> frontend
    gkrtypes --> std/gkr
    gkrinfo --> solver
    internal/gkr --> gkrtypes
    internal/gkr --> gkrinfo

    frontend --> solver
    
```

Depends on https://github.com/Consensys/gnark-crypto/pull/687


backend/backend.go
```diff
@@ -206,3 +207,17 @@ func WithVerifierKZGFoldingHashFunction(hFunc hash.Hash) VerifierOption {
 		return nil
 	}
 }
+
+// SupportedCurves returns a list of curves that gnark backends
+// (GROTH16 and PLONK) have been implemented for.
+func SupportedCurves() []ecc.ID {
```
ivokub, line 213: This is already available in root package method `Curves()`. Better not to duplicate.

std/gkr/types.go
```diff
@@ -0,0 +1,51 @@
...
+type Variable int // Variable represents a value in a GKR circuit.
+
+// GateAPI is a limited version of frontend.API,
+// allowing ring arithmetic operations
+type GateAPI interface {
+	// ---------------------------------------------------------------------------------------------
+	// Arithmetic
+
+	// Add returns res = i1+i2+...in
+	Add(i1, i2 frontend.Variable, in ...frontend.Variable) frontend.Variable
```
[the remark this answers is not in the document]
ivokub: And also imo another issue is that it really prevents us changing `frontend.Variable` definition. If for some reason we would like it to be anything else, then it would break the GKR API and it would be more difficult to implement the changes.

ivokub (changes requested):
I started reviewing - I'll continue on Monday but there are already a few points to address. A few small issues, but imo the `gkr.API` interface right now takes as input `frontend.Variable` as variables and for me it is very counter-intuitive.

As a practical example - at first I also tried implementing non-native API similarly (using frontend.Variable as it was `interface`), but it made writing circuits quite error-prone and the non-native implementation was also a lot more difficult (i.e. we had to handle constants, non-native, native etc in all combinations). Having separate parametric types really forces safety through types and simplifies circuits and implementation.

If possible, then I'd really recommend considering it GKR API would work with dedicated `Variable` instead.

std/gkr/types.go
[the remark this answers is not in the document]
Tabaie: The idea is for `frontend.API` to be a `gkr.GateAPI`. But that's not strictly necessary and I can add a wrapper for frontend.API as with curve arithmetic.
The problem with the `gkr.Variable` approach would then be introducing yet another `any` type which would imo reduce type safety rather than improve it.
Right now `gkr.Variable` is restricted to integer values, denoting a wire in the circuit. But if we want it to be able to take `int`, `fr.Element`, `big.Int` etc it would also have to become an alias for `interface{}` just like `frontend.Variable`, so may as well use the latter.
Are there any suggestions to change the definition of `frontend.Variable`? To me it's any representation of a numerical value, which is fitting to use here.

constraint/solver/gkrgates/registry.go
```diff
@@ -0,0 +1,272 @@
...
+type registerSettings struct {
+	solvableVar               int
+	noSolvableVarVerification bool
+	noDegreeVerification      bool
+	degree                    int
+	name                      gkr.GateName
+	curves                    []ecc.ID
+}
+
+type RegisterOption func(*registerSettings)
```
ivokub, line 39: Add documentation.

constraint/solver/gkrgates/registry.go
```diff
@@ -0,0 +1,272 @@
...
+// GetDefaultGateName provides a standardized name for a gate function, depending on its package and name.
+// NB: For anonymous functions, the name is the same no matter the implicit arguments provided.
+func GetDefaultGateName(fn gkr.GateFunction) gkr.GateName {
+	// copied from solver.GetHintName
+	fnptr := reflect.ValueOf(fn).Pointer()
+	name := runtime.FuncForPC(fnptr).Name()
+	return gkr.GateName(newToOldStyle(name))
+}
+
+func newToOldStyle(name string) string {
```
ivokub, line 215: We don't need it here I think. It was only for compatibility when we upgraded from Go 1.18 imo. But because the APIs have changed anyway, then we don't need to consider to be backwards compatible now as now we force Go 1.23 anyway.

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/consensys/bavard v0.1.31-0.20250406004941-2db259e4b582
 	github.com/consensys/compress v0.2.5
-	github.com/consensys/gnark-crypto v0.17.1-0.20250502112255-56600883e0e9
+	github.com/consensys/gnark-crypto v0.17.1-0.20250515203007-25d707ef7020
```
ivokub, line 12: This commit is out of gnark-crypto master, i.e. do not merge before the dependency on gnark-crypto side is merged!

std/polynomial/polynomial_test.go
```diff
@@ -31,82 +29,26 @@ func testEvalPoly(t *testing.T, p []int64, at int64, evaluation int64) {
...
 		Evaluation: evaluation,
 	}
 
-	assert.CheckCircuit(&evalPolyCircuit{P: make(Polynomial, len(p))}, test.WithValidAssignment(&witness))
+	assert.CheckCircuit(&evalPolyCircuit{P: make(polynomial.Polynomial, len(p))}, test.WithValidAssignment(&witness))
 }
 
 func TestEvalPoly(t *testing.T) {
 	testEvalPoly(t, []int64{1, 2, 3, 4}, 5, 586)
 }
```
ivokub, line 43: WHy the tests removed?

std/gkrapi/example_test.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/consensys/gnark/constraint/solver/gkrgates"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/std/gkr"
+	"github.com/consensys/gnark/std/gkrapi"
+	stdHash "github.com/consensys/gnark/std/hash"
+	"github.com/consensys/gnark/std/hash/mimc"
+	"github.com/consensys/gnark/test"
+)
+
+func Example() {
```
ivokub, line 18: Good to have an example. But I would check and remove the TODOs within.

std/gkrapi/testing.go
```diff
@@ -0,0 +1,121 @@
...
+func WithHashName(name string) SolveInTestEngineOption {
+	return func(s *solveInTestEngineSettings) {
+		s.hashName = name
+	}
+}
+
+// SolveInTestEngine solves the defined circuit directly inside the SNARK circuit. This means that the method does not compute the GKR proof of the circuit and does not embed the GKR proof verifier inside a SNARK.
+// The output is the values of all variables, across all instances; i.e. indexed variable-first, instance-second.
+// This method only works under the test engine and should only be called to debug a GKR circuit, as the GKR prover's errors can be obscure.
+
```
ivokub, line 30: Remove empty line to associate the documentation with method

ivokub (changes requested):
Seems ok, but assuming it is mostly refactoring from gnark-crypto side, I didn't super thoroughly go through the implementation itself.

I'll also check the PR for gnark-crypto so that you could update the dependency.

std/gkrapi/api.go
```diff
@@ -0,0 +1,83 @@
+package gkrapi
```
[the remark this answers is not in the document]
Tabaie: I did a rough diagram of dependencies between packages. From there it seems only `gkrtypes` and `gkrinfo` can be merged. It made some semantic sense to separate them but I'm open to fusing them together.

std/polynomial/polynomial_test.go
Tabaie: I had to change the name of the package to break an import cycle caused by the new hash registry.
The removed tests belong to private functions already covered by tests for their "parent"s.

internal/gkr/gkrtypes/types.go
```diff
@@ -0,0 +1,425 @@
...
+
+	for i := range c {
+		sorted[i] = &c[data.leastReady]
+		data.markDone(data.leastReady)
+	}
+
+	return sorted
+}
+
+type errorString string
```
ivokub, line 382: Also, why this? You can just do
```
var ErrZeroFunction = errors.New("detected a zero function")
```
It is very ideomatic Go and whole standard library uses that approach instead: https://pkg.go.dev/io/fs#pkg-variables

internal/gkr/gkrtypes/types.go
```diff
@@ -0,0 +1,422 @@
...
+
+	for i := range c {
+		sorted[i] = &c[data.leastReady]
+		data.markDone(data.leastReady)
+	}
+
+	return sorted
+}
+
+func ErrZeroFunction() error {
```
ivokub, line 383: Why a function which returns errors? Just use
```
var ErrZeroFunction = errors.New("detected a zero function")
```

Otherwise I don't see the point of defining the error at all. Usually you use fixed errors if you want to test if a returned error value is something special using `errors.Is`, but if you create a new error every time then you lose this ability. See example https://go.dev/play/p/8xWA2ZMk14a

PR #1492, opened 2025-05-13
feat: make automatic element initialization aware of the current field

ivokub:
# Description

We previously had the `GnarkInitHook` interface which is called whenever the circuit definition is parsed. This happens for example when the circuit is compiled or when the assignment is parsed for witness construction. This allowed for example very conveniently define the non-native elements in the circuit definition without needing to add explicit initializers.

However, when targeting smallfields as the native field, then it was incompatible as the default limb widths was larger than the small fields fit. This PR makes the initialization hook field-aware.

Now - this created a few problems in the field emulation as the witness initialization method `emulated.ValueOf` is synchronous at calling time. And its argument list doesn't take the native field as an argument. I resisted changing the function interface as there are many downstream users whose usage would be broken and imo it didn't seem like a good API.

Instead - now the witness assignment for non-native element is asynchronous. This means that when calling `emulated.ValueOf` then we don't perform the limb decomposition directly but rather store the witness value as `*big.Int` and only when `Element.Initialize` (previous `GnarkInitHook`) is called, we perform the witness assignment in limbs.

Another issue which it created is that there was one usage of non-native arithmetic which directly relied on the limb decomposition (GLV scalar mul) where the limbs are constant (i.e. the non-native element is given by the tag `gnark:"-"`) in which case the `Initialize` method is not called, where I also patched the usage.

And the final issue is that when `emulated.ValueOf` is called inside the circuit (for example to define some constants used) then it would perform full limb decomposition (i.e. not take the minimal number of limbs required) as we don't really know if we're in the "in-circuit" or "witness-assignment" setting. Using `emulated.ValueOf` inside circuit was anyway previously not advised (and instead it was recommended to use `Field.NewElement`), but there were still many uses even in the standard library. This PR also changes all such calls inside the standard library, resulting in reduction in circuit sizes.

This PR is a necessary requirement for allowing to implement non-native arithmetic over small fields. Upcoming PRs will add that support also. See #1314 for larger context (new PRs will be cherry-picked from commits from that branch).

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -- in some cases, but was documented before.

# How has this been tested?

- amended witness initialization tests with small field support
- existing tests pass

Not benchmarked

# Checklist:


backend/witness/witness.go
```diff
@@ -356,7 +356,7 @@ func (w *witness) FromJSON(s *schema.Schema, data []byte) error {
 	// collect all secret values; if any are missing, we just deal with the public part.
 	secretValues := make([]any, 0, s.NbSecret)
 	publicOnly := false
-	if _, err := schema.Walk(instance, ptrTyp, func(leaf schema.LeafInfo, tValue reflect.Value) error {
+	if _, err := schema.Walk(nil, instance, ptrTyp, func(leaf schema.LeafInfo, tValue reflect.Value) error {
```
gbotrel, line 359: well, I guess the `ToJSON` / `FromJSON` methods are not really used, but witness could do a type switch in a method to figure out the base field or something, and provide it here to avoid a nil pointer with subtle side effects ? 
Or maybe just document in `Walk` that `nil` is a OK value and is used only in `InitHook`

frontend/schema/leaf.go
```diff
@@ -20,8 +23,20 @@ type LeafCount struct {
...
+// InitHooker is an object which knows how to initialize itself when parsed at
+// compile time.
+//
+// This allows to define new primitive circuit variable types which may require
+// allocations and by using this interface the circuit user doesn't need to
+// explicitly initialize these types themselves.
+//
+// The Initialize method can be called multiple times during different parsing
+// and compilation steps, so the implementation should be idempotent.
+type InitHooker interface {
```
gbotrel, line 35: mmhh the name ***Hooker may have some weird connotations :) 

What about `Initializable` ? ChatGPT says:

"•	Simple and idiomatic Go name for types that can initialize themselves.
	•	Common naming pattern in Go: e.g., io.Reader, json.Marshaler, sql.Scanner.
    "

internal/stats/latest_stats.csv
```diff
@@ -125,20 +125,20 @@ math/bits.ToTernary/unconstrained,bls24_315,plonk,160,319
...
-math/emulated/secp256k1_64,bw6_761,groth16,1070,1950
-math/emulated/secp256k1_64,bw6_633,groth16,1070,1950
-math/emulated/secp256k1_64,bn254,plonk,4497,4388
-math/emulated/secp256k1_64,bls12_377,plonk,4497,4388
-math/emulated/secp256k1_64,bls12_381,plonk,4497,4388
-math/emulated/secp256k1_64,bls24_315,plonk,4497,4388
-math/emulated/secp256k1_64,bls24_317,plonk,4497,4388
-math/emulated/secp256k1_64,bw6_761,plonk,4497,4388
-math/emulated/secp256k1_64,bw6_633,plonk,4497,4388
+math/emulated/secp256k1_64,bn254,groth16,1037,1890
```
gbotrel, line 128: nice :)

internal/stats/snippet.go
```diff
@@ -91,8 +91,8 @@ func initSnippets() {
 
 		x13 := secp256k1.Mul(newElement(), newElement())
 		x13 = secp256k1.Mul(x13, newElement())
-		five := emulated.ValueOf[emulated.Secp256k1Fp](5)
-		fx2 := secp256k1.Mul(&five, newElement())
+		five := secp256k1.NewElement(5)
```
gbotrel, line 94: also nice

std/algebra/emulated/sw_bls12381/map_to_g1.go
```diff
@@ -6,13 +6,12 @@ import (
...
 	"github.com/consensys/gnark-crypto/ecc/bls12-381/hash_to_curve"
 	"github.com/consensys/gnark/frontend"
-	"github.com/consensys/gnark/std/math/emulated"
 )
 
 func (g1 *G1) evalFixedPolynomial(monic bool, coefficients []fp.Element, x *baseEl) *baseEl {
 	emuCoefficients := make([]*baseEl, len(coefficients))
 	for i := range coefficients {
-		emulatedCoefficient := emulated.ValueOf[emulated.BLS12381Fp](coefficients[i])
+		emulatedCoefficient := *g1.curveF.NewElement(coefficients[i])
```
gbotrel, line 14: does it make sense here to do "*value" and the next line to do "&result" ?

std/algebra/emulated/sw_emulated/hints.go
```diff
@@ -95,190 +94,93 @@ func decomposeScalarG1Signs(mod *big.Int, inputs []*big.Int, outputs []*big.Int)
 
 // TODO @yelhousni: generalize for any supported curve as it currently supports only:
 // BN254, BLS12-381, BW6-761 and Secp256k1, P256, P384 and STARK curve.
-func scalarMulHint(_ *big.Int, inputs []*big.Int, outputs []*big.Int) error {
-	return emulated.UnwrapHintWithNativeInput(inputs, outputs, func(field *big.Int, inputs, outputs []*big.Int) error {
+func scalarMulHint(field *big.Int, inputs []*big.Int, outputs []*big.Int) error {
```
gbotrel, line 97: I guess unrelated to the PR ?

gbotrel (approved):
lgtm, small cosmetic comments

frontend/schema/leaf.go
ivokub: Oh, I was auto-coding and didn't connect in my brain. I definitely need to change it.

std/algebra/emulated/sw_emulated/hints.go
ivokub: Not completely unrelated. The thing is that previously we did 
```go
var fp emparams.P256Fp
var fr emparams.P256Fr
```
etc. and then `PXLimbs := inputs[:fp.NbLimbs()]` etc etc. And like this for every curve/emulation-param-pairs this scalar multiplication supports. Now, with dynamic emulation parameters I would instead need to do:
```
nbLimbs, nbBits := emulated.GetEffectiveParameters[emparams.P256Fp](field)
```
and then `PXLimbs := inputs[:nbLimbs]`.

But - there is a problem, because we're working over two different fields (scalar and base field of the curve), then we cannot use `Field[T].NewHint` in the circuit side and we passed the limbs of the non-native elements directly. But now the initialization is more efficient i.e. when we initialize either scalar or point as constant then we don't use full number of limbs, but only as much we actually use (this happens for example for PLONK verification because KZG SRS is fixed so first input to MSM is with constant point). So, we need to also pass in to the hint arguments the number of limbs for Px, Py, S.

However - I did have an unrelated idea to improve this hint. I can also pass the curve coefficients, then this method would be completely curve-agnostic. Will do it in the future.

std/algebra/emulated/sw_bls12381/map_to_g1.go
ivokub: It does not, indeed. Fixed

frontend/schema/leaf.go
ivokub: Fixed

backend/witness/witness.go
ivokub: Actually we already have the underlying field in `schema.Schema`, it wasn't just available. I made it public and I'm providing it here.

ivokub: Also added comment to `schema.Walk` that the field will be passed.

ivokub:
> lgtm, small cosmetic comments

Thanks for the review! Implemented all the comments.

PR #1493, opened 2025-05-13
feat: add field extension support over native fields

ivokub:
# Description

For the upcoming support for the non-native arithmetic over native small fields we need to perform multiplication checks in a field extension to obtain the required soundness level. Instead of doing an ad-hoc implementation in the field emulation package, we separate it into a standard interface so it could be reused.

The implementation is currently very basic and supports only specific types of field extensions (`x^d + a` and `x^d + 1`). Will extend the package if the need arises.

Currently it doesn't support inversion and division as there wasn't direct need for it directly in the field emulation package. However, if we want to support logderivative argument in native small field, then we need to have inversion/division. However, gnark wouldn't directly have backends targeting small fields and the selector columns and witness assignment would be exported to Linea prover, which would handle range checks itself. And currently range checks is the only use case we would need logderivative argument.

But I think it would be good to have in case we would have some other lookup usages in circuits used for Linea. @yelhousni, @ThomasPiellard - do we have a good reference for a generic inversion algorithm in gnark or gnark-crypto over any field and extension?

## Type of change

# How has this been tested?

- [x] TestReduce
- [x] TestAdd
- [x] TestSub
- [x] TestMul

# Checklist:


yelhousni:
For the inverse, we implement it in gnark-crypto generically for ext. 2 and 3 as in Alg.8 and 17 resp. in https://eprint.iacr.org/2010/354.pdf.

yelhousni (approved):
Looks good. Although 2 remarks:
- what is the point of having a generic implementation only for baby/koala bear? we can have `Mul()` optimised if we specialise the implementation. 
- we can also add `Square()` as a separate method.

ivokub:
Merged currently as is - it is good enough for unblocking the rest of the PRs, but there are still many things to do to make the gadget more performant and useful. Will create issue for that.

PR #1495, opened 2025-05-16
feat: implement field emulation over small fields

ivokub:
# Description

This PR adds capability for the multiplication checks in the field emulation to be performed in small fields. Namely, for the multiplication checks we sample the challenge in field extension and use it instead of a field element.

Depends on #1492, #1493 and #1494. Will rebase after other PRs are merged. For reviewing, have a look at the latest non-merge commits (particularly 1d314a151f1b80f273c35f9d2c87e913c4f6b1fc).

## Type of change

# How has this been tested?

Added tests to compile and solve (both with actual solver and test engine) over small fields.

# Checklist:


std/math/emulated/field_mul.go
```diff
@@ -289,43 +356,92 @@ func (f *Field[T]) performDeferredChecks(api frontend.API) error {
...
+		multicommit.WithWideCommitment(api, func(api frontend.API, commitment []frontend.Variable) error {
+			// for efficiency, we compute all powers of the challenge as slice at.
+			coefsLen := int(f.fParams.NbLimbs())
+			for i := range f.deferredChecks {
+				coefsLen = max(coefsLen, f.deferredChecks[i].maxLen())
+			}
+			at := make([]fieldextension.Element, coefsLen)
+			at[0] = commitment
+			for i := 1; i < len(at); i++ {
+				atexti := at[i-1]
```
ThomasPiellard, line 411: what is the purpose of `atexti` here ? Why not write `at[i] = f.extensionApi.Mul(at[i-1], commitment)` directly ?

ThomasPiellard (approved):

std/math/emulated/field_mul.go
ivokub: Good catch! We don't need it separately and can do it inline.

PR #1502, opened 2025-05-28
fix hacky registration of MiMC

Tabaie:


std/hash/hash.go
```diff
@@ -156,3 +157,10 @@ func (h *merkleDamgardHasher) Write(data ...frontend.Variable) {
 func (h *merkleDamgardHasher) Sum() frontend.Variable {
 	return h.state
 }
+
+func init() {
```
ivokub, line 161: Usually the registration happens the other way round - we call `hash.Register` inside the MiMC package. If we do it here, then we cannot import `std/hash` in `std/hash/mimc` anymore as it would create an import cycle.

Now, the problem becomes how to ensure that the `std/hash/mimc` is always imported to ensure that the `init()` method of the `std/hash/mimc` is called. Now, the idiomatic way is to have another package which imports all the hash function packages to ensure the initialization functions are always called. In this case, we could add package `std/hash/all` which has a single file
```go
package all

import (
    _ "github.com/consensys/gnark/std/hash/mimc"
    _ "github.com/consensys/gnark/std/hash/poseidon2"
    // ...
)
```
and to register all hashes we just need to import `_ "github.com/consensys/gnark/std/hash/all"`. See https://go.dev/doc/effective_go#blank_import and for example https://github.com/Consensys/gnark-crypto/blob/master/hash/all/allhashes.go.

Tabaie:
Seems like having `mimc` importing `hash` already creates a cycle.
Because `BinaryHasher` imports `uints` which ultimately imports `mimc`. Perhaps we should move `BinaryHasher` to another package?

ivokub:
> Seems like having `mimc` importing `hash` already creates a cycle. Because `BinaryHasher` imports `uints` which ultimately imports `mimc`. Perhaps we should move `BinaryHasher` to another package?

I'll check the import cycle. Maybe indeed having the `BinaryHasher` inside `std/hash` is not optimal, but I'll see maybe there is a way which still would allow keeping it there.

ivokub (approved):

PR #1504, opened 2025-06-03
feat: GKR Add Instance

Tabaie:
Change the GKR API so that defining the circuit and assigning instances are separated. It removes the necessity of defining hints for the output and deferred "finalize" functions.

Remove the `Println` function in favor of `GetValue`.


constraint/solver/gkrgates/registry.go
```diff
@@ -99,31 +100,62 @@ func WithCurves(curves ...ecc.ID) registerOption {
 // - name is a human-readable name for the gate.
 // - f is the polynomial function defining the gate.
 // - nbIn is the number of inputs to the gate.
-func Register(f gkr.GateFunction, nbIn int, options ...registerOption) error {
-	s := registerSettings{degree: -1, solvableVar: -1, name: GetDefaultGateName(f), curves: []ecc.ID{ecc.BN254}}
+//
+// If the gate is already registered, it will return false and no error.
+func Register(f gkr.GateFunction, nbIn int, options ...registerOption) (registered bool, err error) {
```
ivokub: It would be nice to make `registerOption` public. Otherwise it is a bit strange that you have a public method which returns instance of private type. It works but not imo idiomatic Go.

constraint/solver/gkrgates/registry.go
```diff
@@ -99,31 +100,62 @@ func WithCurves(curves ...ecc.ID) registerOption {
...
+	s := registerSettings{degree: -1, solvableVar: -1, name: GetDefaultGateName(f)}
 	for _, option := range options {
 		option(&s)
 	}
 
-	for _, curve := range s.curves {
+	curvesForTesting := s.curves
```
ivokub, line 157: For gate degree testing - do we need an option at all? Cannot we just take some "good" modulus and test over that field. For me it makes a bit confusing that the user can choose the curve, but then needs to know the relation between the gate degree and modulus.

constraint/solver/gkrgates/registry.go
ivokub: And it usually is better to make `registerOption` as `func(*registerSettings) error` to return errors. It seems that some options are incompatible with each other and it would be good to detect during option parsing.

constraint/solver/gkrgates/registry.go
```diff
@@ -99,31 +100,62 @@ func WithCurves(curves ...ecc.ID) registerOption {
 // - name is a human-readable name for the gate.
 // - f is the polynomial function defining the gate.
 // - nbIn is the number of inputs to the gate.
-func Register(f gkr.GateFunction, nbIn int, options ...registerOption) error {
-	s := registerSettings{degree: -1, solvableVar: -1, name: GetDefaultGateName(f), curves: []ecc.ID{ecc.BN254}}
+//
+// If the gate is already registered, it will return false and no error.
```
ivokub, line 145: Why do we need to return boolean to indicate if the gate was registered or not? When you return an error, then it is always false. And when the gate is already registered with same properties, then it is a no-op.

It seems you only use it for testing, I think it doesn't make sense to break APIs and add extra return values you don't use outside of tests.

constraint/solver/gkrgates/registry_test.go
```diff
@@ -47,15 +65,23 @@ func TestRegisterDegreeDetection(t *testing.T) {
...
 		zeroGate := func(api gkr.GateAPI, x ...frontend.Variable) frontend.Variable {
 			return api.Sub(x[0], x[0])
 		}
-		assert.Equal(t, expectedError, Register(zeroGate, 1, WithName(gateName)))
 
-		assert.Equal(t, expectedError, Register(zeroGate, 1, WithName(gateName), WithDegree(2)))
+		// Attempt to register the zero gate without specifying a degree
+		registered, err := Register(zeroGate, 1, WithName(gateName))
+		assert.Error(t, err, "error must be returned for zero polynomial")
+		assert.EqualError(t, err, expectedError, "error message must match expected error")
```
ivokub: I think it is not idiomatic Go to compare errors by messages. If you really want to ensure that some error is what you expect, then define a new error type
```go
var errSomeKnownError = errors.New("error msg")
///

err := functionThatReturnsErrors()
errors.Is(err, errSomeKnownError)
```

or if you want to add context to error then implement `error` interface
```go
type parametricError struct { name string }
func (pe parametricError) Error() string {
    return fmt.Sprintf("error with parameter: %s", pe.name)
}

func functionThatReturnsErrors() error {
    return &parametricError{name: "aaaa"}
}

err := functionThatReturnsErrors()
var wantedError *parametricError
if errors.As(err, &parametricError) {
    // error is *parametricError
}
```

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,131 @@ import (
...
-	d.circuit = info.Circuit
-	d.circuit.SetNbUniqueOutputs()
+type newSolvingDataSettings struct {
+	assignment gkrtypes.WireAssignment
+}
 
-	d.assignment = make(WireAssignment, len(d.circuit))
-	for i := range d.assignment {
-		d.assignment[i] = make([]fr.Element, info.NbInstances)
+type newSolvingDataOption func(*newSolvingDataSettings)
```
ivokub: Should be public type if `With...` options are public. Also function should return errors and you should check that the options are not incompatible with each other.

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,131 @@ import (
...
+
+func WithAssignment(assignment gkrtypes.WireAssignment) newSolvingDataOption {
```
ivokub: Add comments

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,131 @@ import (
...
+	return func(s *newSolvingDataSettings) {
+		s.assignment = assignment
 	}
 }
 
-// this module assumes that wire and instance indexes respect dependencies
+func NewSolvingData(info gkrtypes.SolvingInfo, options ...newSolvingDataOption) *SolvingData {
```
ivokub: Add method documentation

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -144,3 +154,13 @@ func ProveHint(hashName string, data *SolvingData) hint.Hint {
 
 	}
 }
+
+// RepeatUntilEnd for each wire, sets all the values starting from n > 0 to its predecessor.
+// e.g. {{1, 2, 3}, {4, 5, 6}}.RepeatUntilEnd(2) -> {{1, 2, 2}, {4, 5, 5}}
+func (a WireAssignment) RepeatUntilEnd(n int) {
```
ivokub: It is only used once. I would inline it in `ProveHint` with a comment. Particularly as you have anyway inlined it in SolveHint, so it is not generally usable.

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,131 @@ import (
...
-						}
-					}
+	nbPaddedInstances := int(ecc.NextPowerOfTwo(uint64(info.NbInstances)))
+	for i := range d.assignment {
+		d.assignment[i] = make([]fr.Element, nbPaddedInstances)
+	}
+
+	if s.assignment != nil {
+		if len(s.assignment) != len(d.assignment) {
+			panic(fmt.Errorf("provided assignment has %d wires, expected %d", len(s.assignment), len(d.assignment)))
```
ivokub: When you panic, then you don't need to create a new error using `fmt.Errorf`. Just use `fmt.Sprintf`. See also below.

constraint/solver/gkrgates/registry.go
```diff
@@ -135,16 +167,14 @@ func Register(f gkr.GateFunction, nbIn int, options ...registerOption) error {
...
 
-	gatesLock.Lock()
-	defer gatesLock.Unlock()
-	gates[s.name] = gkrtypes.NewGate(f, nbIn, s.degree, s.solvableVar)
-	return nil
+	gates[s.name] = gkrtypes.NewGate(f, nbIn, s.degree, s.solvableVar, allowedCurves)
+	return true, nil
 }
 
 func Get(name gkr.GateName) *gkrtypes.Gate {
```
ivokub, line 229: Add method documentation.

constraint/solver/gkrgates/registry.go
```diff
@@ -160,6 +190,7 @@ type gateVerifier struct {
 	isAdditive   func(f gkr.GateFunction, i int, nbIn int) bool
 	findDegree   func(f gkr.GateFunction, max, nbIn int) (int, error)
 	verifyDegree func(f gkr.GateFunction, claimedDegree, nbIn int) error
+	equal        func(f1, f2 gkr.GateFunction, nbIn int) bool
 }
 
 func NewGateVerifier(curve ecc.ID) (*gateVerifier, error) {
```
ivokub: Add method documentation

internal/gkr/gkrtypes/types.go
```diff
@@ -17,15 +19,27 @@ type Gate struct {
 	nbIn        int              // number of inputs
 	degree      int              // total degree of the polynomial
 	solvableVar int              // if there is a variable whose value can be uniquely determined from the value of the gate and the other inputs, its index, -1 otherwise
+	curves      []ecc.ID         // curves that the gate is allowed to be used over
 }
 
-func NewGate(f gkr.GateFunction, nbIn int, degree int, solvableVar int) *Gate {
+func NewGate(f gkr.GateFunction, nbIn int, degree int, solvableVar int, curves []ecc.ID) *Gate {
```
ivokub, line 26: Add method documentation

internal/gkr/gkrtypes/types.go
```diff
@@ -17,15 +19,27 @@ type Gate struct {
...
 		evaluate:    f,
 		nbIn:        nbIn,
 		degree:      degree,
 		solvableVar: solvableVar,
+		curves:      curves,
+	}
+}
+
+func (g *Gate) SupportsCurve(curve ecc.ID) bool {
+	for _, c := range g.curves {
```
ivokub: Can use `slices.Contains`

internal/gkr/gkrtypes/types.go
ivokub: I don't understand why we need the new argument `curves`. What is a realistic case where gate is incompatible with some curve? I think it is only used for degree testing? But we're already giving the degree as explicit argument?

internal/gkr/small_rational/gate_testing.go
```diff
@@ -142,6 +142,17 @@ func VerifyGateFunctionDegree(f gkr.GateFunction, claimedDegree, nbIn int) error
 	return nil
 }
 
+// EqualGateFunction checks if two gate functions are equal, by testing the same at a random point.
+func EqualGateFunction(f gkr.GateFunction, g gkr.GateFunction, nbIn int) bool {
```
ivokub: This method doesn't seem to be used anywhere? Is this package for doing GKR over small fields? In this case this method is unsound?

internal/utils/slices.go
```diff
@@ -16,3 +16,16 @@ func References[T any](v []T) []*T {
...
+func ExtendRepeatLast[T any](s []T, n int) []T {
+	if n <= len(s) {
+		return s[:n]
+	}
+	res := make([]T, n)
+	copy(res, s)
+	for i := len(s); i < n; i++ {
+		res[i] = res[i-1]
+	}
+	return res
```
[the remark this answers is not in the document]
ivokub: False positive

internal/utils/slices.go
```diff
@@ -16,3 +16,16 @@ func References[T any](v []T) []*T {
 	}
 	return res
 }
+
+// ExtendRepeatLast extends a non-empty slice s by repeating the last element until it reaches the length n.
+func ExtendRepeatLast[T any](s []T, n int) []T {
+	if n <= len(s) {
+		return s[:n]
+	}
+	res := make([]T, n)
```
ivokub, line 25: Can use `slices.Grow`

std/gkrapi/compile.go
```diff
@@ -15,196 +15,238 @@ import (
...
 		}
 	}
 
-	// arrange inputs wire first, then in the order solved
-	ins := make([]frontend.Variable, 0, solveHintNIn)
-	for i := range circuit {
-		if circuit[i].IsInput() {
-			appendNonNil(&ins, api.assignments[i])
+	res.toStore.GetAssignmentHintID = solver.GetHintID(res.hints.GetAssignment)
+	res.toStore.ProveHintID = solver.GetHintID(res.hints.Prove)
```
ivokub: `res.hints` is test engine hints. Why do we store it here? And it seems you overwrite ProveHintID later in `verify` method?

internal/gkr/engine_hints.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (h *TestEngineHints) Prove(mod *big.Int, ins, outs []*big.Int) error {
+
+	info, err := gkrtypes.StoringToSolvingInfo(*h.info, gkrgates.Get)
+	if err != nil {
+		return fmt.Errorf("failed to convert storing info to solving info: %w", err)
+	}
+
+	// TODO @Tabaie autogenerate this or decide not to
```
ivokub: If it is a small thing to implement, then better just do it or if it is a larger task, then create a Github issue. Code inline TODOs tend no to be implemented.

std/gkrapi/example_test.go
```diff
@@ -19,18 +19,22 @@ func Example() {
 	// This means that the imported fr and fp packages are the same, being from BW6-761 and BLS12-377 respectively. TODO @Tabaie delete if no longer have fp imported
```
ivokub: Avoid inline TODOs.

std/gkrapi/example_test.go
```diff
@@ -19,18 +19,22 @@ func Example() {
...
 	// the proof is being computed in the same session as the
 	// SNARK circuit being compiled.
 	// But in production applications it would be necessary.
 
-	assertNoError(gkrgates.Register(squareGate, 1))
-	assertNoError(gkrgates.Register(sGate, 4))
-	assertNoError(gkrgates.Register(zGate, 4))
-	assertNoError(gkrgates.Register(xGate, 2))
-	assertNoError(gkrgates.Register(yGate, 4))
+	_, err := gkrgates.Register(squareGate, 1)
```
ivokub: Imo here it is a good example why imo it doesn't make sense to return boolean indicating if the gate was registered or not. You always discard the result (outside of tests), so why return it at all?

std/gkrapi/compile.go
```diff
@@ -15,196 +15,238 @@ import (
...
+	res.toStore.SolveHintID = solver.GetHintID(res.hints.Solve)
+
+	parentApi.Compiler().Defer(res.finalize)
+
+	return &res
+}
+
+// AddInstance adds a new instance to the GKR circuit, returning the values of output variables for the instance.
+func (c *Circuit) AddInstance(input map[gkr.Variable]frontend.Variable) (map[gkr.Variable]frontend.Variable, error) {
```
ivokub, line 108: I would keep Import method for backwards compatibility. Otherwise you really break all downstream usage of GKR and do not give good migration path for upgrading gnark as the only option is to rewrite all GKR circuits. And I think keeping `Import` is trivial as you could just do loop of `AddInstance` inside.

std/gkrapi/compile.go
```diff
@@ -15,196 +15,238 @@ import (
...
+	if len(input) != len(c.ins) {
+		for k := range input {
+			if k >= gkr.Variable(len(c.ins)) {
+				return nil, fmt.Errorf("variable %d is out of bounds (max %d)", k, len(c.ins)-1)
+			}
+			if !c.toStore.Circuit[k].IsInput() {
+				return nil, fmt.Errorf("value provided for non-input variable %d", k)
+			}
+		}
```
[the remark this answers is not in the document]
ivokub: Check if the cursor comment makes sense. I do not follow why you are comparing against `len(c.ins)`

std/gkrapi/compile.go
```diff
@@ -15,196 +15,238 @@ import (
...
-		proveHintPlaceholder, gadget.ProofSize(forSnark.circuit, logNbInstances), hintIns...); err != nil {
+	if proofSerialized, err = api.Compiler().NewHint(
+		c.hints.Prove, gadget.ProofSize(forSnark.circuit, bits.TrailingZeros(uint(len(c.assignments[0])))), hintIns...); err != nil {
 		return err
 	}
-	s.toStore.ProveHintID = solver.GetHintID(proveHintPlaceholder)
+	c.toStore.ProveHintID = solver.GetHintID(c.hints.Prove)
 
-	forSnarkSorted := utils.MapRange(0, len(s.toStore.Circuit), slicePtrAt(forSnark.circuit))
+	forSnarkSorted := utils.MapRange(0, len(c.toStore.Circuit), slicePtrAt(forSnark.circuit))
```
ivokub: Imo it is difficult to follow what is happening here. I need to know what `MapRange` and `slicePtrAt` does, but neither have documentation and `slicePtrAt` returns a function. I really recommend just doing small loops inline, otherwise I need to jump several times to see differeunt function definitions to understand what is going on and why.

ivokub: So, for example, here what you do I think is:
```go
forSnarkSorted := make([]*gkrtypes.Wire, len(c.ToStore.Circuit))
for i := range forSnarkSorted {
    forSnarkSorted[i] = &forSnark.circuit[i]
}
```
and for me it is a lot more understandable and explicit. Now, you only use `utils.MapRange` and `slicePtrAt` only once, exactly here. So instead of four lines of explicit implementation there are several function definitions and more difficult reviewing.

std/permutation/poseidon2/gkr-poseidon2/gkr_test.go
```diff
@@ -67,3 +74,27 @@ func TestGkrPermutationCompiles(t *testing.T) {
 	require.NoError(t, err)
 	fmt.Println(cs.GetNbConstraints(), "constraints")
 }
+
+func BenchmarkGkrPermutations(b *testing.B) {
```
ivokub: Leftover? If not, then remove cpu profiles, otherwise running tests create new files in the working tree.

std/permutation/poseidon2/gkr-poseidon2/gkr_test.go
```diff
@@ -67,3 +74,27 @@ func TestGkrPermutationCompiles(t *testing.T) {
...
+	cs, err := frontend.Compile(ecc.BLS12_377.ScalarField(), scs.NewBuilder, &circuit)
+	require.NoError(b, err)
+
+	witness, err := frontend.NewWitness(&assignmment, ecc.BLS12_377.ScalarField())
+	require.NoError(b, err)
+
+	// cpu profile
+	f, err := os.Create("cpu.pprof")
+	require.NoError(b, err)
+	defer func() {
```
ivokub: Just call `defer f.Close()`, no need to wrap in a function doing `NoError`, as it is unhandleable error and usually indicates filesystem corruption.

std/permutation/poseidon2/gkr-poseidon2/gkr_test.go
ivokub: Or if you wanted to create a benchmark, then please see https://pkg.go.dev/testing#hdr-b_N_style_benchmarks how to write a benchmark function. Namely. it should run the part you want to benchmark in a loop
```go
prepareWork()
b.ResetTimer() // if prepareWork takes long time
for i := b.N {
    methodToBenchmark()
}
```
Benchmarking framework automatically sets `b.N` value to be large enough to have sufficient sample for reliable benchmarking results (and you can also set it when you call `go test -bench=...`)

ivokub (changes requested):
I think the idea is good, makes it a bit simpler to use in some cases. But I would leave the current way to import variables for backwards compatibility to avoid breaking all downstream GKR usage. We could implement the changes in Linea ourselves, but gnark is also used elsewhere and breaking API changes essentially prevent our users to upgrade gnark easily. It also creates some issues when we want to backport some hotfixes for Linea as we would need to start maintaining separate branch preceeding the API breakage.

In general, I would recommend adding a lot more method documentation what some methods do, what are the inputs, what are the outputs and what are the assumptions which may cause the methods to return errors. Right now many methods lack documentation so it is trial and error to figure out how to use them correctly.

Also keep in mind that try to keep type visibility consistent. It goes particularly for options -- when you have an option type and methods which return options then both the option type and these methods should be public. A la
```go
type cfg struct {} // dont make it public unless you need to use configuration in other package

type Option func(*cfg) error // make Option type public, as it then also shows it in package documentation and groups all methods returning options together. NB! Don't forget to add documentation what the options are for, where you can use it etc.

func WithDoThis() Option { // function should be public so that can see in the pkg documentation. NB! Don't forget to add documentation what the method does, what is the behaviour if it is not set and what are conflicting options
    return func(conf *cfg) error {
        // now check here if this option is not conflicting with some existing configuration. A la when cfg has fields `doThis, doThat bool` then you should check that `doThat` is not already set and if it is then return an error
       if conf.doThat {
           return errors.New("WithDoThat option already set")
       }
   }
}
```

I reviewed mostly the coding part - it has become quite difficult to understand how the full GKR implementation works end to end and if it is sound, as there are already quite a lot of changes compared to the audited version. I'd add soundness tests i.e. what happens if you try to explicitly modify GKR outputs so that they are not consistent with the expected output. Or what happens if you overwrite GKR hints to modify the results.

std/gkrapi/compile.go
Tabaie: Changed to one (documented) utility function rather than two.

constraint/solver/gkrgates/registry.go
Tabaie: Yes it may be best to pick a random 144 bit or so modulus and just test over that.

std/gkrapi/compile.go
Tabaie: Addressed in https://github.com/Consensys/gnark/pull/1589

constraint/solver/gkrgates/registry.go
Tabaie: While I think it is a good idea to test over random fields, there actually is a good reason to want to define a gate function over a certain curve only. A good example are hash round functions, which contain hard-coded constants, which for performance reasons it makes sense to keep as concrete types such as `fr.Element`. In such a case it's better to stop an attempt to use the gate under an incompatible curve early on instead of a more opaque `SetInterface` panic in proving time.

constraint/solver/gkrgates/registry.go
Tabaie: Yes I think you're right - it's not worth breaking compatibility for.

ivokub: If you absolutely need to know if something was registered or not, then you can define a new option `ErrorOnExisting` etc which would make the gate registration error if key already exists. And then only return an error from `Register`

internal/gkr/bls12-377/solver_hints.go
Tabaie: Note that these are internal, non-user facing functions.

internal/gkr/bls12-377/solver_hints.go
Tabaie: I tried inlining but readability suffered. I turned it private since it wasn't used elsewhere.

internal/gkr/gkrtypes/types.go
Tabaie: There is a good reason to limit which curves the gate can be used over, but perhaps we can decouple it from testing? i.e. the testing be done over a random field still.

internal/gkr/small_rational/gate_testing.go
Tabaie: This package is used to generate common prover/verifier test cases for all curves.

internal/utils/slices.go
Tabaie: `slices.Grow` increases the capacity of the slice, while here we increase the actual size, and in any case it could at most save one line of code.

internal/gkr/small_rational/gate_testing.go
Tabaie: You're right that none of the gate testing functions were used in this package. I deleted that file.

std/gkrapi/compile.go
```diff
@@ -1,240 +1,280 @@
...
+
+// GetValue is a debugging utility returning the value of variable v at instance i.
+// While v can be an input or output variable, GetValue is most useful for querying intermediate values in the circuit.
+func (c *Circuit) GetValue(v gkr.Variable, i int) frontend.Variable {
+	// last input to ensure the solver's work is done before GetAssignment is called
+	res, err := c.api.Compiler().NewHint(c.hints.GetAssignment, 1, int(v), i, c.assignments[c.outs[0]][i])
+	if err != nil {
+		panic(err)
 	}
+	return res[0]
```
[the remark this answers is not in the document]
Tabaie: Compile now returns an error if no output variables are present.

Tabaie:
Included in https://github.com/Consensys/gnark/pull/1661

PR #1505, opened 2025-06-03
fix: range over all limbs in normalized form for lookup

ivokub:
# Description

For non-native arithmetic Lookup2 and Mux we used Lookup2 and Mux on individual limbs. We normalized the inputs so that we would always have the same number of limbs, but when we did the final Lookup2/Mux, we only range over the number of limbs the first input had.

## Type of change

# How has this been tested?

- [x] TestLookup2AndMuxOnAllLimbs

# Checklist:


yelhousni (approved):
good catch!

PR #1533, opened 2025-07-05
Fix typos in documentation for G1 point compression in Solidity verifier

zeevick10:
---

## Description
This PR updates the inline documentation in `backend/groth16/bn254/solidity.go` to correct minor typos:
- Consistently uses “compressed” instead of “compresed” in function parameter and return descriptions.

---


ivokub:
Thanks for the contribution. Unfortunately we don't accept small typo fix PRs anymore.

PR #1538, opened 2025-07-11
Add Celer Network to known gnark users

donatik27:
This commit adds Celer Network to the KNOWN_USERS.md list.
Celer has implemented a GPU-accelerated backend for gnark to improve Groth16 and PLONK proof generation performance in high-throughput environments.
Reference: https://celer.network/


ivokub:
Thanks for the contribution. I understand you're not part of the Celer team? We would like to get direct contributions for KNOWN_USERS from the teams only if they would like to associate their projects with gnark.

If not confirmed, then I'll close the PR without merging soon.

PR #1489, opened 2025-05-05
Feat/eip4844 precompile

ThomasPiellard:
# Description

Implementation of the EIP4844 precompile.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How has this been tested?

See marshall_test.go, kzg_point_evaluation_test.go

# How has this been benchmarked?

`Unmarshall` #constraints: 1772 when compiled on BN254, emulating BLS12-381 base field
`UnmarshalCompressed` #constraints: 585600 when compiled on BN254, emulating BLS12-381 base field
`KzgPointEvaluation` #constraints: 3772424 when compiled on BN254, emulating BLS12-381 base field

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1541, opened 2025-07-14
feat: add separate Bytes API in `std/math/uints` package for handling bytes (U8) directly

### LINKED ISSUE #911: fix: range checker should handle edge cases separately
Lookup-table based range checker should handle separately cases when requested bits count is 0 or 1:
* for 0 bits the input must be 0
* for 1 bits the input must be 0 or 1 (perform boolean test)

ivokub:
# Description

Previously the operations were define over U32 and U64 (as needed or hash functions). But for working directly with bytes didn't allow using boolean operations directly on bytes etc.

Additionally, made the uints API cacheable.

It also resolved #911.

Needed for #1489 refactoring what I'm working on.

## Type of change

# How has this been tested?

- [x] TestKeyStore for testing caching

# Checklist:

### DIFF
--- a/std/math/uints/bytes.go
+++ b/std/math/uints/bytes.go
@@ -1,0 +1,102 @@
+package uints
+
+import (
+	"fmt"
+
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/internal/kvstore"
+	"github.com/consensys/gnark/std/internal/logderivprecomp"
+	"github.com/consensys/gnark/std/rangecheck"
+)
+
+// ctxKey is used to store the API in the key-value store. In case we
+// re-initialize using [NewBytes] method, then we can reuse the existing API. This
+// ensures we don't rebuild any of the tables.
+type ctxKey struct{}
+
+// Bytes implements methods for manipulating bytes in circuit. Use [NewBytes] to
+// create a new instance.
+type Bytes struct {
+	api             frontend.API
+	xorT, andT, orT *logderivprecomp.Precomputed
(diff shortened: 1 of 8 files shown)


std/math/uints/bytes.go
```diff
@@ -0,0 +1,229 @@
...
+	// default (as the argument is uint8). If it is used as a witness
+	// assignment, then the flag `internal` is not set for the actual witness
+	// value inside the circuit, as witness parser only copies
+	// [frontend.Variable] part of U8. And the `internal=false` is set in the
+	// [U8.Initialize] method.
+	return U8{Val: v, internal: true}
+}
+
+// Initialize describes how to initialise the element.
+func (e *U8) Initialize(field *big.Int) {
```
ThomasPiellard: I think this function is not used ? And if it is the `field` argument is not used

ThomasPiellard (approved):

std/math/uints/bytes.go
ivokub: It is used indirectly - the schema parser calls `Initialize` method during walking of the circuit and witness, this ensures that all the witness variables are initialized and flags properly set.

See https://github.com/Consensys/gnark/blob/master/frontend/schema/leaf.go

ThomasPiellard: ah right

ivokub: But I think you're right - we don't do any significant initialization and can use the default initialization (do nothing) as the default values should be good enough (Val = nil and internal = false)

PR #1542, opened 2025-07-14
feat: add conversion utility to convert between bytes and elements

ivokub:
# Description

Utility for converting between bytes and nonnative/native. Needed for EIP-4844 #1489  as we need to compare that the "versioned hash" matches the commitment.

It also allows for an use case where if a circuit requires as an input some byte array, then we can provide it as a single native element and then directly decompose it into bytes in-circuit. This allows to significantly decrease the number of public inputs for some circuits.

The technique used here could also be used in non-native arithmetic for asserting less-equal. Currently we use binary decomposition and compare bit by bit (and also when doing strict modular reduction), but using range checker we could significantly reduce number of constraints. I'll do it later in a separate PR.

Currently it is not fully complete, it works for emulated only in case emulation parameters have limb width divisible by 8. It is currently the case, but changes for small fields. But at least unblocks some work elsewhere for now and we can implement it later.

## Type of change

# How has this been tested?

- [x] TestBytesToEmulatedDivisible
- [x] TestBytesToNative
- [x] TestNativeToBytes
- [x] TestEmulatedToBytes
- [x] TestAssertBytesLeq

# Checklist:


std/conversion/conversion.go
```diff
@@ -0,0 +1,316 @@
...
+	"math/bits"
+
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/std/math/emulated"
+	"github.com/consensys/gnark/std/math/uints"
+	"github.com/consensys/gnark/std/rangecheck"
+)
+
+// should have:
+//  - convert from bytes to native field element ✅
```
gbotrel: I guess this can go in the package doc

std/conversion/conversion.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	res := bytesToNative(api, b)
+	// check that the input was in range of the field modulus. Omit if cfg.allowOverflow is set.
+	if !cfg.allowOverflow {
+		assertBytesLeq(api, b, api.Compiler().Field())
+	}
+	return res, nil
+}
+
+func bytesToNative(api frontend.API, b []uints.U8) frontend.Variable {
```
gbotrel: just to be sure here we have some soft guarantee that uints.U8 are actually constrained to be bytes yes?

std/conversion/conversion.go
```diff
@@ -0,0 +1,316 @@
...
+		if len(res) != nbBytes {
+			return nil, fmt.Errorf("expected %d bytes, got %d", nbBytes, len(res))
+		}
+		res = res[uint(nbBytes)-nbLimbBytes:] // take only the last nbLimbBytes bytes
+		for j := range nbLimbBytes {
+			resU8[uint(i)*nbLimbBytes+j] = uapi.ValueOf(res[j])
+		}
+		computed := bytesToNative(api, resU8[uint(i)*nbLimbBytes:uint(i+1)*nbLimbBytes])
+		api.AssertIsEqual(vr.Limbs[len(vr.Limbs)-i-1], computed)
+	}
```
gbotrel, line 286: so, not sure it's an issue but since we init resU8 with the byte len of fr.Modulus might be an assumption that I missed. Do we actually constrain here the final result (i.e. the resU8) to be strictly smaller than fr.Modulus ?

std/conversion/conversion.go
```diff
@@ -0,0 +1,316 @@
...
+	//  - b[i] == mBytes[i]. If this happens then we need to have the same checks for
+	//    b[i-i] and mBytes[i-1] etc.
+	//
+	// Now, to check that b[i] <= mBytes[i] we can use the range checker gadget
+	// to ensure that the difference mBytes[i]-b[i] is non-negative by checking
+	// that it has up to |mBytes[i]| bits.
+	var eq_i frontend.Variable = 1
+	for i := range mBytes {
+		// compute the difference
+		diff := api.Sub(mBytes[i], bb[i].Val)
```
gbotrel: nice loop -- same remark as before I assume it means we can't construct unconstraint uint8.Val that are not bytes

std/conversion/conversion.go
ivokub: Here we do the check: https://github.com/Consensys/gnark/pull/1542/files/786608173f59eb9f665c68a48a28a0fad65a6c12#diff-55d61d5299a82cc7aec9136ff48f11953a1ce4e0054b7c383f2bc62457b9c426R241-R245

We actually consider consider two cases (and default to more strict case):
- when the option WithAllowOverflow is given, then we only check that the width of emulated element is |fr.Modulus|. This in practice means that the integer representation can either be x or x+fr. But in many cases this is sufficient (i.e. we marshal for transport, not for hashing)
- but when we decompose into bytes for hashing for example, then we need to ensure unique representation. Then we use `f.ReduceStrict` which modreduces and checks that the result is strictly less than modulus.

std/conversion/conversion.go
ivokub: This was essentially my self TODO, but I don't want to tag it as such as it invites PR spam...

I'll clean it up and add to package doc

std/conversion/conversion.go
ivokub: I refactored uints package to be more robust and strict. Now indeed we always assure it (and hopefully cache the range checking status as much as possible.)

std/conversion/conversion.go
ivokub: Now it is explicit, I added Bytes.Value method which returns the value and range checks if it is already not.

PR #1543, opened 2025-07-15
fix: Redirect blob_maker.go reference to Linea monorepo

menzudilat:
# Description

Replace non-working URL https://github.com/Consensys/zkevmmonorepo/blob/main/prover/lib/compressor/blob/blob_maker.go#L299 with working URL https://github.com/Consensys/linea-monorepo/blob/main/prover/lib/compressor/blob/v0/blob_maker.go#L299 as the repository has been renamed and file path structure updated.


ivokub:
We don't generally accept small typo fixes anymore. Sorry.

PR #1546, opened 2025-07-17
perf(debug): pre-allocate memory in Stack() function

kianjib7:
# Description

Pre-allocate memory for strings.Builder in the Stack() function to reduce memory reallocations when building stack traces. This improves performance when generating stack traces, especially for deep call stacks during debugging.

### DIFF
--- a/debug/debug.go
+++ b/debug/debug.go
@@ -9,6 +9,7 @@ import (
 
 func Stack() string {
 	var sbb strings.Builder
+	sbb.Grow(1024) // Pre-allocate memory to reduce reallocations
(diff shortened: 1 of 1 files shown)


ivokub:
> # Description
> Pre-allocate memory for strings.Builder in the Stack() function to reduce memory reallocations when building stack traces. This improves performance when generating stack traces, especially for deep call stacks during debugging.

The method is generally only called once in test engine or circuit compile time in case of failure. Do you have any numbers how much it would reduce runtime? And the constant 1024 seems very arbitrary.

PR #1552, opened 2025-07-21
fix: prevent early return in RegisterHint causing missing registrations

avorylli:
Fix bug where RegisterHint would stop registering subsequent hint functions if one function was already registered. 

Changed `return` to `continue` to ensure all functions in the slice are processed.

This bug could cause missing hint function registrations leading to solver errors.


ivokub (approved):
Good catch! As far as I know we haven't stumbled on it yet, but better to fix it now.

PR #1553, opened 2025-07-21
fix: add missing error checks (errcheck linter fixes)

avorylli:
Add proper error handling for:
- EnterExitWalker.Enter/Exit methods in reflectwalk
- fmt.Fprintln/Fprintf calls in dotgraph (explicit ignore with _ assignment)

Fixes 5 errcheck linter violations.


ivokub:
Thanks for the contribution. But the added error checks seem to be on paths which are not taken frequently. We'll reconsider when we add errcheck to our linter configuration.

PR #1555, opened 2025-07-26
chore: update enabled linters in .golangci.yml

forkfury:
Refines .golangci.yml by adjusting the list of enabled linters. Improves static analysis configuration without affecting runtime behavior.


ivokub:
Thanks for the contribution. We don't accept small fixes anymore.

PR #1559, opened 2025-07-29
fix: bls12-381 edge cases

ivokub:
# Description

This PR fixes the following edge cases:
- Miller loop computation when G1 point is 0
- Subgroup membership check for G1 is point is 0

Additionally, it fixes IsOnG1 method which had a typo.

## Type of change

# How has this been tested?

Added corresponding regression tests.

# Checklist:

### DIFF
--- a/std/algebra/emulated/sw_bls12381/g1.go
+++ b/std/algebra/emulated/sw_bls12381/g1.go
@@ -147,6 +147,17 @@ func (g1 G1) doubleAndAdd(p, q *G1Affine) *G1Affine {
 }
 
 func (g1 *G1) scalarMulBySeedSquare(q *G1Affine) *G1Affine {
+	// It computes the scalar multiplication by the seed square. It is used to
+	// verify if a point is in the subgroup of G1. However, it uses incomplete
+	// formulas and as such doesn't work for point at infinity as we get a
+	// division by zero. But as we represent the point at infinity as (0,0),
+	// then in this case we can run the computations using a dummy point (1,1)
+	// and then later replace it again with the point at infinity.
+	isInfinity := g1.api.And(g1.curveF.IsZero(&q.X), g1.curveF.IsZero(&q.Y))
+	q = &G1Affine{
+		X: *g1.curveF.Select(isInfinity, g1.curveF.One(), &q.X),
+		Y: *g1.curveF.Select(isInfinity, g1.curveF.One(), &q.Y),
+	}
 	z := g1.double(q)
 	z = g1.add(q, z)
 	z = g1.double(z)
@@ -168,6 +179,11 @@ func (g1 *G1) scalarMulBySeedSquare(q *G1Affine) *G1Affine {
 	z = g1.doubleN(z, 32)
 	z = g1.doubleAndAdd(z, q)
 	z = g1.doubleN(z, 32)
+
+	z = &G1Affine{
+		X: *g1.curveF.Select(isInfinity, g1.curveF.Zero(), &z.X),
+		Y: *g1.curveF.Select(isInfinity, g1.curveF.Zero(), &z.Y),
+	}
 
 	return z
 }
--- a/std/algebra/emulated/sw_bls12381/pairing.go
+++ b/std/algebra/emulated/sw_bls12381/pairing.go
@@ -315,13 +315,15 @@ func (pr Pairing) AssertIsOnG1(P *G1Affine) {
 // IsOnG1 returns a boolean indicating if the G1 point is on the curve and in
 // the prime subgroup.
 func (pr Pairing) IsOnG1(P *G1Affine) frontend.Variable {
-	// 1 - is Q on curve
+	// 1 - is P on curve
 	isOnCurve := pr.IsOnCurve(P)
-	// 2 - is Q in the subgroup
+	// 2- Check P has the right subgroup order
+	// [x²]ϕ(P)
 	phiP := pr.g1.phi(P)
 	_P := pr.g1.scalarMulBySeedSquare(phiP)
 	_P = pr.curve.Neg(_P)
-	isInSubgroup := pr.g1.IsEqual(_P, phiP)
+	// [r]Q == 0 <==>  P = -[x²]ϕ(P)
+	isInSubgroup := pr.g1.IsEqual(_P, P)
 	return pr.api.And(isOnCurve, isInSubgroup)
 }
 
--- a/std/algebra/emulated/sw_bls12381/pairing.go
+++ b/std/algebra/emulated/sw_bls12381/pairing.go
@@ -399,10 +399,13 @@ func (pr Pairing) millerLoopLines(P []*G1Affine, lines []lineEvaluations, init *
 	xNegOverY := make([]*baseEl, n)
 
 	for k := 0; k < n; k++ {
-		// P are supposed to be on G1 respectively of prime order r.
-		// The point (x,0) is of order 2. But this function does not check
-		// subgroup membership.
-		yInv[k] = pr.curveF.Inverse(&P[k].Y)
+		// P are supposed to be on G1 respectively of prime order r. The point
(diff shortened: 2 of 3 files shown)


std/algebra/emulated/sw_bls12381/g1.go
```diff
@@ -147,6 +147,17 @@ func (g1 G1) doubleAndAdd(p, q *G1Affine) *G1Affine {
...
+	// verify if a point is in the subgroup of G1. However, it uses incomplete
+	// formulas and as such doesn't work for point at infinity as we get a
+	// division by zero. But as we represent the point at infinity as (0,0),
+	// then in this case we can run the computations using a dummy point (1,1)
+	// and then later replace it again with the point at infinity.
+	isInfinity := g1.api.And(g1.curveF.IsZero(&q.X), g1.curveF.IsZero(&q.Y))
+	q = &G1Affine{
+		X: *g1.curveF.Select(isInfinity, g1.curveF.One(), &q.X),
+		Y: *g1.curveF.Select(isInfinity, g1.curveF.One(), &q.Y),
+	}
```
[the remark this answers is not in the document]
ivokub: @yelhousni what do you think? I wanted explicitly not to use a valid point. But the issue is relevant that maybe in future we could have issues when using dummy points (i.e. when we change the add/double formulas)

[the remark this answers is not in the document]
YaoJGalteland: @ivokub  what made you decide to use the invalid point? how about using an on-curve dummy point instead of an off-curve one?  the final Select operation discards these meaningless intermediate results and imposes the correct semantic result ((0,0)) if the original input was infinity.

std/algebra/emulated/sw_bls12381/pairing.go
```diff
@@ -315,13 +315,15 @@ func (pr Pairing) AssertIsOnG1(P *G1Affine) {
...
+	// 1 - is P on curve
 	isOnCurve := pr.IsOnCurve(P)
-	// 2 - is Q in the subgroup
+	// 2- Check P has the right subgroup order
+	// [x²]ϕ(P)
 	phiP := pr.g1.phi(P)
 	_P := pr.g1.scalarMulBySeedSquare(phiP)
 	_P = pr.curve.Neg(_P)
-	isInSubgroup := pr.g1.IsEqual(_P, phiP)
+	// [r]Q == 0 <==>  P = -[x²]ϕ(P)
```
YaoJGalteland: Q --> P

std/algebra/emulated/sw_bls12381/pairing.go
```diff
@@ -315,13 +315,15 @@ func (pr Pairing) AssertIsOnG1(P *G1Affine) {
 // IsOnG1 returns a boolean indicating if the G1 point is on the curve and in
 // the prime subgroup.
```
YaoJGalteland, line 316: I'd like to extend the description a bit, it's a bit difficult to map the relation between the subgroup membership check and scalarMulBySeedSquare: P is a member of the prime-order subgroup of G1 (let its order be r) iff [r]P == O iff P + [x²]ϕ(P) == O (x^2 is the seed square of BLS12-381, the equivalence is a properties of BLS12-381)

std/algebra/emulated/sw_bls12381/pairing.go
```diff
@@ -315,13 +315,15 @@ func (pr Pairing) AssertIsOnG1(P *G1Affine) {
...
 func (pr Pairing) IsOnG1(P *G1Affine) frontend.Variable {
-	// 1 - is Q on curve
+	// 1 - is P on curve
 	isOnCurve := pr.IsOnCurve(P)
-	// 2 - is Q in the subgroup
+	// 2- Check P has the right subgroup order
+	// [x²]ϕ(P)
```
YaoJGalteland, line 330: Check P == -[x²]ϕ(P)?

std/algebra/emulated/sw_bls12381/pairing_test.go
```diff
@@ -294,13 +374,131 @@ func (c *GroupMembershipCircuit) Define(api frontend.API) error {
 
 func TestGroupMembershipSolve(t *testing.T) {
```
YaoJGalteland, line 378: Potential missing edge cases:
1) points of order 2 (e.g., (x,0) where y=0) not in the subgroup but on curve, 
2) off-curve point?

std/algebra/emulated/sw_bls12381/pairing_test.go
```diff
@@ -75,6 +76,85 @@ func TestMillerLoopTestSolve(t *testing.T) {
...
+	}
+	mlres, err := pairing.MillerLoop([]*G1Affine{&c.InG1}, []*G2Affine{&c.InG2})
+	if err != nil {
+		return fmt.Errorf("miller loop: %w", err)
+	}
+	pairing.AssertIsEqual(mlres, &c.Res)
+	return nil
+}
+
+func TestMillerLoopSingleTestSolve(t *testing.T) {
```
YaoJGalteland, line 98: Potential missing edge cases: 
1) points of order 2 (e.g., (x,0) where y=0). 
2) case=invalid, if the millerLoopLines does not check subgroup membership

std/algebra/emulated/sw_bls12381/pairing_test.go
```diff
@@ -294,13 +374,131 @@ func (c *GroupMembershipCircuit) Define(api frontend.API) error {
...
+		return fmt.Errorf("new pairing: %w", err)
 	}
-	err := test.IsSolved(&GroupMembershipCircuit{}, &witness, ecc.BN254.ScalarField())
-	assert.NoError(err)
+	res := pairing.IsOnG1(&c.InG1)
+	api.AssertIsEqual(c.ExpectedIsOnG1, res)
+	return nil
+}
+
+func TestIsOnG1(t *testing.T) {
```
YaoJGalteland, line 534: Potential missing edge cases:
1) off-curve point? `IsOnG1` first calls `pr.IsOnCurve(P)`. If an off-curve point is provided, it should fail

std/algebra/emulated/sw_bls12381/g1.go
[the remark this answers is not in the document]
ivokub: I chose a random point for which I know the following computation holds. I'm replacing the computation result by (0,0) anyway, so I think there is no difference what to use.

std/algebra/emulated/sw_bls12381/pairing_test.go
ivokub: Added

std/algebra/emulated/sw_bls12381/g1.go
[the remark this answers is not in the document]
YaoJGalteland: true, the computational flow is fixed for the constant dummy point. As long as the intermediate operation performs correctly just once for this point, it's safe to use anytime.

std/algebra/emulated/sw_bls12381/pairing_test.go
ivokub: 1. I think the comment is wrong i.e. there is no such point
2. I think in that case the computation result is undefined - I'm not sure I could test it?

std/algebra/emulated/sw_bls12381/pairing_test.go
ivokub: 1. There is no such point
2. added test

std/algebra/emulated/sw_bls12381/pairing.go
ivokub: Updated comments

std/algebra/emulated/sw_bls12381/g1.go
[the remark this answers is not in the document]
yelhousni: I don't think it matters. The point is to use any dummy (x,y) pair that would not panic the methods and then overwrite the result.

PR #1566, opened 2025-08-05
Fix multiple typos

destinyae:
Fix please, i was confused while reading at some point.


ivokub:
We don't take small bug fix contributions anymore

PR #1567, opened 2025-08-05
fix: KZG proof verification for edge cases

ivokub:
# Description

This PR implements complete formulas for KZG proof verification to allow checking proofs for zero polynomials and constructed polynomials.

Maybe there is a more efficient way than using complete formulas for MSM which adds quite a lot of overhead. But it seems that we don't use this method directly, for PLONK recursive verifier we manually prepare the inputs and use batch verification where we already have an option to use complete formulas.

Maybe here we could also add an option for using complete formulas? Although for me it seems that it is a valid case to commit to a zero polynomial and I'd rather have a default behaviour which allows it, rather than not.

For PLONK previously 2701912, now 2772838. Diff 70926
For Groth16 previously 713893, now 734527. Diff 20634

It doesn't seem to change PLONK recursion cost as there we don't use the method directly, but rather fold multiple and then do the pairing check.

This PR is required for Linea to support PointEval precompile for all inputs allowed in specification.

## Type of change

# How has this been tested?

- Added tests to handle edge cases

# Checklist:

### DIFF
--- a/std/commitments/kzg/verifier.go
+++ b/std/commitments/kzg/verifier.go
@@ -426,14 +426,14 @@ func (v *Verifier[FR, G1El, G2El, GTEl]) CheckOpeningProof(commitment Commitment
 
 	// [f(a)]G1 + [-a]([H(α)]G₁) = [f(a) - a*H(α)]G₁
 	pointNeg := v.scalarApi.Neg(&point)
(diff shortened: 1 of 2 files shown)


ivokub:
@yelhousni - now it seems that it works for the use case. But I'm not sure if I still shouldn't add separate option `kzg.WithCompleteArithmetic()` and do it properly for all methods. For now it seems that it works for the specific use case (CheckOpeningProof for constructed inputs), but I think the same issue would arise also in other methods when we really try construct invalid inputs (i.e. multi proof checking for all zeros etc).

What do you think?

PR #1570, opened 2025-08-06
docs(KNOWN_USERS): add zkBank project

Galoretka:
Add zkBank to KNOWN_USERS.md

zkBank is a capture-the-flag style zero-knowledge banking system built by the zkSecurity team.
It uses gnark to implement and verify GKR-based zero-knowledge proofs in Go.


ivokub:
Are you from zkSecurity? We usually accept contributions to KNOWN_USERS from the projects directly to ensure that they want to be endorsed. If not confirmed then I'll close the PR soon.

PR #1572, opened 2025-08-07
feat: pointeval precompile circuit for failing inputs 

ivokub:
# Description

This PR adds a circuit which allows to assert that some inputs to the POINTEVAL precompile input are invalid. It checks:
* compressed input mask
* mask correspondence with x coord
* subgroup and curve membership
* hash version
* hash value
* pairing value
* compressed x overflow

It will be used in Linea alongside the POINTEVAL precompile for valid inputs.

## Type of change

# How has this been tested?

Corresponding isolated test cases. Larger test suite in linea-monorepo.

# Checklist:


std/algebra/emulated/sw_bls12381/marshal.go
```diff
@@ -56,6 +61,19 @@ func (g1 *G1) UnmarshalCompressed(compressedPoint []uints.U8) (*G1Affine, error)
 	}
 	unmask := uints.NewU8(mMask)                   // unmaks = ^mask
 	prefix := uapi.And(unmask, compressedPoint[0]) // prefix = compressedPoint[0] & unmask
+	// lets ensure that we have a valid prefix (0b100, 0b101, 0b110)
+	prefixShifted := g1.api.Div(uapi.Value(prefix), 32) // prefixShifted = prefix >> 5
+	isValidPrefix := selector.Mux(g1.api, prefixShifted,
+		0, // 0b000 << 5
```
ThomasPiellard: should we keep the `<< 5` in the comment  ? At this stage, the selected values are on 3 bits (nitpick, it is clear from context)

std/algebra/emulated/sw_bls12381/marshal.go
```diff
@@ -84,26 +106,18 @@ func (g1 *G1) UnmarshalCompressed(compressedPoint []uints.U8) (*G1Affine, error)
...
-	// if p=O, we set P'=(0,0) and check equality, else we artificially set P'=P and check equality
-	isInfinity := g1.api.IsZero(g1.api.Sub(mCompressedInfinity, uapi.Value(prefix)))
-	zero := emulated.ValueOf[BaseField](0)
-	infX := g1.curveF.Select(isInfinity, &zero, x)
-	infY := g1.curveF.Select(isInfinity, &zero, y)
-	g1.curveF.AssertIsEqual(infX, x)
-	g1.curveF.AssertIsEqual(infY, y)
+	// if the mask is for zero, then the hint is supposed to have returned also y=0
+	isYZero := g1.curveF.IsZero(&res.Y)
+	isZeroPoint := g1.api.And(isXZero, isYZero)
```
ThomasPiellard: Couldn't we jut write `g1.api.AssertIsEqual(isYZero, isPrefixCompressedInfinity)` and skip `isZeroPoint := g1.api.And(isXZero, isYZero)` ?
-> if we remove it, the only difference is that it forbids the combo x!=0, y=0 if the mask is not `mCompressedInfinity`.

std/algebra/emulated/sw_bls12381/marshal.go
```diff
@@ -56,6 +61,19 @@ func (g1 *G1) UnmarshalCompressed(compressedPoint []uints.U8) (*G1Affine, error)
...
+	prefixShifted := g1.api.Div(uapi.Value(prefix), 32) // prefixShifted = prefix >> 5
+	isValidPrefix := selector.Mux(g1.api, prefixShifted,
+		0, // 0b000 << 5
+		0, // 0b001 << 5
+		0, // 0b010 << 5
+		0, // 0b011 << 5
+		1, // 0b100 << 5 - compressed regular point, with y lexicographically smallest
+		1, // 0b101 << 5 - compressed regular point, with y lexicographically largest
+		1, // 0b110 << 5 - compressed point at infinity
+		0, // 0b111 << 5
```
ThomasPiellard: same remark as ⬆️

std/evmprecompiles/10-kzg_point_evaluation.go
```diff
@@ -204,3 +207,359 @@ func KzgPointEvaluation(
...
+	copy(xCoordProofBytes[1:], proofSerialisedBytes[1:])
+	xCoordComBytes[0] = firstByteCom
+	xCoordProofBytes[0] = firstByteProof
+	// - here we convert from bytes to emulated without range checking. However,
+	// below in EmulatedToBytes we do not pass any option. In this case
+	// EmulatedToBytes will call [emulated.Field.ReduceStrict] on the input,
+	// which ensures that we convert the in-range emulated value to bytes. Thus,
+	// when the initial value xCoordComBytes/xCoordProofBytes is out of range,
+	// then the value after EmulatedToBytes will be different.
+	xCoordComEmul, err := conversion.BytesToEmulated[sw_bls12381.BaseField](api, xCoordComBytes[:], conversion.WithAllowOverflow())
```
ThomasPiellard, line 364: I don't understand, why do you do `conversion.BytesToEmulated` then `conversion.EmulatedToBytes` on the proof and commitment ?

std/evmprecompiles/10-kzg_point_evaluation.go
```diff
@@ -204,3 +207,359 @@ func KzgPointEvaluation(
 
 	return nil
 }
+
+// KzgPointEvaluationFailure checks a failing case of KZG point Evaluation
```
ThomasPiellard: could you add the comment in the PR with the list of checks that might fail ? it is helpful to follow the function ->
```
compressed input mask
mask correspondence with x coord
subgroup and curve membership
hash version
hash value
pairing value
compressed x overflow
```

std/algebra/emulated/sw_bls12381/marshal.go
ivokub: You're right, its mostly because of copy-pasting comments :)

std/evmprecompiles/10-kzg_point_evaluation.go
ivokub: Thanks, good point. Added documentation

std/evmprecompiles/10-kzg_point_evaluation.go
ivokub: It was a bit complicated, let me see if there is nice explanation.

So essentially what I want to check is that if x coordinate in the compressed representation overflows the field or not. By default `conversion.BytesToEmulated` asserts that the byte representation is within range (i.e. the input bytes value is strictly less than the modulus). But as we want to be able to prove incorrect inputs, then we cannot use this as this would lead to inability to compute a proof.

So, instead I'm using the option `conversion.WithAllowOverflow()` option to `conversion.BytesToEmulated` which does not perform the in-range assertion of the byte representation and just maps the byte representation into multi-precision integer representation (i.e. bytes `x` maps to non-native `x` and `x+fr` maps to `x+fr`). Now, we don't have a way to test `*emulated.Element` *IS* less than modulus (only assert), but this again means that we're unable to generate a proof for invalid inputs.

So, we convert the obtained non-native element back to bytes with `conversion.BytesToEmulated`. But that internally also calls `emulated.ReduceStrict` which performs strict modular reduction (i.e. we also assert the the reduced value is strictly less than modulus) before converting back to bytes, so `x -> x'` and `x + fr -> x'`. And now we can check byte-wise, if the initial byte representation of `x` is the same as `bytes(x) -> non-native(x) -> bytes(x)`.

We need all this because in non-native arithmetic by default we don't assert that values are strictly less than the modulus because it is expensive (bitwise comparison against the modulus), but we only check that the bitwidth of the non-native value is same as modulus, but this allows the hints to return `x` or `x+fr`. In arithmetic it is fine, but in serialization, where we need unique representation, it is not.

I also tried to explain this in the short comment in the implementation, but I guess it requires a lot of background knowledge how non-native works :)

std/algebra/emulated/sw_bls12381/marshal.go
ivokub: We can remove it, we have anyway above the assertion:
```go
isXZero := g1.curveF.IsZero(x)
g1.api.AssertIsEqual(isXZero, isPrefixCompressedInfinity)
```
so doing
```go
isYZero := g1.curveF.IsZero(&res.Y)
isZeroPoint := g1.api.And(isXZero, isYZero)
g1.api.AssertIsEqual(isZeroPoint, isPrefixCompressedInfinity)
```
is redundant to doing:
```go
isXZero := g1.curveF.IsZero(x)
g1.api.AssertIsEqual(isXZero, isPrefixCompressedInfinity)
// ...
isYZero := g1.curveF.IsZero(&res.Y)
g1.api.AssertIsEqual(isYZero, isPrefixCompressedInfinity)
```

ivokub (commented):
Thanks for the feedback! I made the corresponding changes.

std/evmprecompiles/10-kzg_point_evaluation.go
```diff
@@ -204,3 +205,382 @@ func KzgPointEvaluation(
...
+	isNotInSubgroups := api.Sub(1, isInSubgroups)
+	isNotCorrectHash := api.Sub(1, isCorrectHash)
+	isAnyPreviousFailure := api.Add(
+		isNotValidMasks,
+		isNotInRange,
+		isNotValidInfinityMask,
+		isNotInSubgroups,
+		isNotCorrectHash,
+	)
+	api.AssertIsBoolean(isAnyPreviousFailure)
```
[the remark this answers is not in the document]
ivokub: That is the goal actually. It sanity checks that we have correctly replaced with dummy values whenever we have encountered a failure.

PR #1573, opened 2025-08-08
Align out-of-circuit MiMC preimage with in-circuit limb hashing

MozirDmitriy:
- Write each 64-bit limb of BN254 Fp emulated inputs as a separate 32-byte BN254 Fr element before hashing with cmimc.
- Preserve limb order (least significant to most significant) to match EmulatedInput*.Limbs... in-circuit.
- Ensures the public hash computed off-circuit matches the in-circuit MiMC digest. Build succeeds for examples/inputpacking.

### DIFF
--- a/examples/inputpacking/inputpacking_test.go
+++ b/examples/inputpacking/inputpacking_test.go
@@ -145,15 +145,34 @@ func Example() {
 	input2.FillBytes(buf[:])
 	h.Write(buf[:])
 	emulatedInput1.FillBytes(buf2[:])
-	h.Write(buf2[24:32])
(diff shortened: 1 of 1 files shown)


ivokub:
Thanks for the contribution. Our MiMC implementation already left pads short inputs with zeros, so explicit left padding is not needed. See https://github.com/Consensys/gnark-crypto/blob/master/ecc/bn254/fr/mimc/mimc.go#L102-L110

However - I agree the current interface is not perfect. We're still trying to figure out a nice generic way to define algebraic hash functions which aligns better with the fact that we're actually hashing field element, not arbitrary bytes.

PR #1591, opened 2025-08-29
fix: add G1 membership for ML and FE in bls precompile

ivokub:
# Description

The Miller loop line computation already checks that second input is in G2, but we don't check it for first argument. Add explicit check.

## Type of change

# Checklist:

### DIFF
--- a/std/evmprecompiles/15-blspairing.go
+++ b/std/evmprecompiles/15-blspairing.go
@@ -94,6 +94,9 @@ func ECPairBLSMillerLoopAndMul(api frontend.API, accumulator *sw_bls12381.GTEl, 
 	if err != nil {
 		return fmt.Errorf("new pairing: %w", err)
 	}
+	// Miller loop below already checks that Q is in G2, but doesn't check that
+	// P is. So we do it manually.
+	pairing.AssertIsOnG1(P)
 	ml, err := pairing.MillerLoopAndMul(P, Q, accumulator)
 	if err != nil {
 		return fmt.Errorf("miller loop and mul: %w", err)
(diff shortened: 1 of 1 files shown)


std/evmprecompiles/15-blspairing.go
```diff
@@ -94,6 +94,9 @@ func ECPairBLSMillerLoopAndMul(api frontend.API, accumulator *sw_bls12381.GTEl,
 	if err != nil {
 		return fmt.Errorf("new pairing: %w", err)
 	}
+	// Miller loop below already checks that Q is in G2, but doesn't check that
+	// P is. So we do it manually.
+	pairing.AssertIsOnG1(P)
```
YaoJGalteland, line 99: I noticed here doesn't check for P: https://github.com/Consensys/gnark/blob/master/std/algebra/emulated/sw_bls12381/pairing.go#L88
would it be good to add the check, or is it not necessary in the Pair func?

ivokub: The method `Pair` in https://github.com/Consensys/gnark/blob/master/std/algebra/emulated/sw_bls12381/pairing.go#L88 is a primitive function which can be used elsewhere to implement larger protocols. We have usually omitted subgroup checks by default there as there are many cases where the inputs come already trusted to be in subgroup. And our subgroup checks are not idempotent - even if subgroup checks are already done, then we perform them every time we call subgroup check, adding a lot of constraints.

But here the `ECPairBLSMillerLoopAndMul` method is already high-level method which is directly called on Linea side where we know the subgroup checks are not done, so I added it here.

YaoJGalteland (approved):

PR #1596, opened 2025-09-07
sha2: fix invalid slice-to-array conversion in test

Galoretka:
# Description

Replace direct conversion [32]uints.U8(uints.NewU8Array(...)) with copying into a predeclared array. Go does not allow converting a slice []T to an array [N]T; copying is required. This aligns with the pattern already used in TestSHA2 and fixes a compile-time error in TestSHA2FixedLengthSum.

## Type of change

# How has this been tested?

go test ./std/hash/sha2 -run TestSHA2FixedLengthSum -v
and
go test ./std/hash/sha2 -v


ivokub (commented):
Conversion from slice to array is valid. See https://go.dev/ref/spec#Conversions_from_slice_to_array_or_array_pointer

PR #1600, opened 2025-09-09
fix: panic on zero-length inputs in Mux

MozirDmitriy:
# Description

Add explicit check in Mux to panic when inputs length is zero.
Prevents undefined behavior: previously n==0 led to bits.Len(n-1) on MaxUint and slicing inputs[:leftCount], causing non-obvious runtime panics and potential big.Int overflow in comparator bound construction.
Aligns with package style where invalid shape/size inputs trigger panic, while out-of-range selectors cause proof failure.

## Type of change

# How has this been tested?

go test -v ./std/selector
go test -v -run "Mux" ./std/selector
go test -v ./...

# Checklist:


ivokub (approved):
Thanks. LGTM.

I merge it with the branch where we collect different small contributions and then do a single merge to master to avoid excessive CI runs.

PR #1613, opened 2025-09-24
Make global logger concurrency-safe using atomic.Value

viktorking7:
### PR Description
- **Summary**: Replace the global logger variable with an atomic.Value to prevent data races under concurrent access.
- **Why**: `logger` was read/written without synchronization, leading to potential data races when `Set`, `SetOutput`, `Disable`, and `Logger` are called concurrently.
- **How**: Store `zerolog.Logger` in `atomic.Value`; update `Set`, `SetOutput`, `Disable`, and `Logger` to use `Load`/`Store`. API remains unchanged.
- **Impact**: No breaking changes; negligible overhead; eliminates race conditions in logging.
- **Verification**: Builds successfully with `go build ./...`. Optional: run `go test -race` to validate no races in practice.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -7,39 +7,42 @@ import (
 	"io"
 	"os"
 	"strings"
+	"sync/atomic"
 
 	"github.com/consensys/gnark/debug"
 	"github.com/rs/zerolog"
 )
 
-var logger zerolog.Logger
+var globalLogger atomic.Value // stores zerolog.Logger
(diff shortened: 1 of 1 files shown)


ivokub:
We're probably going at some point to reimplement logger using `log/slog` for reducing dependencies and make it internal. Usually the internal packages should initialize the loggers in `init()` functions which are called sequentially.

PR #1622, opened 2025-09-30
fix: handle file close error in profile.Stop()

DeVikingMark:
Previously, the f.Close() call was ignoring potential errors, which could lead to resource leaks and data corruption on some filesystems. 

This change adds proper error handling for file closure to ensure:

1. File descriptor leaks are prevented
2. Data integrity is maintained on filesystems that require successful
   close operations for data persistence
3. Error logging consistency with other file operations in the method


ivokub:
Duplicate.

PR #1625, opened 2025-10-06
feat: add BLS12-377, BLS12-381 and BW6-761 GPU acceleration support for Groth16

ivokub:
# Description

The ICICILE backend already has the support for the curves, but we never exposed them before. I added the support without code generation for now, which can be done in the future.

I also refactored the package paths not to contaminate the "native" accelerators with different external accelerators. This now allows to provide some accelerator-specific options. I kept the existing options for backwards compatibility, but it now returns runtime error with helpful message on how to upgrade the implementation.

Tested on AWS g6e.2xlarge and works for all curves.

I also clarified documentation and added examples.

## Type of change

# How has this been tested?

- Example for using the accelerated backend.

Very brief benchmarks show 20-40x speedup over the native prover on CPU (only prover runtime excluding solver runtime).

# Checklist:


backend/accelerated/icicle/doc.go
```diff
@@ -0,0 +1,65 @@
...
+//   - BLS12-381
+//   - BN254
+//   - BW6-761
+//
+// # Setup
+//
+// To initialize the ICICLE backend, follow the instructions in the [ICICLE]
+// repository. Namely, first you should install the ICICLE library:
+//
+//	git clone github.com/ingonyama-zk/icicle-gnark
```
gbotrel: --> missing https:// in front of repo for quick copy paste clone

gbotrel: might be worth to say that it works with cuda 13.0 and needs clang installed

backend/accelerated/icicle/groth16/bls12-381/icicle.go
```diff
@@ -0,0 +1,726 @@
+//go:build icicle
+
+package bls12381
```
gbotrel, line 8: shouldn't we code generate these files ?

gbotrel (dismissed):
looks good to me, tested with Cuda 13.0. 
Just reviewed one of the curve, may be worth it to create an issue to refactor that behind a template for maintenance

backend/accelerated/icicle/doc.go
ivokub: Good points! Will add

backend/accelerated/icicle/groth16/bls12-381/icicle.go
ivokub: Indeed it would be better for maintainability indeed. I was thinking at first I'd do proper code generation when I would move the accelerated backends to separate repo to avoid importing huge third-party dependencies, but imo this is still not planned yet. And I guess accelerated PLONK could come sooner, so it would be actually beneficial to have the code generation templates ready. Particularly that for BW6 we use different extension for G2 than for other curves, so it would probably simplify implementing.

ivokub:
> looks good to me, tested with Cuda 13.0. Just reviewed one of the curve, may be worth it to create an issue to refactor that behind a template for maintenance

Thanks for the review. I'll do proper code generation before merge, I'll try not to take shortcuts :). We already have the PR, so any external parties wanting to use GPU acceleration for other curves could use it directly.

backend/accelerated/icicle/doc.go
ivokub: Added:
* clang requirement
* cuda requirement (with link)
* cuda version reference

backend/accelerated/icicle/groth16/bls12-381/icicle.go
ivokub: Added code generation

PR #1634, opened 2025-10-27
feat: `SumMerkleDamgardDynamicLength`

Tabaie:
A simple utility function that computes the hash of a dynamically sized slice of variables. Used in Linea.

---

> [!NOTE]
> Introduces variable-length hashing support and updates tests accordingly.
> 
> - Add `DynamicLengthFieldHasher` with `SumWithLength(l)`; implement in Merkle–Damgård hasher by tracking per-prefix `state` and using `logderivlookup` for indexed lookup
> - Refactor Merkle–Damgård state handling: `state` becomes a slice, `Reset` trims to IV, `Write` appends compressed states, `Sum`/`State` return the latest state, `SetState` enforces reset-before-set
> - Expand Poseidon2 tests to validate per-prefix hashes across the standard and GKR implementations and to check state store/restore behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 92b24a382102787a0eb8389bca9791bae7873111. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


std/hash/hash.go
```diff
@@ -136,3 +140,24 @@ func (h *merkleDamgardHasher) Write(data ...frontend.Variable) {
...
+
+// SumMerkleDamgardDynamicLength computes the Merkle-Damgård hash of the input data, truncated at the given length.
+// Parameters:
+//   - api: constraint builder
+//   - f: 2-1 hash compression (one-way) function
+//   - initialState: the initialization vector (IV) in the Merkle-Damgård chain. It must be a value whose preimage is not known.
+//   - length: length of the prefix of data to be hashed. The verifier will not accept a value outside the range {0, 1, ..., len(data)}.
+//     The gnark prover will refuse to attempt to generate such an unsuccessful proof.
+//   - data: the values a prefix of which is to be hashed.
+func SumMerkleDamgardDynamicLength(api frontend.API, f Compressor, initialState frontend.Variable, length frontend.Variable, data []frontend.Variable) frontend.Variable {
```
ivokub: Doesn't this function signature allow for length extension attacks? I.e. lets say:
* IV = 0
* f_0 = compress(IV, msg_0)
* f_1 = compress(f_0, msg_1)

Then, when we do either:
`SumMerkleDamgardDynamicLength(api, Compressor, IV, 2, [msg_0, msg_1])`
or
`SumMerkleDamgardDynamicLength(api, Compressor, f_0, 1, [msg_1, msg_xxx])`
we would get the same result? And considering all are user inputs then it could imo lead to possible collisions.

We also provide variable-length mode for binary hashes (see interface BinaryFixedLengthHasher). But there depending on the underlying hash function (sha256, ripemd, sha3) the length is appended to the input, avoid collision problems.

std/hash/hash.go
```diff
@@ -136,3 +140,24 @@ func (h *merkleDamgardHasher) Write(data ...frontend.Variable) {
...
+	resT := logderivlookup.New(api)
```
ivokub: And imo this method diverges from the rest of the APIs we use for hashers - otherwise we have defined `FieldHasher` for algebraic hash, and `BinaryHasher`/`BinaryFixedLengthHasher` for binary hashes (the `..FixedLengthHasher` providing the same functionality i.e. allowing to dynamically set the input length to be hash).

I recommend following similar patter for consistency -- create `FieldFixedLengthHasher` interface
```go
type FieldFixedLengthHasher interface {
    FieldHasher

    FixedLengthSum(length frontend.Variable) frontend.Variable
}
```
and then make in MiMC and Poseidon2 packages return this instead of `FieldHasher`.

ivokub (changes requested):
Reviewed - I'm not sure about the interfaces right now. I think adding as a general purpose function is unsafe as allows to easily create collisions. I think leaving it up to user implement correctly will cause issues. To overcome, we should do standard method of adding the input width as a last block, but this doubles the constraint count imo.

Secondly - I'd follow current interfaces pattern. All other hasher types are defined as interfaces, but this is a single function. For me it is a bit confusing.

Or - if only used in Linea, then I'd instead add the method there directly as we can have better control already within business logic which would avoid possible collisons.

std/hash/hash.go
Tabaie: I wouldn't say that's a collision, since Merkle-Damgard chains initialized with different IV's are considered different hash functions, and you can always design another hash function H' that could at some point produce the same output as a given hash function H.

As to length extension attacks, yes Merkle-Damgard is generally prone to length extension attacks, afaik that attack is not in scope in our stack since we're not privacy focused. (the "attacker" in length extension obtains a valid hash for a longer string when they weren't supposed to know what the preimage of the original hash was)

Your proposed solution for preventing length extension attacks is valid imo (or we could just use Sponge instead of MD for range extension) but I wanted something that would match the current MiMC and Poseidon2 implementations that we have. i.e. `MiMC(a,b) = DynamicLengthMiMC(2, a, b, c)`.

std/hash/hash.go
Tabaie: Ah you're suggesting to take the stateful hasher, `Write` to it normally, and in the end instead of the regular `Sum` do some other `FixedLengthSum` function (wouldn't `DynamicLengthSum` be a better name?)
That does seem much cleaner.

std/hash/hash.go
ivokub: Yes, MD hashes with different IVs are essentially different hash functions (we can consider it as keyed hash function). But usually this means that IV (or the first message to MD) is fixed and constant. But in this implementation it is provided as an input, so for me it completely changes the model.

Imo we didn't consider length extension attack previously as all the hash functions in-circuit return the hash of all written variables. So it is not possible to mount the length-extension attack. But now, when the length is also provided as a variable, so the length-extension attacks at least for me return to scope.

It may be that any of these attacks are not applicable in the specific use case you need, but imo it is also very easy to use it in a way allowing length extension attacks and exposing something in `std/hash/` should have safe-by-default design imo.

std/hash/hash.go
ivokub: Exactly, thats the idea. We already have the exact same interface for binary hashes (sha2, sha3) (see https://pkg.go.dev/github.com/consensys/gnark@v0.14.0/std/hash#BinaryFixedLengthHasher) and imo it works quite well.

Perhaps indeed `DynamicLengthSum` could be better, but at least for me consistency is more important. If one interface defines as `FixedLengthSum` and other as `DynamicLengthSum`, but their are essentially same thing, then it just confuses the users.

std/hash/hash.go
```diff
@@ -26,6 +27,13 @@ type FieldHasher interface {
 	Reset()
 }
 
+// DynamicLengthFieldHasher can compute hashes of lengths unknown at compile time.
+type DynamicLengthFieldHasher interface {
```
ivokub, line 31: I would keep consistency with `BinaryFixedLengthHasher`. Also see the interface description there which imo is more thorough.

std/hash/hash.go
```diff
@@ -108,48 +116,68 @@ type Compressor interface {
 }
 
 type merkleDamgardHasher struct {
-	state frontend.Variable
-	iv    frontend.Variable
-	f     Compressor
-	api   frontend.API
+	state         []frontend.Variable  // state after being updated with each written element
+	stateTable    logderivlookup.Table // stateTable always contains a prefix of h.state
+	stateTableLen int
```
ivokub, line 121: Please comment what is `stateTableLen`: "// stateTableLen indicates the length of of `state` entries written to `stateTable`. It will be updated with every read to `DynamicSum()`" etc.

std/hash/poseidon2/poseidon2_test.go
```diff
@@ -1,51 +1,81 @@
...
+
+	varlen.Write(c.Input...)
+
+	for i := range c.Input {
+		hsh.Write(c.Input[i])
+		api.AssertIsEqual(c.Expected[i], hsh.Sum())
+		gkr.Write(c.Input[i])
+		api.AssertIsEqual(c.Expected[i], gkr.Sum())
+		api.AssertIsEqual(c.Expected[i], varlen.SumWithLength(i+1))
+	}
```
ivokub, line 50: I'l also add test if you're trying to read hash digest not from the end, but from earlier. A la `varlen.SumWithLength(2)` in the end.

std/hash/poseidon2/poseidon2_test.go
```diff
@@ -1,51 +1,81 @@
...
+		&poseidon2Circuit{
+			Input:    make([]frontend.Variable, nbInputs),
+			Expected: make([]frontend.Variable, nbInputs),
+		}, test.WithValidAssignment(&poseidon2Circuit{
+			Input:    input,
+			Expected: expected,
+		}), test.WithCurves(ecc.BLS12_377))
 }
 
 func TestStateStorer(t *testing.T) {
```
ivokub, line 81: I would also add SumWithLen test to the state storer circuit.

std/hash/hash.go
```diff
@@ -108,48 +116,68 @@ type Compressor interface {
...
+	api           frontend.API
 }
 
-// NewMerkleDamgardHasher transforms a 2-1 one-way function into a hash
-// initialState is a value whose preimage is not known
+// NewMerkleDamgardHasher range-extends a 2-1 one-way hash compression function into a hash by way of the Merkle-Damgård construction.
+// Parameters:
+//   - api: constraint builder
+//   - f: 2-1 hash compression (one-way) function
+//   - initialState: the initialization vector (IV) in the Merkle-Damgård chain. It must be a value whose preimage is not known.
```
ivokub, line 130: Please add comment that it is not safe from length extension attacks. Indicate that the users can manually write `len(input)` to the input if they want the safety.

std/hash/hash.go
```diff
@@ -108,48 +116,68 @@ type Compressor interface {
...
 func NewMerkleDamgardHasher(api frontend.API, f Compressor, initialState frontend.Variable) StateStorer {
```
ivokub, line 131: You can potentially also take options, i.e. see `WithMinimalLength`, so that you wouldn't have to construct the full table for the lookups. But its not necessary for now, you can always add it later (options is vararg, so should be transparent to caller).

std/hash/hash.go
ivokub: I re-reviewed your implementation. I think the attack becomes less obvious now. If we don't use `State/SetState` and the IV provided to `NewMerkleDamgardHasher` is well defined, then imo its quite good. There is still the issue that we don't write input length into the stream, but I think after documenting it nicely it is good enough.

std/hash/hash.go
Tabaie: So to call it "fixed length"? That seems like the exact opposite of what it does.

ivokub: We fixed the input length at solving time? Dunno, perhaps it could be better, but imo it also is not wrong.

My point is that we already have an existing interface which does the same thing (but when we get the inputs as bytes instead of field elements), so for me it is logical that the new interface has similar naming. And for boolean hashes the "dynamic length" is also confusing as it could refer to the output length of the hash (as in SHAKE). What is important is that the methods are well documented and explained, we shouldn't rely on the method name only to pass on all information.

PR #1635, opened 2025-10-28
fix: enforce strict < modulus for native byte conversions

GarmashAlex:
Enforce documented strict inequality in BytesToNative() and () by introducing assertBytesLt and using it when WithAllowOverflow is not set.
Prevents accepting byte arrays equal to the field modulus, which previously allowed a non-unique representation of zero and could be exploited by malicious hints.Adds a negative test for the “bytes == modulus” case in BytesToNative().
Emulated conversions already enforced strictness; bringing native conversions into alignment with docs and intent.

---

> [!NOTE]
> Enforces strict less-than modulus in BytesToNative/NativeToBytes by extending assertBytesLeq with a disallowEquality flag and adds tests, while preserving overflow-allowed behavior.
> 
> - **Conversion**:
>   - `BytesToNative` and `NativeToBytes` now enforce strict `< modulus` by calling `assertBytesLeq(api, ..., field, true)`.
>   - Extend `assertBytesLeq` to `assertBytesLeq(api, b, bound, disallowEquality)` to optionally forbid equality.
> - **Tests**:
>   - Add negative/positive cases where bytes equal the modulus (invalid without overflow; valid with `WithAllowOverflow`).
>   - Update `AssertBytesLeq` circuit and test matrix to pass `disallowEquality` and validate both strict and non-strict comparisons.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ba04639775e26a2cb1f7856ca9837058ddcb9a54. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ivokub (approved):
Thanks for the PR! I refactored to have a single implementation with a switch and added a few tests.

PR #1661, opened 2025-12-10
feat: Enable multiple GKR sub-circuits

Tabaie:
Note: deserialization tests are failing due to a change in the structure of constraint systems. Rather than a single "gkr info" sub-object, we now have a slice of them.

---

> [!NOTE]
> Enables multiple GKR sub-circuits per constraint system and modernizes the GKR pipeline (hints, registry, and solving/proving) with clearer APIs and per-curve support.
> 
> - Replace single `GkrInfo` with `[]*gkrinfo.StoringInfo`; add `System.NewGkr()` returning a `*StoringInfo` and sub-circuit index; update all systems and builders accordingly
> - Solver: use `gkrtypes.NewSolvingInfo` and per-curve `NewSolvingData`; override hints via `gkrhints.TestEngineHints` (`GetAssignment`, `Solve`, `Prove`) instead of stored IDs; pad instances to power-of-two and adjust hint I/O formats
> - GKR hints rework: introduce `GetAssignmentHint`, refactor `SolveHint`/`ProveHint` to accept multiple circuits, remove worker pool, add `repeatUntilEnd`, and derive Fiat–Shamir challenges with canonical byte parsing
> - Gate registry: new error-validating `Register` options, curve-allowlist support, function equality checks, and stricter duplicate/name/degree handling; expose `Get` and refactor verifier API
> - gkrtypes: add `NewCircuit`/`NewSolvingInfo`, gate `SupportsCurve`, and per-curve defaults; remove dependency/print machinery; tweak circuit metadata (`setNbUniqueOutputs`)
> - Misc: export `gkr.ComputeLogNbInstances`, fix `MulAcc` to multiply, add utils (`ExtendRepeatLast`, slice refs), update generator templates and tests
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9227b0c1a4ebf5af6bfa632074fb7dd3625f70d4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


constraint/bls12-377/solver.go
```diff
@@ -49,16 +50,22 @@ type solver struct {
...
 			return nil, err
 		}
-		opts = append(opts,
-			csolver.OverrideHint(cs.GkrInfo.SolveHintID, gkr.SolveHint(solvingInfo, &gkrData)),
-			csolver.OverrideHint(cs.GkrInfo.ProveHintID, gkr.ProveHint(cs.GkrInfo.HashName, &gkrData)))
+
+		gkrData := gkr.NewSolvingData(solvingInfo)
+		var gkrHints *gkrhints.TestEngineHints
+
+		newOpts := make([]csolver.Option, len(opts), len(opts)+3)
```
ivokub: What is the point of allocating manually and then copying? Do you want to keep the initial `opts` argument intact?

If you just remove the manual allocation and copy, and only keep `opts = append(opts, ...)` then it achieves the same thing as `append` will allocate sufficient space and copy to fit everything.

constraint/bls12-377/solver.go
```diff
@@ -49,16 +50,22 @@ type solver struct {
...
+		copy(newOpts, opts)
+		opts = append(newOpts,
+			csolver.OverrideHint(csolver.GetHintID(gkrHints.GetAssignment), gkr.GetAssignmentHint(gkrData)),
```
ivokub, line 70: I don't understhand, why do we call `csolver.GetHintID(gkrHints.Solve)`? You have just initialized `gkrHints` as a nil pointer -- so you use a method `(*TestEngineHints)(nil).GetAssignment`.

Instead we could just have a placeholder hint somewhere `PlaceHolderGetAssignment` at package level and use this. This is also how we do for `API.Commit`.

constraint/bls12-377/solver.go
```diff
@@ -49,16 +50,22 @@ type solver struct {
...
 		solvingInfo, err := gkrtypes.StoringToSolvingInfo(cs.GkrInfo, gkrgates.Get)
 		if err != nil {
 			return nil, err
 		}
-		opts = append(opts,
-			csolver.OverrideHint(cs.GkrInfo.SolveHintID, gkr.SolveHint(solvingInfo, &gkrData)),
-			csolver.OverrideHint(cs.GkrInfo.ProveHintID, gkr.ProveHint(cs.GkrInfo.HashName, &gkrData)))
+
+		gkrData := gkr.NewSolvingData(solvingInfo)
+		var gkrHints *gkrhints.TestEngineHints
```
ivokub, line 66: Why do we use `TestEnghineHints`? We run the actual solver here already?

constraint/solver/gkrgates/registry.go
```diff
@@ -35,78 +37,155 @@ type registerSettings struct {
...
 	}
 }
 
 // WithCurves determines which curves the gate is validated on.
 // The default is to validate on BN254.
 // This works for most gates, unless the leading coefficient is divided by
 // the curve's order, in which case the degree will be computed incorrectly.
-func WithCurves(curves ...ecc.ID) registerOption {
-	return func(settings *registerSettings) {
+func WithCurves(curves ...ecc.ID) RegisterOption {
```
ivokub, line 130: The documentation does not match the implementation. Now in the registration you set `allowedCurves` based on the option.

For me this is complex - I think the gate should know itself where it can work (if it doesn't work over all fields) and then panic.

And I think we can still use the approach where we test the degree of the gate on a random modulus instead of letting the users define it themselves. Otherwise the user has to know what is the relation between gate degree and the provided modulus, which doesn't make sense as it is used for automatic degree testing.

constraint/solver/gkrgates/registry.go
```diff
@@ -35,78 +37,155 @@ type registerSettings struct {
...
+	gatesLock.Lock()
+	defer gatesLock.Unlock()
+
+	if g, ok := gates[s.name]; ok {
+		// gate already registered
+		if g.NbIn() != nbIn {
+			return fmt.Errorf("gate \"%s\" already registered with a different number of inputs (%d != %d)", s.name, g.NbIn(), nbIn)
+		}
+
+		for _, curve := range curvesForTesting {
```
ivokub, line 174: I would remove the automatic gate degree testing. For me the implementation becomes more complex and actually more fragile to use (the user could mess up options and have unverfied wrong degree etc). I'd keep this PR only for supporting multiple GKR instances.

constraint/solver/gkrgates/registry.go
```diff
@@ -35,78 +37,155 @@ type registerSettings struct {
...
+	if len(curvesForTesting) == 0 {
+		// no restriction on curves, but only test on BN254
+		curvesForTesting = []ecc.ID{ecc.BN254}
+		allowedCurves = gnark.Curves()
+	}
+
+	gatesLock.Lock()
+	defer gatesLock.Unlock()
+
+	if g, ok := gates[s.name]; ok {
```
ivokub, line 168: I would keep the duplicate checkins simple -- if there is same duplicate (key, gate) then it is a no-op. If there is same key but different gate then it is not.

It becomes more complicated when the gate functions are anonymous. But imo this means we should make the gate functions simpler and non-anonymous. It also significantly helps reviewing and auditing as we don't have to find the actual defined gates at runtime and instead can analyze them statically.

constraint/solver/gkrgates/registry.go
```diff
@@ -221,15 +310,17 @@ func (v *gateVerifier) findSolvableVar(f gkr.GateFunction, nbIn int) int {
...
 // nbIn is the number of inputs to the gate.
 func (v *gateVerifier) isVarSolvable(f gkr.GateFunction, claimedSolvableVar, nbIn int) bool {
 	return v.isAdditive(f, claimedSolvableVar, nbIn)
 }
 
-func (v *gateVerifier) VerifyDegree(g *gkrtypes.Gate) error {
-	if err := v.verifyDegree(g.Evaluate, g.Degree(), g.NbIn()); err != nil {
+// verifyDegree checks that the declared total degree of the gate polynomial
+// is correct.
+func (v *gateVerifier) verifyDegree(g *gkrtypes.Gate) error {
```
ivokub: unused method?

constraint/solver/gkrgates/registry.go
```diff
@@ -239,7 +330,9 @@ func (v *gateVerifier) VerifyDegree(g *gkrtypes.Gate) error {
 	return nil
 }
 
-func (v *gateVerifier) VerifySolvability(g *gkrtypes.Gate) error {
+// verifySolvability checks that the variable declared as "solvable"
+// in fact occurs with degree exactly 1.
+func (v *gateVerifier) verifySolvability(g *gkrtypes.Gate) error {
```
ivokub: unused method?

internal/gkr/gkrtypes/types.go
```diff
@@ -254,6 +224,7 @@ func CircuitInfoToCircuit(info gkrinfo.Circuit, gateGetter func(name gkr.GateNam
 	resCircuit := make(Circuit, len(info))
 	for i := range info {
 		if info[i].Gate == "" && len(info[i].Inputs) == 0 {
+			resCircuit[i].Gate = Identity() // input wire
 			continue
 		}
 		resCircuit[i].Inputs = info[i].Inputs
```
ivokub: Below we use `gateGetter` is there a case where we use some different registry? Why not hardcode the `gkrgates.Get`?

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,162 @@ import (
...
 
-func (d *SolvingData) init(info gkrtypes.SolvingInfo) {
-	d.workers = utils.NewWorkerPool()
-	d.circuit = info.Circuit
-	d.circuit.SetNbUniqueOutputs()
+type newSolvingDataSettings struct {
+	assignment gkrtypes.WireAssignment
+}
+
+type NewSolvingDataOption func(*newSolvingDataSettings)
```
ivokub, line 34: Please add documentation. A la if no option is set then what are default values etc.

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,162 @@ import (
...
-				start := startInChunk + chunkOffset
-				end := endInChunk + chunkOffset
-				inputs := make([]frontend.Variable, maxNIn)
-				dependencyHeads := make([]int, len(data.circuit)) // for each wire, which of its dependencies we would look at next
-				for wI := range data.circuit {                    // skip instances that are not relevant (related to instances before the current task)
-					deps := info.Dependencies[wI]
-					dependencyHeads[wI] = algo_utils.BinarySearchFunc(func(i int) int {
-						return deps[i].InputInstance
-					}, len(deps), start)
+		if s.assignment != nil {
```
ivokub: I'm not fully following -- even if we have different GKR circuits then we set the same assignments for all? What is the use case for that?

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,162 @@ import (
...
+
+		data.assignment.repeatUntilEnd(data.nbInstances)
+
+		// The second input is dummy, just to ensure the solver's work is done before the prover is called.
+		// The rest constitute the initial fiat shamir challenge
+		insBytes := algo_utils.Map(ins[1:], func(i *big.Int) []byte {
+
 			b := make([]byte, fr.Bytes)
 			i.FillBytes(b)
 			return b[:]
```
ivokub: You don't need to reslice, `b` is already a slice.

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,162 @@ import (
...
+		if !ins[0].IsUint64() {
+			return fmt.Errorf("first input to GKR prove hint must be the sub-circuit index")
+		}
+		data := data[ins[0].Uint64()]
+		hashName := data.hashName
+		ins = ins[1:]
+
+		data.assignment.repeatUntilEnd(data.nbInstances)
+
+		// The second input is dummy, just to ensure the solver's work is done before the prover is called.
```
ivokub: I don't understand the comment. We are not doing any enforcement on `ins[1]` so that it is set? So how it works (my understanding):
* we call res := api.NewHint(SolveHint, circID, ...)
* we call res := api.NewHint(ProveHint, circID, res[1], .....)

Because now calling `ProveHint` depends on the outputs of calling `SolveHint`, then the gnark solver automatically orders the calling of hints so that gkr Prove is called after gkr Solve?

I get it and it works, but it most definitely needs to be written down. It really is implicit way to enforce ordering of hints. People don't understand it without knowing how gnark really works.

internal/gkr/bls12-377/solver_hints.go
```diff
@@ -21,121 +21,162 @@ import (
...
+		// The rest constitute the initial fiat shamir challenge
+		insBytes := algo_utils.Map(ins[1:], func(i *big.Int) []byte {
```
ivokub: Why is writing it like this better than:
```
insBytes := make([][]byte, len(ins[1:]))
for i := range insBytes {
    insBytes[i] = make([]byte, fr.Bytes)
    ins[1+i].FillBytes(insBytes[i])
}
```

internal/gkr/gkrhints/engine_hints.go
```diff
@@ -0,0 +1,196 @@
...
+
+func (g gateAPI) Add(i1, i2 frontend.Variable, in ...frontend.Variable) frontend.Variable {
+	in1 := utils.FromInterface(i1)
+	in2 := utils.FromInterface(i2)
+
+	in1.Add(&in1, &in2)
+	for _, v := range in {
+		inV := utils.FromInterface(v)
+		in1.Add(&in1, &inV)
+	}
```
ivokub, line 144: Why no reduce? (also everywhere below)

internal/gkr/gkrhints/engine_hints.go
```diff
@@ -0,0 +1,196 @@
...
+	bls24317 "github.com/consensys/gnark/internal/gkr/bls24-317"
+	bn254 "github.com/consensys/gnark/internal/gkr/bn254"
+	bw6633 "github.com/consensys/gnark/internal/gkr/bw6-633"
+	bw6761 "github.com/consensys/gnark/internal/gkr/bw6-761"
+	"github.com/consensys/gnark/internal/gkr/gkrinfo"
+	"github.com/consensys/gnark/internal/gkr/gkrtypes"
+	"github.com/consensys/gnark/internal/utils"
+)
+
+type TestEngineHints struct {
```
ivokub, line 23: It is called `TestEngineHints`, but it is referenced during actual SNARK circuit solving?

std/gkrapi/compile.go
```diff
@@ -5,236 +5,282 @@ import (
...
 	circuit     gkrtypes.Circuit
 	assignments gkrtypes.WireAssignment
 }
 
-type Solution struct {
-	toStore      gkrinfo.StoringInfo
-	assignments  gkrtypes.WireAssignment
-	parentApi    frontend.API
-	permutations gkrinfo.Permutations
+type InitialChallengeGetter func() []frontend.Variable
```
ivokub, line 29: How many times can it be called? Should it return the same challenges every time or new? Needs documentation!

std/gkrapi/compile.go
```diff
@@ -5,236 +5,282 @@ import (
...
+	hints                *gkrhints.TestEngineHints // hints for the GKR circuit, used for testing purposes
+	index                int                       // index among all GKR circuits
 }
 
-func (api *API) nbInstances() int {
-	if len(api.assignments) == 0 {
-		return -1
+// New creates a new GKR API
+func New(api frontend.API) *API {
+	toStore, index := api.(gkrinfo.ConstraintSystem).NewGkr()
```
ivokub: I would handle the error better - right now if the API doesn't implement it we just get "doesn't implement the interface" panic, but it doesn't allow user to fix the issue (perhaps they wrapped their builder for adding functionality etc)

std/gkrapi/compile.go
```diff
@@ -5,236 +5,282 @@ import (
...
-			appendNonNil(&ins, api.assignments[i])
+	hintIn := make([]frontend.Variable, 2+len(c.ins)) // first and second input denote the circuit and instance, respectively.
+	hintIn[0] = c.index
+	hintIn[1] = c.toStore.NbInstances
+	for hintInI, wI := range c.ins {
+		if inV, ok := input[wI]; !ok {
+			return nil, fmt.Errorf("missing entry for input variable %d", wI)
+		} else {
+			hintIn[hintInI+2] = inV
+			c.assignments[wI] = append(c.assignments[wI], inV)
```
ivokub, line 141: Could we potentially have a perf regression here? If we add instances one-by-one then we append there the slices also one by one. If we have 50k instances, then it just adds up time- and allocation wise.

Thats why it would be nice to keep the previous approach as well -- then we know beforehand how big are the assignemnts and allocate only once. And we could also do solving only once?

std/gkrapi/compile.go
```diff
@@ -5,236 +5,282 @@ import (
...
-	for i := range circuit {
-		if circuit[i].IsOutput() {
-			api.assignments[i] = outsSerialized[:nbInstances]
-			outsSerialized = outsSerialized[nbInstances:]
+	return res, nil
+}
+
+// finalize encodes the verification circuitry for the GKR circuit.
+func (c *Circuit) finalize(api frontend.API) error {
+	if api != c.api {
```
ivokub: I think this check is not particularly necessary - imo there may be legit reasons for API to be different. I.e. it was wrapped to add (but even more, to *remove* functionality e.g. ability to call Commit).

ivokub (changes requested):
I still think the automated gate degree testing during registration makes the implementation a bit too complex. Seems ok implementation wise, but I'm not fully confident.

And imo restricting the gates being only used when compiled on particular curves adds also unnnecessary complexity (and most of the times we provide `gnark.Curves()` everywhere). For example in Poseidon2 package we correctly error out if someone wants to call the circuit on mismatching curve, so the assertion is already done.

For me gate registration is still way too verbose. Instead of 
```
gkrgates.Register(extKeyGate(&p.RoundKeys[round][varIndex]), 2, gkrgates.WithUnverifiedDegree(1), gkrgates.WithUnverifiedSolvableVar(0), gkrgates.WithName(gateNames.linear(varIndex, round)), gkrgates.WithCurves(ecc.BLS12_377))
```
imo it makes so much more sense if I could do
```
gkrgates.Register(xtKeyGate(&p.RoundKeys[round][varIndex]), 2 /*inputs*/, 1 /*degree*/, "NAME-R1-V2")
```

And the one-by-one instance addition. Makes sense, but have you checked the profiles how does it affect circuit compile/solve time? I can imagine when doing one by one there is overhead?


I can try working on the PR tomorrow to wrap it finally up.

internal/gkr/gkrtypes/types.go
ivokub: Oh, import cycles. I think it would be better to have less packages so that we could actually refer to the gate registry.

constraint/bls12-377/solver.go
Tabaie: Yes for an extra safety guarantee. For example if the caller gives a sub-slice and we append to it, it could overwrite the super-slice if I'm not mistaken. It's unlikely for anyone to do that but it costs so little to prevent it.

constraint/bls12-377/solver.go
Tabaie: That hint is not just a placeholder. It is the actual hint for the test engine. Because it computes (at the solving stage) and uses (at the proving stage) the full circuit assignments in the background, so it needs to be tied to an object for those to persist.
But the empty object is only used to get references to those functions.

constraint/bls12-377/solver.go
Tabaie: We're not actually using the test engine hints. This object is used to get references to the test engine hints for them to be replaced.

constraint/solver/gkrgates/registry.go
Tabaie: I'm open to removing the unverified degrees. We can mostly rely on automatic testing and verified user hints (useful for debugging.)

constraint/solver/gkrgates/registry.go
Tabaie: > The documentation does not match the implementation

Fixed!

> I think the gate should know itself where it can work (if it doesn't work over all fields) and then panic.

I prefer a compile time error to a solving/proving time panic.

> And I think we can still use the approach where we test the degree of the gate on a random modulus instead of letting the users define it themselves.

That idea is appealing, but for example in gates for hash functions we have the permutation constants cached as `fr.Element`s, and if we run that on a random modulus the gate will have to panic, making degree testing fail.

std/gkrapi/compile.go
Tabaie: > Could we potentially have a perf regression here? 

I think so but since this is only run during compile time (or with the test engine) I figured there's no need to worry much about perf. If we care about it I think a `Circuit.Grow` function similar to `bytes.Buffer` would be a good idea. Thoughts?

> Thats why it would be nice to keep the previous approach as well 

There's [another PR](https://github.com/Consensys/gnark/pull/1589) that brings back v1. Would you like me to merge it into this?

std/gkrapi/compile.go
Tabaie: The test engine, r1cs and scs APIs all implement this. Do we have a lot of use cases with user implemented APIs? If that's the case I agree we can return proper errors.

internal/gkr/gkrhints/engine_hints.go
Tabaie: The test engine doesn't compile. It directly solves the circuit as it goes through the circuit `Define` function. So the default hints have to be compatible with the test engine, because unlike the R1CS / Plonkish case, we won't have the opportunity to replace them with the right ones later.

internal/gkr/gkrhints/engine_hints.go
Tabaie: `engine.NewHint` reduces the output for you, and `len(in)` is not big enough to cause perf issues in the meantime.

internal/gkr/bls12-377/solver_hints.go
Tabaie: Yes essentially. I'll try to improve the comment.

internal/gkr/bls12-377/solver_hints.go
Tabaie: Fixed.

internal/gkr/gkrtypes/types.go
Tabaie: I'm open to fewer packages. This was the minimum number I could get without causing import cycles.

internal/gkr/bls12-377/solver_hints.go
Tabaie: Added docs for `WithAssignments` and `NewSolvingData`

constraint/solver/gkrgates/registry.go
Tabaie: I just check and that's what happens. If there is a duplicate key, it checks if it's actually the same gate by evaluating at a random point. If they are not the same gate, it errors (which is crucial, otherwise the user will later use the wrong gate having no idea something is wrong.) If they're the same gate, nothing happens.
If there isn't a duplicate key, degree verification and more follows.

ivokub (dismissed):

PR #1671, opened 2026-01-07
test: add small field test in checkcircuit

ivokub:
# Description

Adds automatic small field testing in `assert.CheckCircuit`. It currently will fail for some gadgets in gnark as it is not supported yet.

I'm still not sure what the default should be -- we will probably break a lot of downstream if making small fields check default. But we also want to test circuits inside gnark and Linea.

So perhaps right now we can make the test only warn (and fail only when "smallfield_checks" build tag is set) and then in the future we can make it more restrictive and start failing. WDYT @gbotrel? In that case we can update our CI to also run small field tests so that we can see any failures but we don't break the users.

edit: now looking at the failing tests I guess it is better not to make it default anyway, for some circuits it is impossible to solve with small field anyway. For example MiMC is not defined.

## Type of change

# Checklist:

---

> [!NOTE]
> Introduces optional small-field testing and refactors compile paths to be field-based with deterministic checks.
> 
> - Adds small-field (koalabear) compilation/solving in `assert.CheckCircuit` via `CompileU32` and `widecommitter`, gated by `smallfield_checks` build tag or `WithSmallfieldCheck` option; can be disabled with `WithoutSmallfieldCheck`/`WithoutCurveChecks`
> - Refactors APIs to pass `field *big.Int` instead of `ecc.ID` (e.g., `compile`, `parseAssignment`, `marshalWitnessJSON`); updates fuzz and solver flows accordingly
> - Enforces deterministic compilation for both standard and U32 compilers (`reflect.DeepEqual` double-compile check)
> - Adds profile/flags wiring (`checkSmallField`, build-tag files) and minor plumbing for serialization JSON helpers
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d2f9d0c3f2c2910f0963a72d299811e90d630ce. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/assert_options.go
+++ b/test/assert_options.go
@@ -204,24 +204,24 @@ func WithSmallfieldCheck() TestingOption {
 	}
 }
 
-// WithSkipSmallfieldCheck is a testing option which disables checking circuit
+// WithoutSmallfieldCheck is a testing option which disables checking circuit
 // compilation and solving in a small field. This overrides the
 // "smallfield_checks" build tag.
 //
 // The option is useful for tests that specifically want to skip small field
 // checks. For example, when we test hash functions (MiMC/Poseidon2) which are
 // not defined over small fields.
-func WithSkipSmallfieldCheck() TestingOption {
+func WithoutSmallfieldCheck() TestingOption {
 	return func(tc *testingConfig) error {
 		tc.checkSmallField = false
 		return nil
 	}
 }
 
-// WithNoCurves is a testing option which disables all curve checks. This method
+// WithoutCurveChecks is a testing option which disables all curve checks. This method
 // can be used if we only want to test over small fields. It overrides the build
 // tag settings and any curve-related options.
-func WithNoCurves() TestingOption {
+func WithoutCurveChecks() TestingOption {
 	return func(tc *testingConfig) error {
 		tc.curves = nil
 		return nil


test/assert_options.go
```diff
@@ -187,3 +192,38 @@ func WithSolidityExportOptions(solidityOpts ...solidity.ExportOption) TestingOpt
...
+}
+
+// WithSkipSmallfieldCheck is a testing option which disables checking circuit
+// compilation and solving in a small field. This overrides the
+// "smallfield_checks" build tag.
+//
+// The option is useful for tests that specifically want to skip small field
+// checks. For example, when we test hash functions (MiMC/Poseidon2) which are
+// not defined over small fields.
+func WithSkipSmallfieldCheck() TestingOption {
```
gbotrel: WithoutSmallfieldCheck?

test/assert_options.go
```diff
@@ -187,3 +192,38 @@ func WithSolidityExportOptions(solidityOpts ...solidity.ExportOption) TestingOpt
...
+	return func(tc *testingConfig) error {
+		tc.checkSmallField = false
+		return nil
+	}
+}
+
+// WithNoCurves is a testing option which disables all curve checks. This method
+// can be used if we only want to test over small fields. It overrides the build
+// tag settings and any curve-related options.
+func WithNoCurves() TestingOption {
```
gbotrel: WithoutCurveCheck?

ivokub: Agree, changed!

test/assert_options.go
ivokub: Agree, changed!

PR #1672, opened 2026-01-08
feat: add internal Koalabear E4 extension for field emulation and logderivative argument

ivokub:
# Description

Previously we used direct extension, but ext4 is more performant and allows easier implementation of inverses.

NB! Part of several PRs. Merge only after #1671 

## Type of change

# How has this been tested?

- [x] TestKoalabearExt4Mul
- [x] TestKoalabearExt4Inverse

# Checklist:

---

> [!NOTE]
> Adds an internal extension-field implementation and wires it through small-field flows.
> 
> - New `std/internal/fieldextension` package: Koalabear quadratic-tower E4 (`newKoalabearExt4`) with ops (`Mul`, `Add`, `Sub`, `MulByElement`, `Inverse`), helpers (`Element.Initialize`, `ValueOf`), defaults and degrees, and inverse hints (`inverseE2Hint`, `inverseE4Hint`).
> - Register field-extension hints in `std/hints.go`.
> - Switch emulated field and multicommit to use `std/internal/fieldextension`; evaluate deferred checks in the extension when on small fields.
> - Wide commitment over small fields now uses degree 4; tests updated accordingly.
> - API tweaks: rename `WithExtension` -> `WithDirectExtension`; default Koalabear degree set to 4; support auto-selection of Koalabear E4 when modulus matches.
> - Extensive tests for e2/e4 ops, inverse, flatten/unflatten, `ValueOf`, and multicommit wide commitments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 927cceb6a5c804ffc415df75ac2f95f19f84c46f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/std/internal/fieldextension/koalabear_ext_test.go
+++ b/std/internal/fieldextension/koalabear_ext_test.go
@@ -347,7 +347,7 @@ func TestKoalabearExt4Mul(t *testing.T) {
 		b.B0.A0.SetInt64(-1)
 		b.B0.A1.SetInt64(-1)
 		b.B1.A0.SetInt64(-1)
-		b.B0.A0.SetInt64(-1)
+		b.B1.A1.SetInt64(-1)
 
 		c.Mul(&a, &b)
 


std/internal/fieldextension/fieldextension.go
```diff
@@ -115,6 +130,36 @@ func NewExtension(api frontend.API, opts ...Option) (Field, error) {
...
+
+// ValueOf converts a value of type any to an extension field element. If the
+// type is not supported, it panics.
+//
+// Currently supported types are:
+//   - [github.com/consensys/gnark-crypto/field/koalabear/extensions.E4]
+func ValueOf(a any) Element {
+	switch v := a.(type) {
+	case extensions.E4:
+		return Element{v.B0.A0, v.B0.A1, v.B1.A0, v.B1.A1}
```
YaoJGalteland, line 160: Could you write some tests for this function? For example, test random inputs, zero values, and negative values to check whether the output matches the expected `[]frontend.Variable`.

I previously ran into an issue in the monorepo with negative values, where a Koalabear field element representing −1 (i.e., p−1 = 2130706432) was incorrectly converted to the BLS12-377 max value instead of remaining 2130706432. I asked an AI agent to generate a test, it looks fine in your implementation and doesn’t reproduce the issue I saw in the monorepo. it simply panics on an invalid input type^^

Regarding functionality, It would also be good to include a case for a base field type like E1 / fr.Element, returning `Element{v, 0, 0, 0}`. This pattern is used frequently throughout the monorepo.

std/internal/fieldextension/koalabear_ext_test.go
```diff
@@ -0,0 +1,341 @@
...
+	api.AssertIsEqual(res.A0, c.C.A0)
+	api.AssertIsEqual(res.A1, c.C.A1)
+
+	return nil
+}
+
+func TestKoalabearExt2Add(t *testing.T) {
+	assert := test.NewAssert(t)
+	var a, b, c extensions.E2
+	a.MustSetRandom()
```
YaoJGalteland, line 29:
```
func (z *E2) MustSetRandom() *E2 {
    if _, err := z.A0.SetRandom(); err != nil {
        panic(err)
    }
    return z  // A1 is never set!
}
```
MustSetRandom only sets A0 to random!

std/internal/fieldextension/fieldextension.go
```diff
@@ -43,8 +49,11 @@ type Field interface {
 	Reduce(a Element) Element
 	// Mul multiplies two extension field elements and reduces the result.
 	Mul(a, b Element) Element
```
YaoJGalteland, line 51: The E2/E4 extensions don't implement Square(), would that be a potential optimization?

std/internal/fieldextension/koalabear_ext_test.go
```diff
@@ -0,0 +1,341 @@
+package fieldextension
+
```
YaoJGalteland, line 2: Missing Negative/Edge Case Tests, testing random values may not be enough, consider adding tests for:
- operations with special values like −1, 0, or 1, which may hit edge cases, especially since −1 corresponds to the max value p−1.
- different input lengths: empty, 1, 2, 3, 100, etc. (no test for `unflattenE4`)

std/internal/fieldextension/hints.go
```diff
@@ -0,0 +1,54 @@
...
+// registering the hints when solving/proving serialized circuits. The hints are
+// automatically registered when importing current package.
+func GetHints() []solver.Hint {
+	return []solver.Hint{
+		inverseE2Hint,
+		inverseE4Hint,
+	}
+}
+
+func inverseE2Hint(_ *big.Int, inputs []*big.Int, res []*big.Int) error {
```
YaoJGalteland, line 25: The hints `inverseE2Hint` and `inverseE4Hint` don't validate inputs. Consider adding a check for example: 
```
if len(inputs) != 2 || len(res) != 2 {
        return fmt.Errorf("expected 2 inputs and 2 outputs")
    }
```

std/internal/fieldextension/koalabear_ext.go
```diff
@@ -0,0 +1,217 @@
...
+	B0, B1 e2
+}
+
+// flattenE4 flattens an e4 element into a slice of frontend.Variable. It is
+// used in the fieldextension package and we want to have compatibility.
+func flattenE4(e e4) []frontend.Variable {
+	return []frontend.Variable{e.B0.A0, e.B0.A1, e.B1.A0, e.B1.A1}
+}
+
+func unflattenE4(vars []frontend.Variable) e4 {
```
YaoJGalteland, line 24: Is this function missing a Panic `if len(vars) > 4`? even `unflattenE4` currently only handles cases where `len(vars)` is 0-4. 

It’s also missing a consistency test to check that flattening and then unflattening is the identity.

std/internal/fieldextension/koalabear_ext_test.go
ivokub: Very good catch! Pushed https://github.com/Consensys/gnark-crypto/pull/793. Tested locally with it and all tests still pass. Will have to wait when PR is merged to gnark-crypto to update the dependency as well.

std/internal/fieldextension/fieldextension.go
ivokub: This package is only used internally for implementing non-native multiplication checks. It isn't even exported (it is subpackage of `internal` which cannot be imported outside of the `std/` dir). And there we don't need Square. We can implement it when becomes necessary, but want to keep interfaces minimal for now.

std/internal/fieldextension/koalabear_ext.go
ivokub: Added test

std/internal/fieldextension/hints.go
ivokub: Added checks

std/internal/fieldextension/koalabear_ext_test.go
ivokub: 2. included in flatten tests

std/internal/fieldextension/fieldextension.go
ivokub: Added case for setting only base field element. It is a good point, I think it is useful indeed.

ivokub: Also added edge case tests (0, -1, 1) when setting from base field. For e4 kept only the random case, I think it is sufficient enough that works for base field case.

NB! This works here as we only run the tests over Koalabear, where -1 is as you expect.

std/internal/fieldextension/koalabear_ext_test.go
ivokub: Added more edge case tests to E4 Mul and Inverse. Tried adding to other tests as well, but seems excessive considering that Mul and Inverse pass as expected.

std/internal/fieldextension/koalabear_ext_test.go
```diff
@@ -0,0 +1,596 @@
...
+		var a, b, c extensions.E4
+		a.B0.A0.SetInt64(-1)
+		a.B0.A1.SetInt64(-1)
+		a.B1.A0.SetInt64(-1)
+		a.B1.A1.SetInt64(-1)
+
+		b.B0.A0.SetInt64(-1)
+		b.B0.A1.SetInt64(-1)
+		b.B1.A0.SetInt64(-1)
+		b.B0.A0.SetInt64(-1)
```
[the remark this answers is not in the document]
ivokub: Fixed

std/internal/fieldextension/fieldextension.go
YaoJGalteland: Do you have a `ValueOf` function for just the base field? I forgot to ask this. For that function in the monorepo, we usually take an `int`, `big.Int` or a `field.Element` as input.

ivokub: I modified this method such that depending what is the provided input we create either `{base, 0, 0, 0}` or full extension element. See the switch statement. This could be extend to take also `int`, `*big.Int` etc. But in this case we don't really need it, I added it only for being able to run tests :) It is internal package for field emulation where we already define the extension element in-circuit.

YaoJGalteland (approved):

PR #1674, opened 2026-01-08
fix: handle uints8 Long addition even in small field

ivokub:
# Description

For small fields our long (U32, U64) addition overflowed the native field. Modify the implementation to operate on bytes only to avoid overflows.

NB! Merge only after #1673 is merged.

## Type of change

# How has this been tested?

- Test addition also with small fields.
- added more edge case tests

# Checklist:

---

> [!NOTE]
> Improves long-integer arithmetic and adds byte packing utilities.
> 
> - Refactors `BinaryField.Add` to accept variable inputs (including 0/1) and to switch between native add+partition and bytewise add-with-carry when `maxBitlen` exceeds the field size, preventing overflow on small fields
> - Adds `PackMSB`, `PackLSB`, `UnpackMSB`, `UnpackLSB` for constructing/deconstructing `T` from bytes; adds `zero()` helper
> - Documents bitwise ops and shifts; no functional change to `And`/`Or`/`Xor`/`Not`, `Lrot`, `Rshift` besides comments
> - Tests: rewrite add circuit to slice-based and add cases for 0, 1, 2, and many inputs with `WithSmallfieldCheck`; add pack/unpack circuit tests
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 866e41caa4cb2041596043f11f2c3cf43b998ba4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/std/math/uints/uint8.go
+++ b/std/math/uints/uint8.go
@@ -272,7 +272,7 @@ func (bf *BinaryField[T]) Add(a ...T) T {
 	// the field. So we do the additions step by step, partitioning after every
 	// addition to ensure that the intermediate results never overflow the
 	// field.
-	if maxBitlen <= bf.api.Compiler().FieldBitLen() {
+	if maxBitlen < bf.api.Compiler().FieldBitLen() {
 		// handle the easy case. For this, we just compose the bytes into a
 		// native frontend.Variable, perform the addition natively drop the the
 		// carry bit and then re-split into bytes.


std/math/uints/uint8.go
```diff
@@ -217,21 +250,86 @@ func (bf *BinaryField[T]) Not(a T) T {
...
-	res := bf.ValueOf(vreslow)
-	return res
+	// when we use large fields where maxBitLen < field size, then we can just
+	// add all the values directly and then partition. However, when
+	//    maxBitlen >= field size
+	// then we need to make sure that we never have an addition which overflows
+	// the field. So we do the additions step by step, partitioning after every
+	// addition to ensure that the intermediate results never overflow the
+	// field.
+	if maxBitlen < bf.api.Compiler().FieldBitLen() {
```
YaoJGalteland: Condition should be **<=** not <

when `maxBitlen == FieldBitLen()`, it can still overflow the field

std/math/uints/uint8.go
```diff
@@ -217,21 +250,86 @@ func (bf *BinaryField[T]) Not(a T) T {
...
+			// bytewise addition with carry
+			var vres frontend.Variable
+			if i > 0 {
+				vres = bf.api.Add(carry, aij[0], aij[1:]...)
+			} else {
+				vres = bf.api.Add(aij[0], aij[1], aij[2:]...)
+			}
+			vreslow, vreshigh := bitslice.Partition(bf.api, vres, 8, bitslice.WithNbDigits(maxBitlen), bitslice.WithUnconstrainedOutputs())
+			// store the result byte
+			res[i] = bf.ByteValueOf(vreslow)
```
YaoJGalteland, line 324: `bitslice.Partition` already constrains `vreslow`<2^8, it seem `ByteValueOf` has a redundant range check, because it calls `ValueOf` --> `enforceWidth`, Should we create (or already exist one?) a variant of `ValueOf` that bypasses `enforceWidth` to avoid this repetition?

std/math/uints/uint8.go
```diff
@@ -217,21 +250,86 @@ func (bf *BinaryField[T]) Not(a T) T {
...
+			va[i] = bf.ToValue(a[i])
+		}
+		vres := bf.api.Add(va[0], va[1], va[2:]...)
+		// bitslice.Partition below checks that the input is less than 2^maxBitlen and that we have omitted carry correctly
+		vreslow, _ := bitslice.Partition(bf.api, vres, uint(tLen), bitslice.WithNbDigits(maxBitlen), bitslice.WithUnconstrainedOutputs())
+		res := bf.ValueOf(vreslow)
+		return res
+	} else {
+		// however, if the result when combining the bytes doesn't fit into the
+		// native field (i.e. we work over Koalabear), then we cannot use the
```
YaoJGalteland, line 291: Instead of performing bytewise addition, how about using 16-bit chunk addition? Since we're working within a native field, this wouldn't cause an overflow. This halves the number of iterations and partition operations

std/math/uints/uint8.go
ivokub: Here I have set `bitslice.WithUnconstrainedOutputs()` option for the `bitslice.Partition` which avoids constraining the `vreslow` and `vreshigh`. And I set it because `ValueOf` constrains in-range, so it should avoid doing it twice.

std/math/uints/uint8.go
ivokub: Yep, 16-byte should also work. I'll test it.

std/math/uints/uint8.go
ivokub: Good catch!

std/math/uints/uint8.go
ivokub: It worked, but didn't give any noticeable difference. The problem is that when we work with 16-bit parts, then have to perform 2 partitions per loop:
- first to cut off the carry for next round (at 16-bit mark)
- then to split lower part of the loop result into 8 bit parts (for storing in the result).

Partitioning isn't free (we have to show its correctness). It also created a problem on linea-monorepo side as we need to keep track which variables we range check, and we would have to add a dummy constraint to align the variables.

So I'd stay with 8-bit addition for now until we can come up with a smarter solution.

std/math/uints/uint8.go
ivokub: Oh, but actually it was correct though? When `maxBitLen == FieldBitLen()` then it can overflow --> we will take the byte-by-byte approach. It was previously like this, then I changed the condition to `<=`. I now reverted back.

std/math/uints/uint8.go
```diff
@@ -217,21 +250,86 @@ func (bf *BinaryField[T]) Not(a T) T {
...
-	res := bf.ValueOf(vreslow)
-	return res
+	// when we use large fields where maxBitLen < field size, then we can just
+	// add all the values directly and then partition. However, when
+	//    maxBitlen >= field size
+	// then we need to make sure that we never have an addition which overflows
+	// the field. So we do the additions step by step, partitioning after every
+	// addition to ensure that the intermediate results never overflow the
+	// field.
+	if maxBitlen <= bf.api.Compiler().FieldBitLen() {
```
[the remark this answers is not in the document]
ivokub: Fixed in revert

std/math/uints/uint8.go
YaoJGalteland: yes, you are right. the original `maxBitlen` can be up to 30

YaoJGalteland (approved):

std/math/uints/uint8.go
```diff
@@ -217,21 +250,86 @@ func (bf *BinaryField[T]) Not(a T) T {
...
+			var vres frontend.Variable
+			if i > 0 {
+				vres = bf.api.Add(carry, aij[0], aij[1:]...)
+			} else {
+				vres = bf.api.Add(aij[0], aij[1], aij[2:]...)
+			}
+			vreslow, vreshigh := bitslice.Partition(bf.api, vres, 8, bitslice.WithNbDigits(maxBitlen), bitslice.WithUnconstrainedOutputs())
+			// store the result byte
+			res[i] = bf.ByteValueOf(vreslow)
+			carry = vreshigh // we omit the last carry as performing addition modulo 2^(lenBts*8)
```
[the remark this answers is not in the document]
ivokub: We output input and one of the outputs. This transitively also range checks other output.

PR #1689, opened 2026-01-22
feat: add PointEval precompiles for 16-bit limbs

ivokub:
# Description

Adds new precompile methods for PointEval and PointEvalFailure when operating on 16-bit limbs. For backwards compatibility kept existing 128-bit limb approach.

## Type of change

# How has this been tested?

Converted the tests for 16-bit limb case. Also the tests working on linea-monorepo side.

# Checklist:

---

> [!NOTE]
> Adds support for 16-bit limb encoding for the KZG PointEval precompile alongside the existing 128-bit path, and tightens circuit constraints.
> 
> - New `KzgPointEvaluation16` and `KzgPointEvaluationFailure16` using 2-byte limbs with format checks, modulus/blob-size validation, and shared generic logic (`kzgPointEvaluation*`) parameterized by bytes-per-limb
> - Duplicate fixed KZG verifying key for 16-bit path; add 16-bit BLS modulus and blob-size constants; guard usage by field size (small vs large fields)
> - Factor helpers for selecting slices (`selectSliceUint8`, `selectSliceVariable`) and unify limb-to-byte conversions
> - Tests: add full 16-bit tests; keep 128-bit tests separate; include trusted setup parsing utilities
> - Optimize `bitslice.Partition` to emit a single Plonk constraint where applicable; remove unconstrained outputs usage in `uints` and rely on bounded `Partition`
> 
> Scope: evmprecompiles + std/math; primarily new feature plus minor refactors/optimizations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 01bab82471bb948198b7df2f45337ad3b9e1f0c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/std/evmprecompiles/10-kzg_point_evaluation.go
+++ b/std/evmprecompiles/10-kzg_point_evaluation.go
@@ -8,6 +8,7 @@ import (
 	bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381"
 	kzg_bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381/kzg"
 	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/internal/smallfields"
 	"github.com/consensys/gnark/std/algebra"
 	"github.com/consensys/gnark/std/algebra/algopts"
 	"github.com/consensys/gnark/std/algebra/emulated/sw_bls12381"
@@ -156,6 +157,9 @@ func KzgPointEvaluation(
 	expectedBlobSize [2]frontend.Variable, // arithmetization uses 2-element array. It is constant for all purposes, but we check it anyway.
 	expectedBlsModulus [2]frontend.Variable, // arithmetization uses 2-element array. It is constant for all purposes, but we check it anyway.
 ) error {
+	if smallfields.IsSmallField(api.Compiler().Field()) {
+		return fmt.Errorf("KzgPointEvaluation method cannot be called over small field compiler")
+	}
 	// check expected modulus against 128-bit format
 	api.AssertIsEqual(expectedBlsModulus[0], evmBlsModulusHi)
 	api.AssertIsEqual(expectedBlsModulus[1], evmBlsModulusLo)
@@ -191,6 +195,9 @@ func KzgPointEvaluation16(
 	expectedBlobSize [16]frontend.Variable, // arithmetization uses 16-element array (2-byte words). It is constant for all purposes, but we check it anyway.
 	expectedBlsModulus [16]frontend.Variable, // arithmetization uses 16-element array. It is constant for all purposes, but we check it anyway.
 ) error {
+	if !smallfields.IsSmallField(api.Compiler().Field()) {
+		return fmt.Errorf("KzgPointEvaluation16 method cannot be called over large field compiler")
+	}
 	// check expected modulus against 16-bit format
 	for i := range evmBlsModulus16 {
 		api.AssertIsEqual(expectedBlsModulus[i], evmBlsModulus16[i])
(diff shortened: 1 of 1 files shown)


std/evmprecompiles/10-kzg_point_evaluation.go
```diff
@@ -89,8 +89,37 @@ const (
 	evmBlsModulusLo = "0x53bda402fffe5bfeffffffff00000001"
 )
 
+var (
```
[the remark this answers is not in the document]
ivokub: It is commented though? And it is not exported?

std/evmprecompiles/10-kzg_point_evaluation.go
YaoJGalteland, line 150: This name is a bit confusing, since the generic function is also called `KzgPointEvaluation`

YaoJGalteland, line 152: As below, shall we consider range checks for each limbs? i.e. checking if limb <= 128 bit.

std/evmprecompiles/10-kzg_point_evaluation.go
```diff
@@ -117,8 +156,70 @@ func KzgPointEvaluation(
...
+// Values commitment and proof are the KZG commitment and proof respectively.
+// They are given as compressed points, for which we use 24 native elements to
+// represent. The method performs decompression and all necessary checks. The
+// encoding is given by Appendix C of [PAIRING_FRIENDLY_CURVES].
+//
+// [KZG_POINT_EVALUATION]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md
+// [PAIRING_FRIENDLY_CURVES]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-pairing-friendly-curves/
+func KzgPointEvaluation16(
+	api frontend.API,
+	versionedHash [16]frontend.Variable, // arithmetization gives us a 16-element array. We convert it ourselves to a byte array.
```
YaoJGalteland, line 190: Shall we consider range checks for each limbs? i.e. checking if limb <= 16 bit. without the range check, a malicious prover could exploit this to break the "knowledge soundness" property. Specifically, multiple different witnesses could produce the same proof, allowing the prover to claim knowledge of values they don't actually possess in the expected format.

**Example**: Consider verifying a KZG commitment where the commitment's X coordinate matters:
Legitimate commitment X coordinate (48 bytes):
  0x97f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb

Encoded as 24 limbs:
  limb[0]  = 0x97f1
  limb[1]  = 0xd3a7
  limb[2]  = 0x3197
  ...
**Attack**: Prover submits:

// Malicious limbs with overflow:
limb[0] = 0x197f1   // 17-bit, Overflow 2 bytes! Will be truncated to 0x97f1 as Circuit extracts last 2(=bytesPerLimb) bytes
limb[1] = 0x2d3a7   // 18-bit, Overflow 2 bytes! Will be truncated to 0xd3a7 as Circuit extracts last 2 bytes
...
The circuit extracts the same bytes [0x97, 0xf1, 0xd3, 0xa7, ...], so the KZG verification passes.

std/evmprecompiles/10-kzg_point_evaluation.go
ivokub: It is already checked in arithmetization.

std/evmprecompiles/10-kzg_point_evaluation.go
ivokub: In general we should, but this case the method is called from the BLS glue which passes in directly the traces from arithmetization. And it already checks the limb lengths. So no need to duplicate here.

std/evmprecompiles/10-kzg_point_evaluation.go
ivokub: Yup, it would be better to have `KzgPointEvaluation128` etc. But I didn't want to change existing method name not to break monorepo side.

PR #1697, opened 2026-01-31
perf: optimize scalar multiplications and multi-scalar multiplications circuits via lattice reductions

yelhousni:
# Description

This PR migrates gnark's scalar decomposition hints from the `eisenstein` package to the new `lattice` package in https://github.com/Consensys/gnark-crypto/pull/799, following the lattice-based rational reconstruction approach from _"Fast elliptic curve scalar multiplications in SN(T)ARK circuits"_ by Eagen-ElHousni-Masson-Piellard (https://eprint.iacr.org/2025/933.pdf).

The new approach provides **proven bounds** from LLL lattice reduction theory, replacing heuristic bounds. This allows tighter bit-width bounds for the decomposed scalars, reducing circuit constraints.

The PR also revisits the complete arithmetic path to make it more constraint-optimized.

#### Hint Renames
- `halfGCD` → `rationalReconstruct` (2-part decomposition using `lattice.RationalReconstruct`)
- `halfGCDEisenstein` → `rationalReconstructExt` (4-part decomposition using `lattice.RationalReconstructExt`)

#### Tighter Bounds
The number of bits for decomposed scalars has been reduced:
- **Old**: `r.BitLen()/4 + 9` (heuristic with large safety margin)
- **New**: `(r.BitLen()+3)/4 + 2` (proven bound from LLL: outputs `< 1.25·r^(1/4)`)

This saves ~7 iterations in the scalar multiplication loop.

#### Affected Packages
- `std/algebra/emulated/sw_emulated` - Emulated short Weierstrass curves G1
- `std/algebra/emulated/sw_bls12381`, `std/algebra/emulated/sw_bn254` and `std/algebra/emulated/sw_bw6761` - emulated G2
- `std/algebra/native/sw_bls12377` - Native BLS12-377 G1 and G2
- `std/algebra/native/twistededwards` - Native twisted Edwards curves

## Type of change

- [x] Optimization

# How has this been tested?

All existing tests pass:
```
go test -short ./std/algebra/emulated/sw_emulated/...
go test -short ./std/algebra/native/sw_bls12377/...
go test -short ./std/algebra/native/twistededwards/...
```

#### G1 scalar multiplication:

  | Curve/**G1** | Old (eisenstein) | New (lattice) | Δ | Improvement |
  |-------|------------------|---------------|---|-------------|
 |**emulated**| | | | |
  | secp256k1 | 394,959 | 364,495 | **-30,464** | **7.7%** |
  | BN254 | 390,317 | 364,205 | **-26,112** | **6.7%** |
  | BLS12-381 | 550,299 | 512,235 | **-38,064** | **6.9%** |
  | BW6-761 | 1,376,789 | 1,317,113 | **-59,676** | **4.3%** |

#### G2 scalar multiplication:

  | Curve/**G2** | Old (2D GLV) | New (4D-lattice GLV+FakeGLV) | Δ | Improvement |
  |-------|--------------|----------------------|---|-------------|
|**emulated**| | | | |
  | BN254  | 599,779 | 411,854 | **-187,925** | **31.3%** |
  | BLS12-381  | 913,513 | 584,794 | **-328,719** | **36.0%** |
  | BW6-761  | 1,090,960 | 728,286 | **-362,674** | **33.2%** |

#### G1 MSM of size 2 :

  | Curve | GLV | Method | Old | New | Δ | Improvement |
  |-------|-----|---|--|-----|---|---|
|**emulated (short Weierstrass)**| | | | | | |
  | P256 | No | 2 scalar muls + add |  523,062 | 277,544 | **-245,518** | **46.9%** |
|**native (twisted Edwards)**| | | | | | |
  | BabyJubjub (BN254) | No | lattice 3-MSM with LogUp | 9,956 | 6,785 | **-3,171** | **32%** |
  | Jubjub (BLS12-381) | No | lattice 3-MSM with LogUp | 9,930 | 6,793 | **-3,137** | **32%** |
  | Bandersnatch (BLS12-381) | Yes | lattice 6-MSM with LogUp | 10,185 | 6,820 | **-3,365** | **33%** |

#### Applications:

  | Precompile | Old | New | Δ | Improvement |
  |-----------|-----|-----|---|-------------|
  | P256Verify | 666,146 | 533,201 | **-132,945** | **20%** |
  | BLSG2MSM (10 pairs) | 9,304,517 | 7,135,929 | **-2,168,588** | **23.3%** |
  | ECMul (BN254) | 210,369 | 195,663 | **-14,706** | **7.0%** |
  | BLSG1MSM (10 pairs) | 4,397,157 | 4,243,497 | **-153,660** | **3.5%** |
  | KZGPointEval | 2,928,188 | 2,897,456 | **-30,732** | **1.0%** |

  | PLONK recursion | Old | New | Δ | Improvement |
  |-----------|-----|-----|---|-------------|
  | Emulated (BW6-761 in BN254) | 15,042,004 | 14,713,771 | **-328,233** | **2.2%** |

  | EdDSA | GLV | Old | New | Δ | Improvement |
  |-------|-----|-----|-----|---|---|
  | Jubjub (BLS12-381) | No | 13,570 | 10,680 | **-2,890** | **21%** |
  | Bandersnatch (BLS12-381) | Yes | 13,835 | 10,706 | **-3,129** | **23%** |

#### 1. Hint Computation Time

**2-part decomposition**

  | Method | Time | Notes |
  |--------|------|-------|
  | Old (HalfGCD) | 9.5 μs | xgcd (PrecomputeLattice) |
  | New (uncached) | 5.0 μs | xgcd |
  | **New (cached)** | **3.8 μs** | xgcd (Cached reconstructor) |

**4-part decomposition (RationalReconstructExt) - GLV curves**

  | Method | Time | Notes |
  |--------|------|-------|
  | Old (HalfGCDEisenstein) | 43 μs | Eisenstein HalfGCD |
  | New (uncached) | 43 ms | LLL from scratch |
  | **New (cached)** | **43 ms** | LLL (Caching doesn't help here) |

The new approach is slower for hint computation (4D) because it runs LLL reduction from scratch rather than using 2-step Eisenstein half-GCD. However, hint computation happens outside the prover and is negligible compared to proof generation time. The constraint reduction provides a net benefit.

**3-part decomposition (MultiRationalReconstruct) - 2 scalars**

  | Method | Time | Notes |
  |--------|------|-------|
  | New (uncached) | 1.7 ms | LLL from scratch |
  | **New (cached)** | **0.6 μs** | LLL (Huge speedup with caching) |

**6-part decomposition (MultiRationalReconstructExt) - 2 scalars**

  | Method | Time | Notes |
  |--------|------|-------|
  | New (uncached) | 563 ms | LLL from scratch |
  | **New (cached)** | **500 ms** | Minimal improvement |

#### 2. `logup` vs `Mux`
For G1 we can do a 4-MSM. For G2 we can leverage the Frobenius as a second endomorphism, we can apply it to all and get a 8-MSM or to half and get a 6-MSM. But with big tables `Mux` becomes the bottlneck, so we can try with `logup`.

- G1 BLS12-381:

  | Method | Mux | Logup | Savings |
  |--------|-----|-------|---------|
  | G1 4D GLV+FakeGLV | **272,520** | 340,273 | +24.9% worse |

- G2 BLS12-381:

  | Method | Mux | Logup | Savings |
  |--------|-----|-------|---------|
  | G2 4D GLV+FakeGLV | **584,794** | 741,975 | +26.9% worse |
  | G2 6D GLV+GLS+FakeGLV | 1,645,337 | 1,512,349 | -8.1% better |
  | G2 8D GLV+GLS+FakeGLV | 6,239,417 | 5,786,450 | -7.3% better |

The 4D GLV+FakeGLV method with `Mux` remains optimal for single scalar multiplication on both G1 and G2. Higher-dimensional methods (6D, 8D) using the ψ endomorphism don't reduce constraints because the `Mux`/`logup` overhead outweighs the benefits of fewer loop iterations, even with logup optimization.

#### 3. MSM
According to [EEMP25], we can turn a `MSM(2,n)` verification (i.e. a MSM of size 2 with scalars of n bits) into a `MSM(3,2n/3)` or `MSM(6,n/3)` verification. We implemented this for the native (SW and tEd) and emulated (SW) cases with `Mux` and `logup` (for native). For all the scenario existing algorithms were better except for:
- native non-GLV tEd `MSM(3,2n/3)` with `LogUp` 
- emulated non-GLV SW `MSM(3,2n/3)` with `Mux` 
- native GLV tEd `MSM(6,n/3)` with `Mux`(Bandersnatch).

# Checklist:

- [x] Any dependent changes have been merged and published in downstream modules (gnark-crypto lattice package)

---

> [!NOTE]
> **High Risk**
> High risk because it rewrites core elliptic-curve scalar multiplication verification logic and hint interfaces (including complete-arithmetic edge cases) across multiple curves; any mistake can silently break proof soundness or curve arithmetic correctness.
> 
> **Overview**
> **Optimizes elliptic-curve scalar multiplication verification in circuits** by replacing Eisenstein/half-GCD based scalar decompositions with LLL-backed `lattice` rational reconstruction (`rationalReconstruct`, `rationalReconstructExt`) and tightening sub-scalar bit bounds.
> 
> Updates emulated SW code to use the new hints (including new G2 hints) and introduces a new GLV+fakeGLV-based `G2.ScalarMul` path for `bls12-381`, `bn254`, and `bw6-761`, with precomputed generator constants and expanded edge-case handling under `algopts.WithCompleteArithmetic()`.
> 
> Refactors native SW `bls12-377` to use lattice reconstruction for GLV+fakeGLV checks, adds a hint-backed *complete* `G1` joint-scalar-mul verifier, and adjusts arithmetic to avoid overflow via emulated scalar-field checks.
> 
> Improves MSM/joint-scalar-mul behavior in `sw_emulated` (prefer `ScalarMul`+Add for non-GLV, bias-point strategy, stricter edge-case/soundness handling) and updates twisted Edwards `DoubleBaseScalarMul` to choose between GLV and non-GLV optimized paths; adds extensive new tests/benchmarks and updates `internal/stats/latest_stats.csv` accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 986d2619fb90014dad00ab608252ce11b90d2e44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


std/algebra/native/sw_bls12377/g1.go
```diff
@@ -455,16 +457,188 @@ func (p *G1Affine) jointScalarMul(api frontend.API, Q, R G1Affine, s, t frontend
...
+	tContribZero := api.Or(tIsZero, RIsZero)
+
+	// when s contribution is zero, set s=1 to avoid issues with scalar decomposition
+	_s := api.Select(sContribZero, 1, s)
+	// when t contribution is zero, set t=1 to avoid issues with scalar decomposition
+	_t := api.Select(tContribZero, 1, t)
+
+	// Dummy points for edge cases - must be different to avoid table construction issues
+	dummyQ := G1Affine{X: 1, Y: 1}
+	dummyR := G1Affine{X: 2, Y: 1}
```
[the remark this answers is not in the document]
yelhousni: already fixed

std/algebra/native/sw_bls12377/g1.go
```diff
@@ -455,16 +457,154 @@ func (p *G1Affine) jointScalarMul(api frontend.API, Q, R G1Affine, s, t frontend
...
+	// Acc now equals [_s]*_Q + [_t]*_R
+	// For the common case (no edge cases), this equals the hinted result
+	// For edge cases, we skip verification and trust the hint
+	// The hint correctly computes edge cases, and the edge case conditions
+	// (s=0, t=0, Q=0, R=0) are verified through IsZero checks above
+
+	// Only verify for the common case (no edge cases)
+	// For edge cases, select Acc = result to make the assertion pass
+	Acc.Select(api, anyEdgeCase, result, Acc)
+	Acc.AssertIsEqual(api, result)
```
ivokub: I think there is a soundness issue here - in case we have any edge case we set `Acc` to hinted `result`. But if we have only partial case: s==0, Q any, t!=0, R!=(0,0) then the whole result should be `[t]*R`. But due to the select we don't actually check it is correct.

std/algebra/native/twistededwards/point.go
```diff
@@ -292,3 +245,355 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+	var R Point
+	// We need Q1 + Q2 = R
+	var Q1, Q2 Point
+	Q1.X, Q1.Y = qHint[0], qHint[1]
+	Q2.X, Q2.Y = qHint[2], qHint[3]
+	R.add(api, &Q1, &Q2, curve)
+
+	// Decompose (s1, s2) using MultiRationalReconstructExt
+	// Returns |x1|, |y1|, |x2|, |y2|, |z|, |t|, signX1, signY1, signX2, signY2, signZ, signT
+	h, err := api.NewHint(multiRationalReconstructExtHint, 12, s1, s2, curve.Order, endo.Lambda)
```
ivokub: I'm trying to find, but it seems we never assert in-circuit that the decomposition is correctly hinted?

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -523,261 +554,272 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
 // a non-zero accumulator point (R0). To do this, we skip the LSB (bit at
```
ivokub: It seems we have some comment left-over here?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -1288,26 +1318,28 @@ func (c *Curve[B, S]) scalarMulFakeGLV(Q *AffinePoint[B], s *emulated.Element[S]
...
 	// sub-scalars s1 and s2 are 1, and set:
 	// 		Acc = Q + R
-	Acc := addFn(tableQ[1], tableR[1])
+	// For complete arithmetic, we add a bias point G to avoid Acc == ±Bi during the loop.
+	// T2 = Q + R (without bias, used in table construction)
+	T2 := c.Add(tableQ[1], tableR[1])
+	Acc := T2
+	if cfg.CompleteArithmetic {
+		g := c.Generator()
+		Acc = c.Add(Acc, g)
```
ivokub: I also recall this idea previously that for edge cases instead of using unified addition we add the bieas. But wouldn't it create an issue when `tableQ[1]+tableR[1] == -G`?

ivokub (changes requested):
I had a quick pass but imo we have soundness issues right now as we don't check the scalar decomposition and trust the hinted joint scalarmul result in case of edge cases.

Also, a few tests are failing, possibly because we hit the edge cases already and the hint doesn't cover it?

std/algebra/native/sw_bls12377/g1.go
yelhousni: Replaced off-curve dummy points (1,1) and (2,1) with on-curve generator points (G1 generator and [3]*G1). Since the dummies are now on the curve, the loop produces valid results even for edge cases. At the end, the dummy contributions are subtracted using AddUnified (which handles (0,0) as identity, so it's a no-op for common cases). The assertion Acc.AssertIsEqual(api, result) now holds for all cases through actual computation, not by overwriting.

std/algebra/native/twistededwards/point.go
yelhousni: Extended multiRationalReconstructExtHint to output 8 additional verification values (d, kd, n1, kn1, n2, kn2, k1, k2). Added in-circuit verification of the algebraic decomposition s_i*(z + λ*t) ≡ x_i + λ*y_i (mod r) using intermediate reductions to avoid native field overflow. The verification is split into 5 checks, each sound because the overflow bound is strictly less than the curve order.

std/algebra/emulated/sw_bls12381/g2.go
yelhousni: removed leftover comment from the old scalarMulGeneric function that was partially overwritten by the new ScalarMul comment.

std/algebra/emulated/sw_emulated/point.go
yelhousni: Added a guard check t2EqNegG = (T2.X == G.X). When triggered, T2 is replaced with a safe dummy ([2]*G) before the bias addition. The t2EqNegG flag is added to selectorEdge so the fallback path is used for this edge case.

ivokub:
Hmm, something strange is happening right now -- testing with `-short` passes on twistededwards package, but not with `-tags prover_checks` (using actual solver).

And I think the issue is that we're using `DivUnchecked(0,0)` in
```go
func (p *Point) phi(api frontend.API, p1 *Point, curve *CurveParams, endo *EndoParams) *Point {

	xy := api.Mul(p1.X, p1.Y)
	yy := api.Mul(p1.Y, p1.Y)
	f := api.Sub(1, yy)
	f = api.Mul(f, endo.Endo[1])
	g := api.Add(yy, endo.Endo[0])
	g = api.Mul(g, endo.Endo[0])
	h := api.Sub(yy, endo.Endo[0])

	p.X = api.DivUnchecked(f, xy) // <---- here
	p.Y = api.DivUnchecked(g, h)

	return p
}
```
which is unconstrained by the `frontend.API`:
```go
	// DivUnchecked returns i1 / i2
	// If i1 == i2 == 0, the return value (0) is unconstrained.
	DivUnchecked(i1, i2 Variable) Variable
```
Here test engine silently returns 0 and R1CS solver as well (it could return anything though), but PLONK solver explicitly fails here.

So I think there is still the issue that the twistededwards path doesn't handle edge cases GLV in twistededwards.

Additionally, imo in another PR we should make test engine more strict to panic explicitly in case we have `DivUnchecked(0,0)` to avoid having unconstrained circuits during development time.

std/algebra/native/sw_bls12377/g1.go
```diff
@@ -682,68 +847,81 @@ func (p *G1Affine) scalarMulGLVAndFakeGLV(api frontend.API, P G1Affine, s fronte
 	// Checking Q - [s]P = 0 is equivalent to [v]Q + [-s*v]P = 0 for some nonzero v.
 	//
 	// The GLV curves supported in gnark have j-invariant 0, which means the eigenvalue
-	// of the GLV endomorphism is a primitive cube root of unity.  If we write
-	// v, s and r as Eisenstein integers we can express the check as:
+	// of the GLV endomorphism is a primitive cube root of unity λ. Using this we can
```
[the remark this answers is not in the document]
yelhousni: this seems to be pattern-matching from the emulated version's structure without understanding that the native GLV decomposition works differently and the H offset already prevents incomplete addition issues. It's a false positive.

std/algebra/emulated/sw_bls12381/hints.go
```diff
@@ -24,6 +25,8 @@ func GetHints() []solver.Hint {
 		pairingCheckHint,
 		millerLoopAndCheckFinalExpHint,
 		decomposeScalarG1,
```
ivokub, line 27: Hint not used anymore?

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -510,274 +541,276 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
-	// s == s3 + [λ]s4
-	g2.fr.AssertIsEqual(
-		g2.fr.Add(s3, g2.fr.Mul(s4, g2.eigenvalue)),
-		s,
+		panic(fmt.Sprintf("rationalReconstructExtG2 hint: %v", err))
+	}
+	u1, u2, v1, v2 := sd[0], sd[1], sd[2], sd[3]
+	isNegu1, isNegu2, isNegv1, isNegv2 := signs[0], signs[1], signs[2], signs[3]
+
+	// Check that: s*(v1 + λ*v2) + u1 + λ*u2 = 0
```
ivokub, line 600: Could use g2.fr.Eval perhaps? But can keep as is for brevity.

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -510,274 +541,276 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
+	_Q := Q
+	if cfg.CompleteArithmetic {
+		one := g2.Ext2.One()
+		// if Q=(0,0) we assign a dummy point
+		_selector0 = g2.api.And(g2.Ext2.IsZero(&Q.P.X), g2.Ext2.IsZero(&Q.P.Y))
+		_Q = g2.Select(_selector0, &G2Affine{P: g2AffP{X: *one, Y: *one}}, Q)
+		// if R.X == Q.X (happens when s=±1, so R=±Q), the incomplete addition fails
+		// We check this BEFORE potentially modifying R
+		_selector1 = g2.Ext2.IsZero(g2.Ext2.Sub(&Q.P.X, &R.P.X))
+		// if s=0/s=-1 (selector0), Q=(0,0) (_selector0), or R.X==Q.X (_selector1),
```
ivokub: The selectorAny is correct, but comment imo is incorrect -- selector0 handles case s=0 and _selector1 case where s=+-1. But I think we can comment keep as is.

std/algebra/emulated/sw_bls12381/g2_test.go
```diff
@@ -356,3 +355,74 @@ func TestMultiScalarMul(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("new G2 struct: %w", err)
+	}
+	res := g2.scalarMulGLVAndFakeGLV(&c.In, &c.S, algopts.WithCompleteArithmetic())
+	g2.AssertIsEqual(res, &c.Res)
+	return nil
+}
+
+// TestScalarMulG2EdgeCases tests edge cases: s=0, s=1, s=-1, Q=(0,0)
+func TestScalarMulG2EdgeCases(t *testing.T) {
```
ivokub, line 376: Doesn't test for Q=(0,0)

std/algebra/emulated/sw_bn254/g2_test.go
```diff
@@ -156,3 +160,130 @@ func TestEndomorphismG2TestSolve(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("new G2 struct: %w", err)
+	}
+	res := g2.scalarMulGLVAndFakeGLV(&c.In, &c.S, algopts.WithCompleteArithmetic())
+	g2.AssertIsEqual(res, &c.Res)
+	return nil
+}
+
+// TestScalarMulG2EdgeCases tests edge cases: s=0, s=1, s=-1, Q=(0,0)
+func TestScalarMulG2EdgeCases(t *testing.T) {
```
ivokub, line 237: Doesn't test for Q=(0,0)

std/algebra/emulated/sw_bw6761/g2_test.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return fmt.Errorf("new G2 struct: %w", err)
+	}
+	res := g2.scalarMulGLVAndFakeGLV(&c.In, &c.S, algopts.WithCompleteArithmetic())
+	g2.AssertIsEqual(res, &c.Res)
+	return nil
+}
+
+// TestScalarMulG2EdgeCases tests edge cases: s=0, s=1, s=-1, Q=(0,0)
+func TestScalarMulG2EdgeCases(t *testing.T) {
```
ivokub, line 90: Doesn't test for Q=(0,0)

std/algebra/emulated/sw_bn254/g2.go
```diff
@@ -290,3 +335,333 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
+	var _selector0, _selector1 frontend.Variable
+	_Q := Q
+	if cfg.CompleteArithmetic {
+		one := g2.Ext2.One()
+		// if Q=(0,0) we assign a dummy point
+		_selector0 = g2.api.And(g2.Ext2.IsZero(&Q.P.X), g2.Ext2.IsZero(&Q.P.Y))
+		_Q = g2.Select(_selector0, &G2Affine{P: g2AffP{X: *one, Y: *one}}, Q)
+		// if R.X == Q.X (happens when s=±1, so R=±Q), the incomplete addition fails
+		// We check this BEFORE potentially modifying R
+		_selector1 = g2.Ext2.IsZero(g2.Ext2.Sub(&Q.P.X, &R.P.X))
```
ivokub: Is this sound here? We choose _selector1 based on hinted `R`? Later on line 663 we switch to R back when `_selector1` is 1, but then we always trust the hinted (unconstrained) result.

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -510,274 +541,276 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
+	var _selector0, _selector1 frontend.Variable
+	_Q := Q
+	if cfg.CompleteArithmetic {
+		one := g2.Ext2.One()
+		// if Q=(0,0) we assign a dummy point
+		_selector0 = g2.api.And(g2.Ext2.IsZero(&Q.P.X), g2.Ext2.IsZero(&Q.P.Y))
+		_Q = g2.Select(_selector0, &G2Affine{P: g2AffP{X: *one, Y: *one}}, Q)
+		// if R.X == Q.X (happens when s=±1, so R=±Q), the incomplete addition fails
+		// We check this BEFORE potentially modifying R
+		_selector1 = g2.Ext2.IsZero(g2.Ext2.Sub(&Q.P.X, &R.P.X))
```
ivokub: Similar comment as for BN254, we set `_selector1` based on hinted result and later on line 810 switch back to it in case we have s=0 or Q=(0,0)

std/algebra/emulated/sw_bw6761/g2.go
```diff
@@ -236,3 +271,283 @@ func (g2 *G2) AssertIsEqual(p, q *G2Affine) {
...
+	var _selector0, _selector1 frontend.Variable
+	_Q := Q
+	if cfg.CompleteArithmetic {
+		one := g2.curveF.One()
+		// if Q=(0,0) we assign a dummy point
+		_selector0 = g2.api.And(g2.curveF.IsZero(&Q.P.X), g2.curveF.IsZero(&Q.P.Y))
+		_Q = g2.Select(_selector0, &G2Affine{P: g2AffP{X: *one, Y: *one}}, Q)
+		// if R.X == Q.X (happens when s=±1, so R=±Q), the incomplete addition fails
+		// We check this BEFORE potentially modifying R
+		_selector1 = g2.curveF.IsZero(g2.curveF.Sub(&Q.P.X, &R.P.X))
```
ivokub: Same issue with hinted R

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -1262,17 +1282,29 @@ func (c *Curve[B, S]) scalarMulFakeGLV(Q *AffinePoint[B], s *emulated.Element[S]
...
 
-	var selector2 frontend.Variable
-	one := c.baseApi.One()
-	dummy := &AffinePoint[B]{X: *one, Y: *one}
-	addFn := c.Add
+	// Handle Q=(0,0), s=0/s=-1, s=±1 (where R=±Q), and s=±3 (where R=±[3]Q)
+	// for complete arithmetic
+	var selector1, selector2, selector3 frontend.Variable
+	_Q := Q
 	if cfg.CompleteArithmetic {
```
ivokub, line 1304: Hmm - I remember for some reason the change from `c.Add` to switched `c.Add/c.AddUnified` was important and it allowed to avoid some edge cases.

Here again we seem to rely on the hinted result `R` for selecting and dummy point addition but imo it was not sound. I cannot pinpoint right now the soundness issues, but I have a strange feeling. I'll need to think about it a bit more.

ivokub (commented):
I have reviewed the emulated cases, still reviewing 2-chains. I'm posting my comments for now. I'm not confident that the changes are correct, particularly we seem to trust the hinted scalar mul result before we constrain it. And later we also may switch back to hinted result without constraining for particular edge cases (scalar=0 for example).

std/algebra/emulated/sw_bn254/g2.go
ivokub: I was able to create a POC in g2_test.go
```go
func TestG2ScalarMulMalicious(t *testing.T) {
	assert := test.NewAssert(t)
	_, _, _, gen := bn254.Generators()

	// Test case: s = 1. The malicious hint returns a point with the same X as Q
	// but a different Y, which triggers the R.X == Q.X edge-case path and skips
	// the in-circuit verification of the hinted scalar multiplication result.
	var s fr.Element
	s.SetOne()
	var res bn254.G2Affine
	res.Set(&gen)
	res.Y.A1.SetInt64(456)

	witness := scalarMulG2CompleteCircuit{
		In:  NewG2Affine(gen),
		S:   NewScalar(s),
		Res: NewG2Affine(res),
	}
	err := test.IsSolved(&scalarMulG2CompleteCircuit{}, &witness, ecc.BN254.ScalarField(),
		test.WithReplacementHint(solver.GetHintID(scalarMulG2Hint), maliciousG2ScalarMulHint))
	assert.NoError(err)
}

func maliciousG2ScalarMulHint(field *big.Int, inputs []*big.Int, outputs []*big.Int) error {
	return emulated.UnwrapHintContext(field, inputs, outputs, func(hc emulated.HintContext) error {
		moduli := hc.EmulatedModuli()
		if len(moduli) != 2 {
			return fmt.Errorf("expecting two moduli, got %d", len(moduli))
		}
		baseModulus, scalarModulus := moduli[0], moduli[1]
		baseInputs, baseOutputs := hc.InputsOutputs(baseModulus)
		scalarInputs, _ := hc.InputsOutputs(scalarModulus)
		if len(baseInputs) != 4 {
			return fmt.Errorf("expecting four base inputs (Q.X.A0, Q.X.A1, Q.Y.A0, Q.Y.A1), got %d", len(baseInputs))
		}
		if len(baseOutputs) != 4 {
			return fmt.Errorf("expecting four base outputs, got %d", len(baseOutputs))
		}
		if len(scalarInputs) != 1 {
			return fmt.Errorf("expecting one scalar input, got %d", len(scalarInputs))
		}

		// compute the resulting point [s]Q on G2
		var Q bn254.G2Affine
		Q.X.A0.SetBigInt(baseInputs[0])
		Q.X.A1.SetBigInt(baseInputs[1])
		Q.Y.A0.SetBigInt(baseInputs[2])
		Q.Y.A1.SetBigInt(baseInputs[3])
		Q.ScalarMultiplication(&Q, scalarInputs[0])
		Q.X.A0.BigInt(baseOutputs[0])
		Q.X.A1.BigInt(baseOutputs[1])
		Q.Y.A0.BigInt(baseOutputs[2])
		Q.Y.A1.BigInt(baseOutputs[3])

		// Make the hint malicious when s=1: keep the same X-coordinate so the
		// complete-arithmetic path detects R.X == Q.X and skips the MSM check,
		// but change Y so the returned point is not [1]Q.
		if scalarInputs[0].Cmp(big.NewInt(1)) == 0 {
			baseOutputs[0].Set(baseInputs[0])
			baseOutputs[1].Set(baseInputs[1])
			baseOutputs[2].Set(baseInputs[2])
			baseOutputs[3].SetInt64(456)
		}
		return nil
	})
}
```
requires changes in the test engine in #1737. I was able to create similar POC attacks also for other curves.

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -1229,18 +1247,20 @@ func (c *Curve[B, S]) scalarMulFakeGLV(Q *AffinePoint[B], s *emulated.Element[S]
...
-	var selector1 frontend.Variable
+	// Handle edge cases for complete arithmetic: s=0, Q=(0,0)
+	var selector0 frontend.Variable
 	_s := s
 	if cfg.CompleteArithmetic {
-		selector1 = c.scalarApi.IsZero(s)
-		_s = c.scalarApi.Select(selector1, c.scalarApi.One(), s)
+		one := c.scalarApi.One()
+		selector0 = c.scalarApi.IsZero(s)
+		_s = c.scalarApi.Select(selector0, one, s)
```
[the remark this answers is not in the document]
yelhousni: false positive. the rational reconstruction produces correct sub-scalars s1, s2. the case -1 handle previously was for a different reason to avoid incomplete add, now fixed with selector2.

std/algebra/native/twistededwards/emulatedparams.go
```diff
@@ -0,0 +1,64 @@
...
+
+package twistededwards
+
+import "math/big"
+
+// Emulated field parameters for twisted Edwards curve orders.
+// These are used for overflow-safe scalar decomposition verification.
+
+// edBN254Order is the BabyJubjub curve order (251 bits).
+type edBN254Order struct{}
```
ivokub, line 12: Should move the parameters to `std/math/emulated/emparams`?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -1229,18 +1247,20 @@ func (c *Curve[B, S]) scalarMulFakeGLV(Q *AffinePoint[B], s *emulated.Element[S]
 		panic(err)
```
ivokub, line 1251: This doesn't still handle all edge cases in case we have `WithCompleteArithmetic` set. See the test:
```go
func TestScalarMulFakeGLVUnsafeS1Fails(t *testing.T) {
	assert := test.NewAssert(t)
	p256 := elliptic.P256()

	s := big.NewInt(1)
	px, py := p256.Params().Gx, p256.Params().Gy

	unsafeCircuit := ScalarMulFakeGLVTest[emulated.P256Fp, emulated.P256Fr]{}
	completeCircuit := ScalarMulFakeGLVEdgeCasesTest[emulated.P256Fp, emulated.P256Fr]{}

	witness := ScalarMulFakeGLVTest[emulated.P256Fp, emulated.P256Fr]{
		S: emulated.ValueOf[emulated.P256Fr](s),
		Q: AffinePoint[emulated.P256Fp]{
			X: emulated.ValueOf[emulated.P256Fp](px),
			Y: emulated.ValueOf[emulated.P256Fp](py),
		},
		R: AffinePoint[emulated.P256Fp]{
			X: emulated.ValueOf[emulated.P256Fp](px),
			Y: emulated.ValueOf[emulated.P256Fp](py),
		},
	}

	err := test.IsSolved(&unsafeCircuit, &witness, testCurve.ScalarField())
	assert.Error(err)

	completeWitness := ScalarMulFakeGLVEdgeCasesTest[emulated.P256Fp, emulated.P256Fr]{
		S: witness.S,
		P: witness.Q,
		R: witness.R,
	}
	err = test.IsSolved(&completeCircuit, &completeWitness, testCurve.ScalarField())
	assert.NoError(err)
}
```

I think we should revert to using complete arithmetic in case the option is set instead of using dummy point to shift. As the dummy point is known beforehand, then it would otherwise always be possible to create an edge case imo which leads to incomplete circuit.

ivokub (changes requested):
There is still the completeness issue for sw_emulated.scalarMulFakeGLV

yelhousni:
Close in favor of #1763

PR #1699, opened 2026-02-02
perf: optimize range checks for small number of small field ops

### LINKED ISSUE #1698: Perf: optimize small field range check optimizations for small circuits
> Forcing base length 16 creates a lookup table with 65,536 entries. This table cost dominates for circuits with fewer than ~50,000 range checks.
> ```  ┌──────────────┬──────────────┬──────────────┬──────────────┬─────────────┐
>   │ Circuit Size │ Optimal Base │ Optimal Cost │ Base-16 Cost │    Ratio    │
>   ├──────────────┼──────────────┼──────────────┼──────────────┼─────────────┤
>   │ 100 muls     │ 7            │ 829          │ 65,837       │ 79x worse   │
>   ├──────────────┼──────────────┼──────────────┼──────────────┼─────────────┤
>   │ 1,000 muls   │ 8            │ 6,257        │ 68,537       │ 11x worse   │
>   ├──────────────┼──────────────┼──────────────┼──────────────┼─────────────┤
>   │ 10,000 muls  │ 11           │ 52,049       │ 95,537       │ 1.8x worse  │
>   ├──────────────┼──────────────┼──────────────┼──────────────┼─────────────┤
>   │ 100,000 muls │ 16           │ 465,537      │ 365,537      │ 1.3x better │
>   └──────────────┴──────────────┴──────────────┴──────────────┴─────────────┘
> ```
> 
> Add a threshold check that only forces base-16 when range check count exceeds ~50,000 ?
> 
> 

Yup, this is the trade off. I'll think how to approach this as the base length defines how much overflow we allow in small field optimisations, but it is defined every time we define non native operations and don't know how many range checks there is going to be.

I'll think about a little how to best approach it and do some changes. Perhaps we could just allow known base lengths 8 and 16 only. Then the tricks still keep working. And I think this could actually allow me to remove the range checker option. It would still occasionally be suboptimal, but fewer times.

PS, for large field emulation this doesn't change behaviour, we only use it for small field case.

_Originally posted by @ivokub in https://github.com/Consensys/gnark/pull/1693#issuecomment-3824638433_

ivokub:
# Description

Additionally, sets MulNoReduce as default when calling `Field.Mul` on small field parameter.

Fixes #1698 

## Type of change

# How has this been tested?

- [ ] Test A
- [ ] Test B

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes `Field.Mul` semantics for single-limb emulated fields by defaulting to non-reducing multiplication, which can affect callers that assumed `Mul` always reduced (now must use `MulMod` for that guarantee). Adds stateful range-check deduplication and a threshold-triggered switch in range-checking behavior, which impacts constraint generation and overflow handling in hot paths.
> 
> **Overview**
> **Optimizes emulated-field range checks and small-field multiplication.** Range checks now go through a new `Field.rangeCheck` helper that deduplicates checks per limb (tracking the *widest* enforced bit-width) and counts how many checks have been emitted.
> 
> For small-field emulation, range checking switches *only after* `thresholdForInexactOverflow` checks to a base-length (16-bit) range checker, and `smallAdditionalOverflow()` now returns non-zero only after that threshold to keep overflow accounting consistent.
> 
> `Field.Mul` now defaults to `MulNoReduce` when `useSmallFieldOptimization()` is enabled, falling back to reducing multiplication only when overflow would exceed limits; callers needing guaranteed modular reduction should use `MulMod`. Tests/benchmarks are adjusted accordingly (bench compile loop guard + suppress `ns/op`, and malicious-hint circuit uses `MulMod`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c99474561bed15500164695e1f32688dcbd472f4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/std/math/emulated/field.go
+++ b/std/math/emulated/field.go
@@ -24,9 +24,9 @@ const (
 	// the base length only when the number of range checks exceeds
 	// thresholdOptimizeOptimizedOverflow.
 	rangeCheckBaseLengthForSmallField = 16
-	// thresholdOptimizeOptimizedOverflow is the number of range checks after
+	// thresholdForInexactOverflow is the number of range checks after
 	// which we start enforcing the base length for small field optimization.
-	thresholdOptimizeOptimizedOverflow = 55000
+	thresholdForInexactOverflow = 55000
 )
 
 // Field holds the configuration for non-native field operations. The field
@@ -421,7 +421,7 @@ func (f *Field[T]) rangeCheck(v frontend.Variable, nbBits int) bool {
 	// emulating small field.
 	f.nbRangeChecks++
 
-	if f.nbRangeChecks == thresholdOptimizeOptimizedOverflow {
+	if f.nbRangeChecks == thresholdForInexactOverflow {
 		// the threshold is reached, set the range checker to use base length.
 		// Now we know that when constructing non-native elements, then we should
 		// set overflow=f.smallAdditionalOverflow()
--- a/std/math/emulated/field_smallmul.go
+++ b/std/math/emulated/field_smallmul.go
@@ -420,7 +420,7 @@ func (f *Field[T]) smallAdditionalOverflow() int {
 	}
 	// when we haven't performed too many range checks, then we still use exact
 	// overflow
-	if f.nbRangeChecks < thresholdOptimizeOptimizedOverflow {
+	if f.nbRangeChecks < thresholdForInexactOverflow {
 		return 0
 	}
 	// otherwise, we use the additional overflow which reduced number of


std/math/emulated/field.go
```diff
@@ -19,7 +19,14 @@ import (
...
 
 const (
+	// rangeCheckBaseLengthForSmallField is the base length used for range
+	// checking when using small field optimization. We start enforcing
+	// the base length only when the number of range checks exceeds
+	// thresholdOptimizeOptimizedOverflow.
 	rangeCheckBaseLengthForSmallField = 16
+	// thresholdOptimizeOptimizedOverflow is the number of range checks after
+	// which we start enforcing the base length for small field optimization.
+	thresholdOptimizeOptimizedOverflow = 50000
```
YaoJGalteland: How is the threshold determined? I’ve found two values and was wondering if they seem reasonable to you.

**R1CS Backend - Detailed Threshold View**

| Num Checks | Optimal Base | Optimal Cost | 16-bit Cost | Diff | 16-bit Optimal? |
|------------|--------------|--------------|-------------|------|-----------------|
| 40,000 | 11 | 202,049 | 225,537 | 23,488 | no |
| 45,000 | 11 | 227,049 | 245,537 | 18,488 | no |
| 50,000 | 11 | 252,049 | 265,537 | 13,488 | no |
| 55,000 | 11 | 277,049 | 285,537 | 8,488 | no |
| 60,000 | 11 | 302,049 | 305,537 | 3,488 | no |
| 60,490 | 11 | 304,499 | 307,497 | 2,998 | no |
| 60,990 | 11 | 306,999 | 309,497 | 2,498 | no |
| 61,490 | 11 | 309,499 | 311,497 | 1,998 | no |
| 61,990 | 11 | 311,999 | 313,497 | 1,498 | no |
| 62,490 | 11 | 314,499 | 315,497 | 998 | no |
| 62,990 | 11 | 316,999 | 317,497 | 498 | no |
| **63,490** | **16** | **319,497** | **319,497** | **0** | **YES ✓ THRESHOLD** |
| 63,990 | 16 | 321,497 | 321,497 | 0 | YES ✓ |
| 64,490 | 16 | 323,497 | 323,497 | 0 | YES ✓ |
| 64,990 | 16 | 325,497 | 325,497 | 0 | YES ✓ |
| 65,000 | 16 | 325,537 | 325,537 | 0 | YES ✓ |
| 70,000 | 16 | 345,537 | 345,537 | 0 | YES ✓ |
| 75,000 | 16 | 365,537 | 365,537 | 0 | YES ✓ |


**PLONK Backend - Detailed Threshold View**

| Num Checks | Optimal Base | Optimal Cost | 16-bit Cost | Diff | 16-bit Optimal? |
|------------|--------------|--------------|-------------|------|-----------------|
| 30,000 | 11 | 486,145 | 556,609 | 70,464 | no |
| 35,000 | 11 | 566,145 | 616,609 | 50,464 | no |
| 40,000 | 11 | 646,145 | 676,609 | 30,464 | no |
| 44,620 | 11 | 720,065 | 732,049 | 11,984 | no |
| 45,000 | 11 | 726,145 | 736,609 | 10,464 | no |
| 45,620 | 11 | 736,065 | 744,049 | 7,984 | no |
| 46,120 | 11 | 744,065 | 750,049 | 5,984 | no |
| 46,620 | 11 | 752,065 | 756,049 | 3,984 | no |
| 47,120 | 11 | 760,065 | 762,049 | 1,984 | no |
| **47,620** | **16** | **768,049** | **768,049** | **0** | **YES ✓ THRESHOLD** |
| 48,120 | 16 | 774,049 | 774,049 | 0 | YES ✓ |
| 48,620 | 16 | 780,049 | 780,049 | 0 | YES ✓ |
| 49,120 | 16 | 786,049 | 786,049 | 0 | YES ✓ |
| 50,000 | 16 | 796,609 | 796,609 | 0 | YES ✓ |
| 55,000 | 16 | 856,609 | 856,609 | 0 | YES ✓ |
| 60,000 | 16 | 916,609 | 916,609 | 0 | YES ✓ |

YaoJGalteland: `thresholdOptimizeOptimizedOverflow`: the name is confusing, double **Optimize**, and why is it called **Overflow**?

std/math/emulated/field_mul.go
```diff
@@ -625,6 +632,12 @@ func (f *Field[T]) Mul(a, b *Element[T]) *Element[T] {
 	if a.isStrictZero() || b.isStrictZero() {
 		return f.Zero()
 	}
```
YaoJGalteland, line 634:
```
result := f.Mul(&c.A, &c.B)
f.AssertIsLessOrEqual(result, &c.C)
```
for above test 
the `AssertIsLessOrEqual` will panic with                                 

> first input must have 0 overflow

the result.overflow is 32

YaoJGalteland: Similarly, AssertIsInRange calls AssertIsLessOrEqual, so it will also panic in the following test

```
result := f.Mul(&c.A, &c.B)
f.AssertIsInRange(result)
```

YaoJGalteland: if we call ToBits after Mul, there will be 31+32 bits, 2X constraints cost compared to the reduced mul value

std/math/emulated/field.go
ivokub: I found it by trial and error and to have a nice round number.

So I want to balance between two goals:
* because the optimizations for the overflow kick in when number of multiplications > threshold, then I want the threshold to be small enough that I would do least amount of multiplications without applying the optimizations
* meanwhile, I also do not want to have too much overhead for building the table (i.e. table size compared to the optimal size). So the threshold should be as close to the optimal value.

Now the question is that the optimal threshold is at 63490 for SCS and 47620 for R1CS. When the threshold is at 50k, then it means the overhead for table building is approx 13488/252049 ~ 5.3%. If I would increase it to 55000 then it would be 8488/277049 ~ 3% which is reasonable imo. Imo moving it exactly to 62k doesn't make sense as it is very small range where we have suboptimal size and the overhead is also negligible. For R1CS it is already optimal.

ivokub: Yeah, not reasonable name. Essentially I wanted to say that this is the threshold after which we start applying the optimization which uses inexact overflow. I renamed, perhaps it is better. But it is anyway commented and internal constant, so perhaps now is good enough?

std/math/emulated/field_mul.go
ivokub: Imo it is working as intended - `AssertIsLessOrEqual` assumes that the inputs are already reduced. When we do a reduction then we can do it two ways:
* we just check that the width of the limb is right (`Reduce` method). But this would allow to have a small overflow (i.e. lets say modulus is 0xf000, then we may have 0xf001 as well). In this case AssertIsLessOrEqual fails the assertion (but it is given in the documentation that the user should use `ReduceStrict`)
* we check strictly that the limb is less than the modulus (`ReduceStrict` method). But this is very expensive, we only use it when we absolutely need to (i.e. we do `ToBitsCanonical` etc.)

And this behaviour is not new - this also holds when we emulate large field. Just there the default multiplication is reducing. But if we do
```
result := f.MulNoReduce(&c.A, &c.B)
f.AssertIsLessOrEqual(result, &c.C)
```

And it is the same for `ToBits` - if we want canonical representation we should use `ToBitsCanonical`. And it has exactly the expected number of bits.

PR #1700, opened 2026-02-05
perf: windowed-4 ModExp precompile

yelhousni:
# Description

Optimizes the `ModExp` function in the emulated arithmetic package by replacing the binary square-and-multiply algorithm with **windowed exponentiation** (window size 4).

The original implementation performed ~8190 `ModMul` operations for a 4096-bit exponent (2 multiplications per bit: 1 squaring + 1 conditional multiply). The new implementation reduces this to ~5135 `ModMul` operations by:

1. Precomputing a table of 16 values: `base^0, base^1, ..., base^15` (15 multiplications)
2. Processing the exponent in 4-bit windows from MSB to LSB
3. For each window: 4 squarings + 1 table lookup multiplication
4. Using an efficient binary tree selection for table lookups

This results in a **33% reduction in constraint count** for the MODEXP precompile circuit.

## Type of change

# How has this been tested?

- [x] All existing `TestRandomInstance` tests pass (256, 512, 1024, 2048, 4096 bit inputs)
- [x] All existing `TestEdgeCases` tests pass (0^0, mod 0, mod 1, etc.)
- [x] `TestVariableExp` in emulated package passes
- [x] New `TestVariableExpEdgeCases` added to `custommod_test.go`:
  - `2^10 mod 1000` (small exponent)
  - `2^0 mod 1000` (zero exponent)
  - `0^10 mod 1000` (zero base)
  - `1^12345 mod 1000` (base = 1)
  - `7^256 mod 13` (small modulus)
  - Large base (2^64 - 1)
  - Large exponent (2^32)
  - Large modulus (2^256)

SCS constraint count comparison for MODEXP with `Mod1e4096` (4096-bit inputs), compiled over BN254 scalar field:

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| SCS Constraints | 41,311,129 | 27,552,166 | **33% reduction** |

Window size analysis (SCS constraints for Mod1e4096 on BN254):
- Window size 3: 28,281,065
- Window size 4: 27,552,166 (optimal)
- Window size 5: 28,251,179

Window size 4 provides the best tradeoff between ModMul savings and table lookup overhead.

# Checklist:

- [x] `golangci-lint` does not output errors locally (pre-existing warnings in other files)

---

> [!NOTE]
> **Medium Risk**
> Touches core arithmetic used to build circuit constraints; logic changes are performance-motivated but any subtle bug in window processing or lookup selection could break proof correctness across edge cases.
> 
> **Overview**
> **Optimizes non-native exponentiation** by replacing bit-by-bit square-and-multiply with 4-bit windowed exponentiation in both `Field.Exp` (fixed modulus) and `Field.ModExp` (variable modulus), adding a 16-entry precomputation table and per-window squaring + multiply flow.
> 
> Adds a `tableLookup` helper that selects `base^window` via a 2-level `Lookup2` tree, and expands `custommod_test.go` with new variable-modulus exponentiation edge-case coverage (zero/one inputs, small/large exponents and moduli).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7d785743002a7fcc83a41a0b6b080113a7132591. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/std/math/emulated/field_mul.go
+++ b/std/math/emulated/field_mul.go
@@ -780,7 +780,7 @@ func (f *Field[T]) Exp(base, exp *Element[T]) *Element[T] {
 	table[0] = f.One()
 	table[1] = base
 	for i := 2; i < tableSize; i++ {
-		table[i] = f.Mul(table[i-1], base)
+		table[i] = f.MulMod(table[i-1], base)
 	}
 
 	// Get exponent bits (LSB first)


std/math/emulated/custommod.go
```diff
@@ -96,19 +96,97 @@ func (f *Field[T]) ModAssertIsEqual(a, b *Element[T], modulus *Element[T]) {
...
+	// Extract MSB window bits (with padding of zeros)
+	msbWindowBits := make([]frontend.Variable, windowSize)
+	for i := 0; i < windowSize; i++ {
+		bitIdx := n - 1 - i // Start from MSB
+		if bitIdx >= 0 {
+			msbWindowBits[windowSize-1-i] = expBts[bitIdx]
+		} else {
+			msbWindowBits[windowSize-1-i] = 0
+		}
+	}
```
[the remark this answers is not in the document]
yelhousni: hmm this bug is algorithmically valid, but is somewhat not practical in the current settings. The bug would manifest if:
1. `n % 4 ≠ 0` (i.e., `ToBits` returns a non-multiple-of-4 bits)
2. The bits just below MSB are non-zero

This is  Currently masked because:
- `n = BitsPerLimb × NbLimbs`
- `BitsPerLimb = 64` (multiple of 4)
- Therefore n is always divisible by 4

To Actually Trigger the bug we'd need a field parameter where `BitsPerLimb × NbLimbs` is not a multiple of 4, which doesn't exist in the current `emparams`.

It should be fixed anyway 👍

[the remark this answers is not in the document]
yelhousni: We should use `msbBaseIdx = (numWindows - 1) * windowSize` and extracts bits in LSB-first order, with padding zeros for indices `>= n`

std/math/emulated/custommod.go
```diff
@@ -96,19 +96,99 @@ func (f *Field[T]) ModAssertIsEqual(a, b *Element[T], modulus *Element[T]) {
...
-		res = f.Select(expBts[i], f.ModMul(base, res, modulus), res)
-		base = f.ModMul(base, base, modulus)
+
+	// Pad to multiple of windowSize
+	padding := (windowSize - (n % windowSize)) % windowSize
+	paddedLen := n + padding
+
+	// Process windows from MSB to LSB
+	// expBts is LSB-first, so expBts[n-1] is MSB
+	numWindows := paddedLen / windowSize
```
ivokub, line 129: I usually use
```
numWindows := (n + windowSize-1)/windowSize
```
to get the number of whole parts. But just a remark, no need to change, the explicit approach is better actually.

std/math/emulated/custommod.go
```diff
@@ -96,19 +96,99 @@ func (f *Field[T]) ModAssertIsEqual(a, b *Element[T], modulus *Element[T]) {
...
+		}
+
+		// Extract window bits for this window
+		// Window w covers bits from position (numWindows-1-w)*windowSize to (numWindows-w)*windowSize - 1
+		// In the original LSB-first array
+		windowBits := make([]frontend.Variable, windowSize)
+		baseIdx := (numWindows - 1 - w) * windowSize
+		for i := 0; i < windowSize; i++ {
+			actualIdx := baseIdx + i
+			if actualIdx < n && actualIdx >= 0 {
```
ivokub: Hmm, I'm trying to follow, but do we have case where in this loop we would have padding? I think we would only have 0 padding in the first window, no?

ivokub (approved):
Thanks for the PR! I think it is really nice optimisation. The implementation is correct and the tests are well covered. There is only a if condition in the non-MSB iteration (because we have already covered the padding bits for the MSB window of exponent) which is always true, so we could remove and simplify it.

I think in principle the same idea would also work for the `Field.Exp` method in `field_mul.go` package as well. Imo would make sense to implement for it as well? It doesn't affect the precompiles directly, but I think it is still used. I think we manually defined exponentiation in PLONK recursive verifier, but it was manual as the number of bits was bounded and currently our interfaces takes as input `*Element[T]` as exponent which decomposes into too many bits (for example when the exponent is bounded at 25 bits etc.)

std/math/emulated/custommod.go
ivokub: I checked and imo it is indeed the case. We can simplify the part from lines 160-165 as

windowBits[i] = expBts[ActualIdx]

std/math/emulated/custommod.go
```diff
@@ -96,19 +96,99 @@ func (f *Field[T]) ModAssertIsEqual(a, b *Element[T], modulus *Element[T]) {
...
+	// We need to select table[b0 + 2*b1 + 4*b2 + ...]
+	current := make([]*Element[T], len(table))
+	copy(current, table)
+
+	for _, bit := range bits {
+		half := len(current) / 2
+		next := make([]*Element[T], half)
+		for j := 0; j < half; j++ {
+			// If bit=0, take current[2j], if bit=1, take current[2j+1]
+			next[j] = f.Select(bit, current[2*j+1], current[2*j])
```
ivokub: We could probably also use Lookup2 or Mux. But relatively it is very small saving and I think using Select for simplicity makes sense. But if you want to try out Mux as well then why not.

yelhousni: I tried with Mux but the binary tree Select approach was ~1.4% more efficient in SCS constraint

yelhousni: Actually using 5 Lookup2 calls instead of 15 Select calls saves ~3,000 constraints for 512-bit ModExp (~0.5%)

std/math/emulated/custommod.go
```diff
@@ -172,23 +167,21 @@ func (f *Field[T]) ModExp(base, exp, modulus *Element[T]) *Element[T] {
...
-
-	for _, bit := range bits {
-		half := len(current) / 2
-		next := make([]*Element[T], half)
-		for j := 0; j < half; j++ {
-			// If bit=0, take current[2j], if bit=1, take current[2j+1]
-			next[j] = f.Select(bit, current[2*j+1], current[2*j])
-		}
-		current = next
+	// For 4 bits selecting from 16 elements:
```
ivokub, line 177: I would add the `len(table) == 16` and `len(bits) == 4` check explicitly here. It is mentioned in the method documentation, but the check is in principle free, so doesn't hurt at least.

std/math/emulated/field_mul.go
```diff
@@ -763,20 +763,74 @@ func (f *Field[T]) mulNoReduce(a, b *Element[T], nextoverflow uint) *Element[T]
...
+
+	const windowSize = 4
+	const tableSize = 1 << windowSize // 16
+
+	// Build precomputation table: table[i] = base^i for i in [0, 2^windowSize)
+	table := make([]*Element[T], tableSize)
+	table[0] = f.One()
+	table[1] = base
+	for i := 2; i < tableSize; i++ {
+		table[i] = f.Mul(table[i-1], base)
```
ivokub: I would do explicitly `MulMod` here - we have different default operations when we emulate large or small field:
- for large field `Mul == MulMod` as it is heuristically more efficient except for very specific use cases (inner product computation)
- for small field `Mul == MulNoReduce` as it is more efficient (because we operate on a single limb and don't get the exponential blowup when doing lazy modular reduction)

Now the difference is that Select, Lookup2, Mux take the maximum overflow of the inputs, so in case we emulate small fields we would be doing early reductions in the `Mul` below. I'll try out and comment.

ivokub: Yep, seems to help a bit. In `TestExp` for the case emulated Goldilocks over BN254 it reduces the constraints from 8140 to 7687.

std/math/emulated/custommod.go
ivokub: Added the check.

std/math/emulated/field_mul.go
ivokub: But this is relevant only for `Field.Exp` (fixed modulus), not for `Field.ModExp` (variable modulus) as in case of variable modulus we don't have the "small parametrization optimizations" defined (i.e. Mod1e64 or Mod1e32).

PR #1704, opened 2026-02-09
perf(Mux): shared selector decomposition and ±Y symmetry optimization in emulated MSM

yelhousni:
# Description

Optimizes the `Mux` (multiplexer) function in the selector package and its usage across emulated field arithmetic and elliptic curve scalar multiplication.

The optimizations consist of three key improvements:

## 1. Share binary decomposition across emulated field limbs

Previously, `emulated.Mux` decomposed the selector into bits **independently for each limb** (e.g., 6 limbs × 4 bits = 24 redundant boolean assertions for a 16-to-1 Mux). Now we decompose once and reuse across all limbs.

- Added `BinaryMuxUnchecked` and `GeneralMuxUnchecked` variants that skip boolean assertions when bits are already constrained (e.g., from `bits.ToBinary`).

Also:
- Add constant selector folding: return `inputs[sel]` directly when `sel` is compile-time constant
- Add n=2 specialization: use `api.Select` directly for 2-to-1 Mux

**Files:** `std/math/emulated/field_ops.go`, `std/selector/mux.go` and `std/selector/multiplexer.go`.

## 2. Exploit ±Y symmetry in MSM

In GLV-based scalar multiplication, the 16 precomputed points satisfy `Y[i] = -Y[15-i]`. This allows replacing a **16-to-1 Mux** with an **8-to-1 Mux + conditional negation**, saving constraints.

Added `muxY8Signed` helper that selects from 8 Y-values and conditionally negates based on the MSB.

**Files:** `std/algebra/emulated/sw_emulated/point.go`, `std/algebra/emulated/sw_bls12381/g2.go`

## Type of change

# How has this been tested?

- [x] All existing `std/selector` tests pass
- [x] All existing `std/math/emulated` tests pass
- [x] All existing `std/algebra/emulated/sw_emulated` tests pass
- [x] All existing `std/algebra/emulated/sw_bls12381` tests pass
- [x] All existing `std/evmprecompiles` tests pass

SCS constraint count comparison:

## Single Scalar Multiplication

| Circuit | Before | After | Improvement |
|---------|--------|-------|-------------|
| BLS12-381 G1.ScalarMul | 291,552 | 279,576 | **-4.1%** |
| BLS12-381 G2.ScalarMul | 913,513 | 891,913 | **-2.4%** |

## EVM Precompiles

| Circuit | Before | After | Improvement |
|---------|--------|-------|-------------|
| BN254 ECMul | 210,369 | 202,669 | **-3.7%** |
| BLS12-381 G1MSM(10) | 4,397,157 | 4,277,397 | **-2.7%** |
| BLS12-381 G2MSM(10) | 9,304,517 | 9,088,517 | **-2.3%** |
| P256 Verify | 666,146 | 652,482 | **-2.1%** |

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches low-level constraint-building and mux selection logic used broadly across circuits; while changes are performance-oriented, mistakes could cause unsound or failing constraints across many call sites.
> 
> **Overview**
> Speeds up constraint generation for multiplexers and emulated scalar multiplication by reducing redundant selector constraints and leveraging GLV point symmetry.
> 
> `selector.Mux` now short-circuits constant selectors, uses `api.Select` for 2-way muxes, and replaces the previous bounded-compare check with a builder-level `MustBeLessOrEqCst` on the already-decomposed selector bits. `emulated.Field.Mux` is reworked to decompose `sel` into bits once, reuse them across all limbs via `selector.BinaryMux`, and pad/check bounds only for non-power-of-two input counts.
> 
> Emulated GLV scalar-mul paths in `sw_emulated/point.go` and BLS12-381 `G2` replace 16-to-1 Y-coordinate muxing with an 8-to-1 mux plus conditional negation (`muxY8Signed`/`muxE2Y8Signed`), reflecting the `Y[i] = -Y[15-i]` symmetry. Constraint builders (`r1cs`/`scs`) also skip adding trivially satisfied `MustBeLessOrEqCst` constraints when the compared bit is a constant zero; `latest_stats.csv` is updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de2cef3b3a76ad619acaa25c480dcb728969c8ed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


std/selector/multiplexer.go
```diff
@@ -63,19 +63,35 @@ func Mux(api frontend.API, sel frontend.Variable, inputs ...frontend.Variable) f
 		api.AssertIsEqual(sel, 0)
 		return inputs[0]
 	}
+
+	// Fast path: if selector is a constant, return the selected input directly
+	if s, ok := api.Compiler().ConstantValue(sel); ok {
+		idx := int(s.Int64())
+		if idx < 0 || idx >= len(inputs) {
+			panic(fmt.Sprintf("constant selector %d out of bounds [0, %d)", idx, len(inputs)))
+		}
```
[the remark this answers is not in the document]
yelhousni: Practically not an issue. If a constant selector is larger than MaxInt64, the number of inputs would need to be astronomically large (more than 9 quintillion inputs), which is impossible. The check idx >= len(inputs) would catch any realistic overflow.

std/selector/multiplexer.go
```diff
@@ -63,19 +63,35 @@ func Mux(api frontend.API, sel frontend.Variable, inputs ...frontend.Variable) f
...
+		return inputs[idx]
+	}
+
+	// Special case for n=2: use Select directly (most efficient)
+	if n == 2 {
```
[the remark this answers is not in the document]
yelhousni: This is a valid concern but likely not a real bug in practice. The existing `Mux` documentation states that `sel` should be in `[0, n-1]`. For `n=2`, the caller is responsible for providing a valid selector. However, for consistency with the rest of `Mux` (which does constrain via `bits.ToBinary`), adding `api.AssertIsBoolean(sel)` would maybe be safer?

std/selector/mux.go
```diff
@@ -44,3 +44,22 @@ func binaryMuxRecursive(api frontend.API, selBits, inputs []frontend.Variable) f
 	right := binaryMuxRecursive(api, nextSelBits, inputs[pivot:])
 	return api.Add(left, api.Mul(msb, api.Sub(right, left)))
 }
+
+// BinaryMuxUnchecked is like BinaryMux but assumes selBits are already
+// constrained to be boolean (e.g., from bits.ToBinary). This avoids redundant
+// boolean assertions when the same bits are reused across multiple Mux calls.
+// len(inputs) must be 2^len(selBits).
+func BinaryMuxUnchecked(api frontend.API, selBits, inputs []frontend.Variable) frontend.Variable {
```
ivokub: I'm not sure we need the variant. Imo the `AssertIsBoolean` method caches the variables we assert boolean (see also `MarkBoolean` method in `api.Compiler()`). And imo when we have savings by omitting the boolean assertions then it means that from the input we get something which does not have `MarkBoolean` done. In that case imo it is correct to fix it on the caller side already.

std/selector/mux.go
```diff
@@ -44,3 +44,22 @@ func binaryMuxRecursive(api frontend.API, selBits, inputs []frontend.Variable) f
...
+		panic(fmt.Sprintf("invalid input length for BinaryMuxUnchecked (%d != 2^%d)", len(inputs), len(selBits)))
+	}
+	return binaryMuxRecursive(api, selBits, inputs)
+}
+
+// GeneralMuxUnchecked is a multiplexer that handles any number of inputs
+// (not just powers of 2). It assumes selBits are already constrained to be
+// boolean and that sel < len(inputs) has been verified by the caller.
+// This is useful when the same selector bits are reused across multiple Mux calls.
+func GeneralMuxUnchecked(api frontend.API, selBits, inputs []frontend.Variable) frontend.Variable {
```
ivokub: Same comment as for `BinaryMuxUnchecked` -- I'm not sure we need it, better fix on the calling side to mark variables as Booleans if we know they are.

std/math/emulated/field_ops.go
```diff
@@ -341,9 +348,29 @@ func (f *Field[T]) Mux(sel frontend.Variable, inputs ...*Element[T]) *Element[T]
...
+
+	if bits.OnesCount(n) == 1 {
+		// Power of 2: binary decomposition guarantees sel < 2^nbBits = n
+		for i := range nbLimbs {
+			e.Limbs[i] = selector.BinaryMuxUnchecked(f.api, selBits, normLimbsTransposed[i])
+		}
+	} else {
+		// Non-power of 2: need additional bound check sel <= n-1
+		bcmp := cmp.NewBoundedComparator(f.api, big.NewInt((1<<nbBits)-1), false)
+		bcmp.AssertIsLessEq(sel, n-1)
```
ivokub: There is actually not-advertised interface:
```
// bitsComparatorConstant allows to use the built-in comparison against a
// constant bound. We use the direct implementation due to the added efficiency
// of directly creating constraints instead of using API.
type bitsComparatorConstant interface {
	MustBeLessOrEqCst(aBits []frontend.Variable, bound *big.Int, aForDebug frontend.Variable)
}
```
which all the frontends we have satisfy (see std/math/bits/intf_constcheck.go). We can use this for comparing selBits against `n-1`. Then can avoid importing `cmp` package.

But if you use it then you should implement a fallback in case it is not implemented (or just panic explicitly, this is also fine imo).

std/selector/mux.go
ivokub: Hmm - I think this method would requires a test on edge cases. Looking at it now I'm not 100% sure it works.

ivokub: But better imo not to add and just use `BinaryMux` with padding inputs by const `0`. (I haven't reviewed whole PR yet, perhaps later on there is point for it)

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -270,6 +270,20 @@ func (g2 G2) neg(p *G2Affine) *G2Affine {
 	}
 }
 
+// muxE2Y8Signed selects from 8 E2 Y-values using selector (0-7) and conditionally
+// negates based on signBit. This optimizes the common GLV pattern where Y[i] =
+// -Y[15-i], reducing a 16-to-1 Mux to an 8-to-1 Mux plus conditional negation.
+func (g2 *G2) muxE2Y8Signed(signBit frontend.Variable, selector frontend.Variable, yA0, yA1 [8]*emulated.Element[BaseField]) *fields_bls12381.E2 {
+	baseA0 := g2.fp.Mux(selector, yA0[0], yA0[1], yA0[2], yA0[3], yA0[4], yA0[5], yA0[6], yA0[7])
```
ivokub: `baseA0 := g2.fp.Mux(selector, yA0[:]...)` better

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -692,18 +706,12 @@ func (g2 *G2) scalarMulGLV(Q *G2Affine, s *Scalar, opts ...algopts.AlgebraOption
 	// T = Q + [3]Φ(Q)
 	// P = B1 and P' = B4
 	T4 := g2.add(tableQ[1], tablePhiQ[2])
```
ivokub: I would rename such that we have table entries only `TN` and rest as `tN` (which we use to define table entries). Imo then it is more clear which go into table and which not.

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -1360,8 +1352,6 @@ func (c *Curve[B, S]) scalarMulFakeGLV(Q *AffinePoint[B], s *emulated.Element[S]
 	// P = B2 and P' = B3
 	T7 := c.Neg(T4)
 	// T = -[3]Q - R
```
ivokub, line 1354: Similarly also in this file perhaps lets distinguish between `TN` and `tN` to indicate which values will be muxed and which not (temporary values for computing rest). Also in other methods (`BN` and `bN`)

std/math/emulated/field_ops.go
yelhousni: Replaced `cmp.NewBoundedComparator` with a direct `MustBeLessOrEqCst` call via interface type assertion on `api.Compiler()`.  But here was a latent bug in `MustBeLessOrEqCst`: when padding short bit arrays to field bit length, the raw `int(0)` padding was not compatible with newR1C/direct type assertions. Changed to use proper builder-typed zero values (`builder.toVariable(0)` for R1CS, `expr.Term[E]{}` for SCS).

yelhousni: actually SCS plonk panics: `expr.Term[E]{}` is treated as constant zero by `builder.Sub`, which returns `*big.Int`, and then the `.(expr.Term[E])` assertion at line 281 fails. I will revert the padding to raw `int(0)` and skip constraints in the `bound.Bit(i) == 0` branch when `aBits[i]` is constant zero.

ivokub (approved):
Thanks for the fixes! Looks good now!

Issue #1724, opened 2026-03-02
perf: GKR - Section 3.2 of Gru24


Tabaie:
Section 3.2 of https://eprint.iacr.org/2024/108.pdf describes a clever optimization on Zero-Check reducing the degree of round polynomials and thus saving an entire gate evaluation pass and a multiplication pass at every round. This is not available when claims from multiple evaluation points are folded, in which case factoring out parts of the eq term is not possible.

peter941221:
I took a narrow v1 pass on this and opened a PR:

- `https://github.com/Consensys/gnark/pull/1744`

Scope-wise, I kept it intentionally strict:

- only enable the Section 3.2 optimization when the whole `GkrSumcheckLevel` collapses to one distinct `GkrClaimSource`
- keep folded multi-source levels on the legacy path

Implementation-wise, this is not just a degree constant tweak:

- add a schedule-side single-source eligibility boundary
- switch eligible levels to the weighted Section 3.2 verifier recurrence
- update both verifier surfaces together:
  - native verifier
  - in-circuit verifier
- use a reduced eq-table prover path that strips one current-variable eq factor per round

I also added a couple of reviewer-confidence checks around the trickiest algebraic edge:

- direct `coeff == 1` weighted-reconstruction tests for the native verifier path
- a matching frontend / circuit-side verifier test

On the evidence side, the regenerated vectors shrink exactly on the eligible levels and stay unchanged on legacy-only shapes.

If you'd prefer, I can also split the write-up further or add a short code comment around the optimized final-check path.

PR #1744, opened 2026-04-01
perf(gkr): apply Gru24 Section 3.2 to single-source sumcheck levels

peter941221:
## Summary

This implements a narrow v1 of the optimization discussed in `#1724`.

For `GkrSumcheckLevel`s whose claims all refer to the same distinct `GkrClaimSource`, we switch the zero-check round relation from the legacy unweighted form to the weighted Section 3.2 recurrence from Gru24. This lets the prover send a degree-`d` round polynomial instead of degree-`d+1` for those eligible levels.

Folded multi-source levels keep the legacy path unchanged.

## What changes

- add a schedule-side eligibility helper:
  - `constraint.GkrSumcheckLevel.SingleClaimSource()`
- reduce `ZeroCheckDegree(...)` only for eligible levels
- update both verifier paths:
  - native verifier
  - in-circuit verifier
- update the prover to use a reduced eq-table path for eligible levels
  - the current-variable eq factor is stripped after each fold
- refresh generated GKR backends and affected test vectors

## Why the scope is narrow

Section 3.2 is not just a local prover micro-optimization. It changes the round relation seen by the verifier.

In gnark's current GKR schedule, some levels fold claims coming from multiple distinct evaluation points. I did not try to force the new recurrence onto those shapes in this PR. Instead, the optimization is only enabled when the whole level collapses to one distinct source.

That keeps the patch aligned with the paper identity without widening into a larger schedule/protocol refactor.

## Evidence

The regenerated test vectors show the expected proof-size reduction only on eligible levels. For example:

- `single_mul_gate_two_instances.json`: total partial-sum coefficient count `3 -> 2`
- `single_mimc_gate_two_instances.json`: `8 -> 7`
- `single_mimc_gate_four_instances.json`: `16 -> 14`

Non-eligible / legacy-only vector shapes stayed unchanged.

## Validation

- `go test ./internal/gkr/... -count=1`
- `go run .` from `internal/gkr/test_vectors`

I also added a `TestZeroCheckDegreeDispatch` assertion to the generated curve-test packages to check that:

- eligible single-source levels use the reduced degree
- folded multi-source levels stay on the legacy degree

And I added direct `coeff == 1` reconstruction tests for both verifier surfaces:

- generated native sumcheck verifier tests
- frontend / in-circuit sumcheck verifier test

## Notes

- I intentionally did not try to optimize folded multi-source levels in this PR.
- I also did not include benchmark numbers here because the existing `BenchmarkGkrMimc17` benchmark currently panics on my Windows setup on both clean `HEAD` and this branch, so it did not provide a trustworthy regression signal for this change.

---

> [!NOTE]
> **High Risk**
> Changes GKR sumcheck verification and proof shape (native + in-circuit); incorrect round reconstruction or degree dispatch would break soundness or reject valid proofs.
> 
> **Overview**
> This PR applies **Gru24 Section 3.2** to GKR zero-check levels where every claim shares one **`GkrClaimSource`**, shrinking sumcheck round polynomials by one coefficient per variable while leaving **multi-source folded** levels on the legacy path.
> 
> **Schedule & degree:** Eligibility is exposed via **`SingleClaimSource()`**; **`ZeroCheckDegree`** drops by one only for those levels. **`TestZeroCheckDegreeDispatch`** (and schedule tests) assert reduced vs legacy partial-sum lengths on a Poseidon2 proof.
> 
> **Verifiers:** Native and in-circuit **`sumcheckVerify`** gain **`roundCombinationCoeff`** so rounds can reconstruct **`gⱼ(0)`** when the combination coefficient is **1** (optimized path) or use the old **`gⱼ(0) + gⱼ(1) = gⱼ₋₁(r)`** rule otherwise. Frontend **`zeroCheckLazyClaims`** skips **`eq`** weighting in final eval when the level is single-source optimized.
> 
> **Tests & codegen:** Direct **`coeff == 1`** reconstruction tests and regenerated GKR backends / vectors document smaller proofs on eligible circuits only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 954c057a55cf3ae07606680d03af1a9edc135d36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Tabaie:
Thank you @peter941221 for the submission! This duplicates https://github.com/Consensys/gnark/pull/1742 as far as I can see.
Also curious about the benchmark, will it also fail with a smaller size constant?

peter941221:
Thanks for the feedback!

I agree there is significant overlap with #1742. My intent here was to validate the narrower single-source-only path and keep the protocol / schedule surface area small. If you'd prefer to keep the main optimization consolidated in #1742, I'm happy to rework this into a supporting PR that contributes any useful pieces (for example tests, eligibility plumbing, or narrower verifier/prover scaffolding) rather than competing with it.

On the benchmark side, the panic reproduces for me on clean HEAD as well, so I didn't want to treat it as a branch-specific signal. I haven't yet tried reducing the size constant, but I'm happy to do that and report back.

Tabaie:
Hi @peter941221! https://github.com/Consensys/gnark/pull/1742 is now merged. Please let me know if you'd like to keep working on this.

peter941221:
for sure, I'd like to contribute. Let me know which hardcore issue we need to conquer.

peter941221:
Closing this one.

I rebased the branch state locally and rechecked the current head against master. The old bugbot finding was addressed on the branch, but the PR now overlaps directly with #1742 at the GKR protocol / schedule layer and no longer looks like the right vehicle for a follow-up merge.

If there is still interest in the dynamic single-source scheduling direction, I can spin that into a fresh, smaller PR on top of current master.

Issue #831, opened 2023-09-13
feat: implement lookup-table base comparison and less-than assertion


ivokub:
**Is your feature request related to a problem? Please describe.**
The current approach for comparison and less-than assertion is costly, requiring several constraints per bit of the values being compared. It is slightly more efficient when using `std/math/cmp` gadget.

**Describe the solution you'd like**
Define comparison and less-than assertion using lookup tables. There are possibly several approaches - one approach is to have a lookup table consisting of values `x | y | x<y` and then we can use the inequality. The benefit of this approach is that we can partition the values into groups of 8 bits, but on the other hand requires big initial overhead of constructing the lookup table.

@ThomasPiellard also described that the current approach is Cmp could be extended to larger values than bits, but before efficient range checks using lookup tables wasn't worth it.

**Describe alternatives you've considered**
None

**Additional context**
None

rico201:
gh pr checkout 835

PR #1746, opened 2026-04-03
feat: map-to-curve relations

yelhousni:
# Description

Add increment-and-check map-to-curve gadgets for short Weierstrass curves, implementing the constructions from https://eprint.iacr.org/2026/590.pdf.

Two methods are provided:
- **X-increment**: encodes X = M·256 + K, verifies Y² = X³ + aX + b, and checks a 2^S-th root witness for inverse-exclusion. Only practical for low 2-adicity fields (S ≤ 4).
- **Y-increment**: encodes Y = M·256 + K, verifies Y² = X³ + aX + b. Simpler (no inverse-exclusion witness), works for any 2-adicity, recommended for j=0 curves to avoid the algebraic attack from the paper.

### `std/algebra/emulated/sw_emulated/maptocurve_increment/`
Generic emulated map-to-curve for any supported curve;
- **BN254** (y² = x³ + 3): x-increment and y-increment
- **secp256k1** (y² = x³ + 7): x-increment and y-increment
- **P-256/secp256r1** (y² = x³ − 3x + b): x-increment and y-increment (y-increment uses Cardano cubic solver from https://github.com/Consensys/gnark-crypto/pull/831)

### `std/algebra/native/sw_grumpkin/maptocurve_increment/`
Native y-increment for Grumpkin (y² = x³ − 17), compiled over BN254.

### `std/algebra/native/sw_bls12377/maptocurve_increment/`
Native y-increment for BLS12-377 (y² = x³ + 1), compiled over BW6-761.

## Constraint counts

| Curve | Method | R1CS | SCS (PLONK) |
|-------|--------|------|-------------|
| **Emulated** | | | |
| BN254 | x-increment | 821 | 2,984 |
| BN254 | y-increment | 626 | 2,397 |
| secp256k1 | x-increment | 821 | 2,963 |
| secp256k1 | y-increment | 626 | 2,397 |
| P-256 | x-increment | 825 | 2,976 |
| P-256 | y-increment | 630 | 2,410 |
| **Native** | | | |
| Grumpkin | y-increment | 15 | 37 |
| BLS12-377 | y-increment | 15 | 37 |

## Type of change

# How has this been tested?

All tests verify circuit satisfiability via `test.CheckCircuit` (both Groth16 and PLONK backends):

- [x] `TestXIncrementEmulatedBN254` — x-increment on BN254
- [x] `TestXIncrementEmulatedSecp256k1` — x-increment on secp256k1
- [x] `TestXIncrementEmulatedP256` — x-increment on P-256
- [x] `TestYIncrementEmulatedBN254` — y-increment on BN254
- [x] `TestYIncrementEmulatedSecp256k1` — y-increment on secp256k1
- [x] `TestYIncrementEmulatedP256` — y-increment on P-256 (Cardano solver)
- [x] `TestYIncrement` (maptocurve_grumpkin) — native y-increment on Grumpkin
- [x] `TestYIncrement` (maptocurve_bls12377) — native y-increment on BLS12-377

```bash
go test ./std/algebra/emulated/sw_emulated/maptocurve_increment/...
go test ./std/algebra/native/sw_grumpkin/maptocurve_increment/...
go test ./std/algebra/native/sw_bls12377/maptocurve_increment/...
```

- [x] Constraint count benchmarks for all curves and methods (R1CS + SCS), on Macbook Pro M5

```bash
go test ./internal/stats/...
# regenerate the full reference CSV (run from the generate dir; do NOT pass
# -run, as the filter would truncate latest_stats.csv to only matching rows):
cd internal/stats/generate && go run . -s
```

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> New protocol-facing curve-mapping gadgets affect soundness if misused (documented non-uniqueness and missing cofactor clear); scope is additive with broad circuit tests but depends on a pre-release gnark-crypto bump.
> 
> **Overview**
> Adds **increment-and-check map-to-curve** relations (eprint 2026/590): encode a message as `M·256 + K`, prove an on-curve point via hints plus in-circuit checks. **Emulated** `maptocurve_increment` exposes `Mapper` with **X-increment** (curve equation + `Z^{2^s}=Y` witness) and **Y-increment** (reconstruct `Y`, on-curve only) for BN254, secp256k1, and P-256, with `Increment` picking the variant by curve shape. **Native** y-increment packages cover **Grumpkin** (BN254) and **BLS12-377** (BW6-761).
> 
> Hints are registered in `std/hints.go`; **internal/stats** snippets and **latest_stats.csv** add Groth16/PLONK constraint rows for each gadget. **gnark-crypto** is bumped (pre-release) so y-increment can use **`Cbrt`** on `tinyfield` and P-256 y-increment can use **Cardano** roots; `tinyfield` also gains local **`Cube`/`Cbrt`**. `tinyfield` vector parallel work now uses **`gnark-crypto/parallel`** instead of a copied helper.
> 
> Package docs stress this is **not** hash-to-curve: mapping is **non-unique**, **no cofactor clearing**, and **`msg < q/256`** is caller-enforced.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a93b9ba134685c40b129c9cd414d8633d1e63bd9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gbotrel:
Review pass from 2026-05-07.

I did not find blocking issues in this pass. The gnark-crypto dependency PR #831 is already merged, and the map-to-curve gadgets compile and solve in the targeted native/emulated packages I checked.

Verification run locally:
- `go test ./std/algebra/emulated/maptocurve/... ./std/algebra/native/maptocurve_grumpkin/... ./std/algebra/native/maptocurve_bls12377/...`
- `go test ./internal/smallfields/tinyfield`

internal/smallfields/tinyfield/element.go
```diff
@@ -907,6 +907,31 @@ func (z *Element) Sqrt(x *Element) *Element {
...
+	return z
+}
+
+// Cube sets z to x^3 and returns z
+func (z *Element) Cube(x *Element) *Element {
+	var t Element
+	t.Square(x).Mul(&t, x)
+	z.Set(&t)
+	return z
+}
```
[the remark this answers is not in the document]
yelhousni: Not worth a template conditional emit for tinyfield

std/algebra/emulated/maptocurve/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+	assert.CheckCircuit(&yIncrementCircuit[F]{}, opts...)
+}
+
+func TestYIncrementEmulatedBN254(t *testing.T)     { testYIncrement[emulated.BN254Fp](t) }
+func TestYIncrementEmulatedSecp256k1(t *testing.T) { testYIncrement[emulated.Secp256k1Fp](t) }
+func TestYIncrementEmulatedP256(t *testing.T)      { testYIncrement[emulated.P256Fp](t) }
+
+// --- Benchmarks ---
+
+func BenchmarkXIncrementEmulated(b *testing.B) {
```
ivokub: The benchmarks I think are not effective as benchmarks. I think it would be better to remove them from here and instead add code snippets in the stats test.

std/algebra/emulated/maptocurve/hints.go
```diff
@@ -0,0 +1,365 @@
...
+	bn254fp "github.com/consensys/gnark-crypto/ecc/bn254/fp"
+	secp256k1fp "github.com/consensys/gnark-crypto/ecc/secp256k1/fp"
+	"github.com/consensys/gnark-crypto/ecc/secp256r1"
+	secp256r1fp "github.com/consensys/gnark-crypto/ecc/secp256r1/fp"
+	"github.com/consensys/gnark/constraint/solver"
+	"github.com/consensys/gnark/std/algebra/emulated/sw_emulated"
+)
+
+func init() {
+	solver.RegisterHint(GetHints()...)
```
ivokub: We should also add hint registration to std/hints.go

std/algebra/emulated/maptocurve/hints.go
```diff
@@ -0,0 +1,365 @@
...
+		decompose(roots[0].BigInt(&xBig), nbLimbs, nbBits, outputs[1:1+nbLimbs])
+		return nil
+	}
+	return fmt.Errorf("yIncrementHint: no valid k found for secp256r1")
+}
+
+// --- limb helpers ---
+
+// recompose reconstructs a big.Int from its limbs (little-endian, nbBits per limb).
+func recompose(limbs []*big.Int, nbLimbs int, nbBits uint) *big.Int {
```
ivokub: We could use the methods defined in std/internal/limbcomposition/composition.go instead? I.e. instead of recompose we can use limbcomposition.Recompose and instead of decompose we could use limbcomposition.Decompose.

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+func (m *Mapper[F]) XIncrement(msg *emulated.Element[F]) (x, y *emulated.Element[F], err error) {
+	fp := m.field
+	nbLimbs := int(m.nbLimbs)
+
+	// hint inputs: [nbLimbs, nbBits, q_limbs..., msg_limbs...]
+	// hint outputs: [k, x_limbs..., y_limbs..., z_limbs...]
+	hintOutputs := 1 + 3*nbLimbs // k + x + y + z
+	hintInputs := m.buildHintInputs(msg)
+
+	res, err := m.api.Compiler().NewHint(xIncrementHint, hintOutputs, hintInputs...)
```
ivokub: We should use field.NewHint? And then inside hint Unwraphint as we do elsewhere where do call non-native hints. This already ensures that everything is properly range checked etc.

std/algebra/emulated/maptocurve/hints.go
```diff
@@ -0,0 +1,365 @@
...
+func GetHints() []solver.Hint {
+	return []solver.Hint{
+		xIncrementHint,
+		yIncrementHint,
+	}
+}
+
+// parseHintInputs extracts the field modulus and message from the hint inputs.
+// Format: [nbLimbs, nbBits, q_limbs..., msg_limbs...]
+func parseHintInputs(inputs []*big.Int) (q *big.Int, nbLimbs int, nbBits uint, msg *big.Int, err error) {
```
ivokub: We don't need to do the hint parsing here I think -- if we would use non-native Hint calling mechanism and inside hint Unwrap then it already automatically does the modulus and input packing.

We also don't need to pass nbBits and nbLimbs -- because the x/y increment hints are specialized for curves then we can call emulated.GetEffectiveParams[emparams.XXX](field) to get nbBits and nbLimbs already.

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	bVal := fp.NewElement(m.b)
+	rhs = fp.Add(rhs, bVal)
+	fp.AssertIsEqual(lhs, rhs)
+}
+
+// nativeToEmulated converts a native variable (fitting in one limb) to an
+// emulated element.
+func (m *Mapper[F]) nativeToEmulated(v frontend.Variable) *emulated.Element[F] {
+	nbLimbs := int(m.nbLimbs)
```
ivokub: We don't need to allocate whole limb space -- I think v is small enough always that we can pack everything into a single limb.

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	result := make([]*big.Int, nbLimbs)
+	for i := 0; i < nbLimbs; i++ {
+		result[i] = new(big.Int).And(tmp, mask)
+		tmp.Rsh(tmp, nbBits)
+	}
+	return result
+}
+
+// assertOnCurve checks Y² = X³ + a·X + b.
+func (m *Mapper[F]) assertOnCurve(x, y *emulated.Element[F]) {
```
ivokub: Perhaps instead of defining assertOnCurve here we should just instantiate the sw_emulated.Curve as a field in Mapper?

ivokub: And then we can call sw_emulated.Curve.AssertOnCurve method directly

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+
+	msgLimbs := fp.Reduce(msg).Limbs
+	for i := 0; i < nbLimbs; i++ {
+		inputs = append(inputs, msgLimbs[i])
+	}
+	return inputs
+}
+
+// decomposeBigInt splits v into nbLimbs limbs of nbBits each (little-endian).
+func decomposeBigInt(v *big.Int, nbLimbs int, nbBits uint) []*big.Int {
```
ivokub: It is duplicate with the method in hints.go? And I think we can also use limbdecomposition.Decompose as well.

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// (2) Range: 0 ≤ K < 256
+	rangecheck.New(m.api).Check(k, 8)
+
+	return xEl, yEl, nil
+}
+
+// buildHintInputs constructs hint inputs: [nbLimbs, nbBits, q_limbs..., msg_limbs...]
+// Curve coefficients are not passed; the hint dispatches on q to look them up.
+func (m *Mapper[F]) buildHintInputs(msg *emulated.Element[F]) []frontend.Variable {
```
ivokub: I think we don't need the method when we use the Emulated.NewHint method.

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	return &Mapper[F]{api: api, field: field, a: a, b: b, s: s, nbLimbs: nbLimbs, nbBits: nbBits}, nil
+}
+
+// XIncrement maps msg to a curve point (x, y) using the x-increment method:
+//
+//	X = msg·256 + k, Y² = X³ + aX + b, Z^{2^s} = Y
+//
+// where k ∈ [0, 256) is found by the hint, and the Z witness chain ensures Y
+// is not the inverse of a valid y-coordinate (needed for j=0 curves).
+func (m *Mapper[F]) XIncrement(msg *emulated.Element[F]) (x, y *emulated.Element[F], err error) {
```
ivokub: Don't we have to check that msg is small? I.e. that there is eight bits of space to add k? Otherwise I feel there may be some soundness issues due to overflows. But I haven't thought much about it.

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	fp := m.field
```
ivokub: I think the method should error if called for curves which don't support it (as described in documentation)

ivokub: Same for yincrement

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	yEl := fp.Add(fp.Mul(msg, Tconst), kEl)
+
+	// (1) Curve equation: Y² = X³ + a·X + b
+	m.assertOnCurve(xEl, yEl)
+
+	// (2) Range: 0 ≤ K < 256
+	rangecheck.New(m.api).Check(k, 8)
+
+	return xEl, yEl, nil
+}
```
ivokub: Perhaps we should have canonical `Increment` method which chooses XIncrement or YIncrement based on curve?

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return nil, err
+	}
+	a, b := curveCoefficients[F]()
+	s := twoAdicity[F]()
+	nbLimbs, nbBits := emulated.GetEffectiveFieldParams[F](api.Compiler().Field())
+	return &Mapper[F]{api: api, field: field, a: a, b: b, s: s, nbLimbs: nbLimbs, nbBits: nbBits}, nil
+}
+
+// XIncrement maps msg to a curve point (x, y) using the x-increment method:
```
ivokub: Why instead of returning x,y we don't return a point?

std/algebra/emulated/maptocurve/maptocurve.go
```diff
@@ -0,0 +1,207 @@
...
+	a, b    *big.Int
+	s       int  // 2-adicity v₂(q-1) for x-increment inverse-exclusion
+	nbLimbs uint // effective number of limbs for the emulated field
+	nbBits  uint // effective bits per limb
+}
+
+// NewMapper creates a new Mapper for the curve defined by field type F.
+// Curve coefficients are read from [sw_emulated.GetCurveParams] when available.
+// The 2-adicity s (for x-increment) is auto-detected from the field modulus.
+func NewMapper[F emulated.FieldParams](api frontend.API) (*Mapper[F], error) {
```
ivokub: Imo we should take as input the concrete curve as well.

std/algebra/emulated/maptocurve/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+type xIncrementCircuit[F emulated.FieldParams] struct {
+	M emulated.Element[F]
+}
+
+func (c *xIncrementCircuit[F]) Define(api frontend.API) error {
+	m, err := NewMapper[F](api)
+	if err != nil {
+		return err
+	}
+	_, _, err = m.XIncrement(&c.M)
```
ivokub: We don't check the return value?

std/algebra/emulated/maptocurve/hints.go
```diff
@@ -0,0 +1,365 @@
...
+		decompose(z.BigInt(&zBig), nbLimbs, nbBits, outputs[1+2*nbLimbs:1+3*nbLimbs])
+		return nil
+	}
+	return fmt.Errorf("xIncrementHint: no valid k found for secp256k1 (s=%d)", s)
+}
+
+func nthRoot2SSecp256k1(a *secp256k1fp.Element, s int) *secp256k1fp.Element {
+	z := new(secp256k1fp.Element).Set(a)
+	for i := 0; i < s; i++ {
+		if z.Sqrt(z) == nil {
```
ivokub: Imo it would be a bit cleaner if we would have `Sqrt2th` etc method in gnark-crypto for computing `s^{1/2^s}`. But keep the current helper. But in that case I would inline the function into corresponding xIncrementSecp256k1 function for clarity.

std/algebra/emulated/maptocurve/doc.go
```diff
@@ -0,0 +1,13 @@
...
+//
+// Two methods are provided:
+//   - [Mapper.XIncrement]: encodes X = M·256 + K, verifies the curve equation, and
+//     ensures Y has a 2^S-th root (inverse-exclusion witness) for j=0 curves.
+//     Only practical for low 2-adicity fields (S ≤ 4). For high 2-adicity
+//     fields (e.g. BLS12-377 with S=46, Grumpkin with S=28) the witness search
+//     becomes infeasible — use [Mapper.YIncrement] instead.
+//   - [Mapper.YIncrement]: encodes Y = M·256 + K, verifies the curve equation.
+//     Simpler (no inverse-exclusion witness), works for any 2-adicity, and is
+//     the recommended method for j=0 curves.
```
ivokub: I would add that the mapping is not unique! Compared to SSWU mapping we already have for BLS12-381 then this is a big distinction when used in the context where uniqueness is assumed.

One option to solve it would be deterministically check in the circuit that returned K is indeed smallest unique (i.e. imo the folklore approach is to check that all other k'<k lead to point not on curve.)

And am I right that it is only map to curve, but not to subgroup? What are practical use cases? When the increment approach is not unique then imo it is unsafe to use hashing context and if we cannot ensure it is not in subgroup then the curve arithmetic can break down.

std/algebra/emulated/maptocurve/doc.go
```diff
@@ -0,0 +1,13 @@
+// Package maptocurve implements increment-and-check map-to-curve gadgets for
+// short Weierstrass curves y² = x³ + ax + b over emulated fields.
+//
+// Two methods are provided:
```
ivokub: Imo we should also mention that the method is not guaranteed to converge. Imo 8-bit search space is sufficient --> we have 256 candidates and if probability that every candidate leads to quadratic non-residue is 1/2 then the failure rate is 2^{-256}. It is small (but holds only for XIncrement method, dunno about YIncrement).

Though I'm afraid that for some very specially crafted inputs the failure probability could be smaller. It is probably fine, but good to document.

ivokub (changes requested):
Please see comments. I currently reviewed only emulated/maptocurve, not bls12377/grumpkin.

Imo we should also reconsider the package structure - for BLS12-381 curve we already have the MapToCurve (and also MapToG1/MapToG2) implemented. Imo having a separate package maptocurve is confusing.

I'd recommend instead:
* std/algebra/emulated/sw_emulated/maptocurve_increment
* std/algebra/native/sw_bls12377/maptocurve_increment
* std/algebra/native/sw_grumpkin/maptocurve_increment

And we should be clear in the documentation that these are very fast non-unique algorithm.

std/algebra/emulated/maptocurve/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/consensys/gnark-crypto/ecc"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/frontend/cs/r1cs"
+	"github.com/consensys/gnark/frontend/cs/scs"
+	"github.com/consensys/gnark/std/math/emulated"
+	"github.com/consensys/gnark/test"
+)
+
+// test message values
+var testMessages = []*big.Int{
```
ivokub: I'd add more test cases -- large messages which don't have 8 bits of space and message which has exactly 8 bits of space.

std/algebra/native/sw_grumpkin/maptocurve_increment/maptocurve.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// YIncrement maps msg to a point (x, y) on the Grumpkin curve y² = x³ − 17
+// using the y-increment method:
+//
+//	Y = msg·256 + K, Y² = X³ − 17
+//
+// Caller-side precondition: msg < q/256. The precondition is NOT enforced
+// in-circuit — see the package doc.
+func YIncrement(api frontend.API, msg frontend.Variable) (x, y frontend.Variable, err error) {
```
ivokub, line 22: I guess we return x/y instead of a point because we don't want to indicate its in subgroup (`G1Affine`)?

std/algebra/native/sw_grumpkin/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/test"
+)
+
+type yIncrementCircuit struct {
+	M frontend.Variable
+}
+
+func (c *yIncrementCircuit) Define(api frontend.API) error {
+	_, _, err := YIncrement(api, c.M)
```
ivokub: Doesn't test that it is expected result.

std/algebra/native/sw_bls12377/maptocurve_increment/maptocurve.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// YIncrement maps msg to a point (x, y) on the BLS12-377 curve y² = x³ + 1
+// using the y-increment method:
+//
+//	Y = msg·256 + K, Y² = X³ + 1
+//
+// Caller-side precondition: msg < q/256. The precondition is NOT enforced
+// in-circuit — see the package doc.
+func YIncrement(api frontend.API, msg frontend.Variable) (x, y frontend.Variable, err error) {
```
ivokub, line 22: Returning x/y vs G1Affine as well.

std/algebra/native/sw_bls12377/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/test"
+)
+
+type yIncrementCircuit struct {
+	M frontend.Variable
+}
+
+func (c *yIncrementCircuit) Define(api frontend.API) error {
+	_, _, err := YIncrement(api, c.M)
```
ivokub: No correctness test.

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+type xIncrementCircuit[B, S emulated.FieldParams] struct {
+	M emulated.Element[B]
+}
+
+func (c *xIncrementCircuit[B, S]) Define(api frontend.API) error {
+	crv, err := sw_emulated.New[B, S](api, sw_emulated.GetCurveParams[B]())
+	if err != nil {
+		return err
+	}
+	m, err := NewMapper[B, S](api, crv)
```
ivokub: The generic parameter is inferred, no need to define `[B, S]`

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+	p, err := m.XIncrement(&c.M)
+	if err != nil {
+		return err
+	}
+	// touch the returned point so the compiler keeps it constrained.
+	crv.AssertIsOnCurve(p)
```
ivokub, line 131: We only perform AssertIsOnCurve, but not a correctness test (i.e. that with a honest hint our computed point corresponds to expected result)

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+type yIncrementCircuit[B, S emulated.FieldParams] struct {
+	M emulated.Element[B]
+}
+
+func (c *yIncrementCircuit[B, S]) Define(api frontend.API) error {
+	crv, err := sw_emulated.New[B, S](api, sw_emulated.GetCurveParams[B]())
+	if err != nil {
+		return err
+	}
+	m, err := NewMapper[B, S](api, crv)
```
ivokub: Generic parameters are inferred

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+	p, err := m.YIncrement(&c.M)
+	if err != nil {
+		return err
+	}
+	crv.AssertIsOnCurve(p)
```
ivokub, line 184: No correctness test

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve_test.go
```diff
@@ -0,0 +1,175 @@
...
+	testYIncrement[emulated.P256Fp, emulated.P256Fr](t)
+}
+
+// --- Increment dispatcher tests ---
+
+type incrementCircuit[B, S emulated.FieldParams] struct {
+	M emulated.Element[B]
+}
+
+func (c *incrementCircuit[B, S]) Define(api frontend.API) error {
```
ivokub, line 223: Same comments as for X and Y

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve.go
```diff
@@ -0,0 +1,172 @@
...
+	api   frontend.API
+	curve *sw_emulated.Curve[B, S]
+	field *emulated.Field[B]
+	a, b  *big.Int
+	s     int // 2-adicity v₂(q-1); used by XIncrement only.
+}
+
+// NewMapper returns a [Mapper] bound to the given emulated curve. The curve
+// must use a supported base field — currently BN254, secp256k1 or P-256.
+// Other moduli will surface as an error at hint-execution time.
```
ivokub: I think this is not sufficient. It would be better to already fail here. Otherwise the user is able to compile a circuit which is unsolvable and we only encounter it when trying to prove. But this is dangerous in the sense that the circuit may already be deployed etc.

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return p, nil
+}
+
+// nativeAsEmulated lifts a small native variable (must fit in one limb; the
+// caller enforces the range) into an emulated element of width nbLimbs by
+// padding the higher limbs with zero. We can't use [emulated.Field.NewElement]
+// directly here because it expects exactly nbLimbs limbs.
+func (m *Mapper[B, S]) nativeAsEmulated(v frontend.Variable) *emulated.Element[B] {
+	nbLimbs, _ := emulated.GetEffectiveFieldParams[B](m.api.Compiler().Field())
```
ivokub: I don't think we need to set all limbs. We should be able to do:
```go
m.field.NewElement([]frontend.Variable{v})
```
and in this case I think we should just inline it at caller side and remove this function.

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, fmt.Errorf("new base field: %w", err)
+	}
+	params := sw_emulated.GetCurveParams[B]()
+	return &Mapper[B, S]{
+		api:   api,
+		curve: curve,
+		field: field,
+		a:     params.A,
+		b:     params.B,
+		s:     twoAdicity[B](),
```
ivokub: And if we check that the curve is one of the supported ones, then we could just already hardcode the 2-adicity as well instead of having a helper.

ivokub (changes requested):

std/algebra/native/sw_grumpkin/maptocurve_increment/maptocurve.go
yelhousni: yes, bare x/y is intentional to avoid implying G1Affine subgroup membership

std/algebra/emulated/sw_emulated/maptocurve_increment/maptocurve.go
yelhousni: Good call on removing the helper — done. One wrinkle worth flagging: `m.field.NewElement([]frontend.Variable{v})` actually panics here. The strict path runs `enforceWidth` with `nbBits = Modulus().BitLen()`, which requires *exactly* `ceil(254/64) = 4` limbs, so a single-limb element is rejected (`enforcing width element with inexact number of limbs`). That's precisely why the original helper zero-padded to `nbLimbs`.

So I inlined it as a single-limb element built from K's 8-bit decomposition instead:

```go
kEl := m.field.FromBits(bits.ToBinary(m.api, k, bits.WithNbDigits(8))...)
```

This also lets me drop the separate `rangecheck.New(m.api).Check(k, 8)` — the `ToBinary` already enforces `0 ≤ K < T` — and removes the redundant 64-bit limb range-check the strict `NewElement` path was emitting on limb 0.

Net: helper gone, single limb, inlined, and constraints down across all emulated variants (e.g. `y_bn254` 673→626 R1CS, `x_bn254` 3135→2984 SCS). `internal/stats/latest_stats.csv` and the PR description table are updated to match.

ivokub (approved):
Thanks for the fixes! I updated a bit more to use the new feature we added for the subscalar-GLV fix (custom range checking the emulated hint outputs). It gives a bit more reduction in the constraints and imo is cleaner.

PR #1749, opened 2026-04-12
feat(emulated): checkZero mulcheck in hint

### LINKED ISSUE #1748: fix(emulated): checkZero mulcheck should fail in hint
**Is your feature request related to a problem? Please describe.**
When an emulated element assert fails, it's hard to track the point of failure.

This is because when mulcheck is returned by the hint, the remainder is swapped out with zero, which is very constraint efficient.
However, a failing assert only triggers much later in deferred check call, making it hard to track the bug.

**Describe the solution you'd like**

Instead of adding a failing mulcheck in assert, assert call can tell the hint if a particular mulcheck should have zero remainder (forward existing `isMulMod` var) to the hint, and then hint can fail on computation itself allowing for easy error tracking.

This will make sure only valid `mulcheck`s are added to deferred checks. And have better devX for writing emulated circuits.

**Describe alternatives you've considered**
Tried script for debugging (thanks @ivokub). Which helped with the debugging :)
But I believe this should be cleaner.

**Additional context**
Almost ended up thinking `mulcheck`s weren't working properly for my large unruly overflowing emulated elements. Validated the added mulcheck in some python playground. It was failing. Found out about assert calls setting mulcheck rem to `0`.

With this, we'll have no extra constraints, and error call stacks will provide easier path to follow for finding issues.

shramee:
closes consensys/gnark#1748

# Description

**Is your feature request related to a problem? Please describe.**
When an emulated element assert fails, it's hard to track the point of failure.

This is because when mulcheck is returned by the hint, the remainder is swapped out with zero, which is very constraint efficient.
However, a failing assert only triggers much later in deferred check call, making it hard to track the bug.

**Describe the solution you'd like**

Instead of adding a failing mulcheck in assert, assert call can tell the hint if a particular mulcheck should have zero remainder (forward existing `isMulMod` var) to the hint, and then hint can fail on computation itself allowing for easy error tracking.

This will make sure only valid `mulcheck`s are added to deferred checks. And have better devX for writing emulated circuits.

**Describe alternatives you've considered**
Tried script for debugging (thanks @ivokub). Which helped with the debugging :)
But I believe this should be cleaner.

**Additional context**
Almost ended up thinking `mulcheck`s weren't working properly for my large unruly overflowing emulated elements. Validated the added mulcheck in some python playground. It was failing. Found out about assert calls setting mulcheck rem to `0`.

With this, we'll have no extra constraints, and error call stacks will provide easier path to follow for finding issues.

## Type of change

# How has this been tested?

- [x] All emulated tests for BN254 fields

- na -

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Modifies the `mulHint` input contract and adds an early error path in a core emulated-field hint; could change when/where failing circuits error and may break any callers that assume the old hint input layout.
> 
> **Overview**
> Propagates a `checkZero` flag into the multiplication hint inputs (via `callMulHint`), extending the hint’s input header by one value.
> 
> Updates `mulHint` to parse the new flag and, when invoked for `checkZero`/mulcheck cases, **fail immediately** if `a*b mod p` produces a non-zero remainder, instead of letting an invalid mulcheck surface later during deferred verification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 434bd2dcc6bc3d1ce5489dd5643ed70d47430efb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/std/math/emulated/field_mul.go
+++ b/std/math/emulated/field_mul.go
@@ -534,8 +534,12 @@ func (f *Field[T]) callMulHint(a, b *Element[T], isMulMod bool, customMod *Eleme
 	if customMod != nil {
 		modulusLimbs = customMod.Limbs
 	}
-	hintInputs := make([]frontend.Variable, 0, 4+len(modulusLimbs)+len(a.Limbs)+len(b.Limbs))
-	hintInputs = append(hintInputs, nbBits, nbLimbs, len(a.Limbs), nbQuoLimbs)
+	isCheckZero := 0
+	if !isMulMod {
+		isCheckZero = 1
+	}
+	hintInputs := make([]frontend.Variable, 0, 5+len(modulusLimbs)+len(a.Limbs)+len(b.Limbs))
+	hintInputs = append(hintInputs, nbBits, nbLimbs, len(a.Limbs), nbQuoLimbs, isCheckZero)
 	hintInputs = append(hintInputs, modulusLimbs...)
 	hintInputs = append(hintInputs, a.Limbs...)
 	hintInputs = append(hintInputs, b.Limbs...)
@@ -566,8 +570,9 @@ func mulHint(field *big.Int, inputs, outputs []*big.Int) error {
 	nbLimbs := int(inputs[1].Int64())
 	nbALen := int(inputs[2].Int64())
 	nbQuoLen := int(inputs[3].Int64())
-	nbBLen := len(inputs) - 4 - nbLimbs - nbALen
-	ptr := 4
+	isCheckZero := inputs[4].Int64() == 1
+	nbBLen := len(inputs) - 5 - nbLimbs - nbALen
+	ptr := 5
 	plimbs := inputs[ptr : ptr+nbLimbs]
 	ptr += nbLimbs
 	alimbs := inputs[ptr : ptr+nbALen]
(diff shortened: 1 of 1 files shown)


ivokub:
Thanks @shramee for the PR. I completely agree currently debugging emulated circuits is a bit cumbersome particularly because we do the deferred checks and the actual failing constraints do not match the call site in the circuit definition.

However, on the other hand we try to be not opinionated for circuit definition and solving, i.e. we want the users to be able to define unsolvable witness.

I think for debugging purposes we would have to make the emulated gadget a bit more helpful, i.e. we should track the call stack instead in `-tags debug` mode which points to concrete failing place in the circuit. This is for example how API works for native computation. And when we don't use the debug tag we still get the meaningless "constraint failed" without call stack information.

ivokub:
> Thanks @shramee for the PR. I completely agree currently debugging emulated circuits is a bit cumbersome particularly because we do the deferred checks and the actual failing constraints do not match the call site in the circuit definition.
> 
> However, on the other hand we try to be not opinionated for circuit definition and solving, i.e. we want the users to be able to define unsolvable witness.
> 
> I think for debugging purposes we would have to make the emulated gadget a bit more helpful, i.e. we should track the call stack instead in `-tags debug` mode which points to concrete failing place in the circuit. This is for example how API works for native computation. And when we don't use the debug tag we still get the meaningless "constraint failed" without call stack information.

I created a new issue #1773 for tracking the implementation and will address it separately in a more comprehensive manner.

PR #1760, opened 2026-05-04
perf(bls12-377): randomized Schwartz-Zippel in native Fp6 and Fp12

yelhousni:
# Description

This PR adds randomized Schwartz-Zippel (SZ) verification for Fp6 and Fp12 multiplications in the native BLS12-377 field tower, targeting SCS (Plonk) circuits. Instead of computing products via the Karatsuba tower, the prover witnesses the result and a quotient polynomial via a hint, then defers a polynomial identity check at a random Fiat-Shamir challenge point:

```
a(r) · b(r) = q(r) · P(r) + c(r)   in Fp
```

where P(X) = X⁶+5 for Fp6 and P(X) = X¹²+5 for Fp12. Since BLS12-377's base field is ~377 bits, each check provides ~372 bits of soundness — well above the 128-bit requirement.

Multiple checks (across all Fp6 and Fp12 multiplications in the circuit) are batched via random linear combination using a single commitment, amortizing the challenge derivation cost.

## Type of change

### Changes

**Fp12 operations** (`e12.go`, `e12_pairing.go`, `e12_sz.go`):

| Operation | Before (SCS) | After (SCS, marginal) | Reduction |
|---|---:|---:|---|
| E12.Mul | 264 | 128 | 51% |
| E12.Square | 180 | 106 | 41% |
| E12.MulBy034 | 131 | 102 | 22% |
| E12.MulBy01234 | 237 | 112 | 53% |

**Fp6 operations** (`e6.go`, `e6_sz.go`):

| Operation | Before (SCS) | After (SCS) | Reduction |
|---|---:|---:|---|
| E6.Mul | 78 | 67 | 14% |
| E6.Square | 44 | 57 | ❌ regression |
| E6.MulBy01 | 53 | 63 | ❌ regression |

Note: E6.Square and E6.MulBy01 show a standalone regression because the SZ overhead (commitment, power precomputation) isn't amortized. In the pairing circuit where many E6 operations share one commitment, the amortized cost drops significantly:

**Pairing circuits:**

| Circuit | Before (SCS) | After (SCS) | Reduction |
|---|---:|---:|---|
| Torus pairing (default) | 21,273 | 17,830 | **16.2%** |
| Classical pairing (Fp12) | 51,292 | 36,681 | **28.5%** |

**Design:**
- SZ dispatch is gated on `frontendtype.SCS` + `frontend.Committer` — R1CS and SCS without commitment support fall back to Karatsuba (no regression)
- Tower-to-monomial permutations verified with Sage against gnark-crypto native arithmetic
- Sparse multiplications (MulBy034, MulBy01, MulBy01234) exploit zero coefficients in the evaluation to save gates
- Separate E6 and E12 checkers batch independently via `multicommit.WithCommitment`

# How has this been tested?

**SZ correctness tests** (10 tests, all verify against gnark-crypto native arithmetic with random inputs + reject wrong witnesses):
- `TestE12MulSZCorrectness` — E12.Mul
- `TestE12MulSZMultiple` — batched E12.Mul (4 operations, shared commitment)
- `TestE12SquareSZCorrectness` — E12.Square
- `TestE12MulBy034SZCorrectness` — E12.MulBy034 (sparse)
- `TestE12MulBy01234SZCorrectness` — E12.MulBy01234 (sparse)
- `TestE6MulSZCorrectness` — E6.Mul
- `TestE6SquareSZCorrectness` — E6.Square
- `TestE6MulBy01SZCorrectness` — E6.MulBy01 (sparse)

**Karatsuba non-regression** (existing tests via `test.CheckCircuit`, which uses plain SCS without Committer → falls through to Karatsuba):
- `go test -short ./std/algebra/native/fields_bls12377/...` — all pass
- `go test -short ./std/algebra/native/sw_bls12377/...` — all pairing tests pass

Constraint counts measured via `frontend.CompileGeneric[constraint.U64]` with `scs.NewBuilder` on Apple M5.

Per-operation marginal costs (N=8, amortized across shared commitment):

| Operation | Before | After | Marginal |
|---|---:|---:|---:|
| E12.Mul | 264 | — | 128 |
| E12.Square | 180 | — | 106 |
| E12.MulBy034 | 131 | — | 102 |
| E12.MulBy01234 | 237 | — | 112 |
| E6.Mul | 78 | — | 25 (amortized) |

End-to-end pairing benchmarks:

| Circuit | Before | After |
|---|---:|---:|
| BLS12-377 torus pairing check | 21,273 | **17,830** |
| BLS12-377 classical pairing | 51,292 | **36,681** |

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Alters core BLS12-377 field arithmetic in-circuit by swapping explicit multiplication constraints for hint-based Schwartz-Zippel checks under SCS+commitments; correctness and soundness depend on the new hint implementations and deferred batching logic.
> 
> **Overview**
> Adds a Schwartz-Zippel (hint + deferred polynomial identity) fast path for native BLS12-377 `E6` and `E12` operations when compiling to `frontendtype.SCS` with `frontend.Committer`, falling back to the existing Karatsuba/Toom-Cook implementations otherwise.
> 
> Introduces new SZ hint implementations and a shared, deferred batching checker (`sz.go`) that uses `multicommit.WithCommitment` to derive a Fiat–Shamir challenge and aggregate multiple multiplication/squaring checks into a single constraint.
> 
> Covers the new paths with dedicated correctness/negative tests and updates registered hints; internal pairing constraint stats are updated (notably `pairing_bls12377` on Plonk).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6412569278de75a0b3db367c9fdbc92e8e165f36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/std/algebra/native/fields_bls12377/e12_sz.go
+++ b/std/algebra/native/fields_bls12377/e12_sz.go
@@ -6,14 +6,9 @@ import (
 	"sync"
 
 	bls12377 "github.com/consensys/gnark-crypto/ecc/bls12-377"
-	"github.com/consensys/gnark/constraint/solver"
 	"github.com/consensys/gnark/frontend"
 	"github.com/consensys/gnark/std/multicommit"
 )
-
-func init() {
-	solver.RegisterHint(mulE12SZHint, squareE12SZHint, mulBy034E12SZHint)
-}
 
 // Schwartz-Zippel based E12 multiplication for SCS.
 //
@@ -156,8 +151,8 @@ func (ch *e12SZChecker) resolve(api frontend.API) error {
 		// P(r) = r^12 + 5
 		pEval := api.Add(rPow[12], 5)
 
-		// Batching coefficient: α = r^13
-		alpha := api.Mul(rPow[12], r)
+		// Batching coefficient: r^23 ensures non-overlapping degree ranges (deg(e_i)=22)
+		alpha := api.Mul(rPow[12], rPow[11]) // r^23
 
 		lhsAcc := frontend.Variable(0)
 		rhsAcc := frontend.Variable(0)
@@ -209,12 +204,11 @@ func evalAtPowers12(api frontend.API, coeffs []frontend.Variable, rPow []fronten
 }
 
 // evalSparse12 evaluates a polynomial where only some coefficients are nonzero.
-// varIdx: indices with circuit variable coefficients (need api.Mul)
-// constIdx: indices with constant coefficients (api.Mul by constant is free, just api.Add)
+// varIdx: indices with circuit variable coefficients (multiplication gate required)
+// constIdx: indices with constant coefficients (scalar multiplication, no gate in SCS)
 func evalSparse12(api frontend.API, coeffs []frontend.Variable, rPow []frontend.Variable, varIdx, constIdx []int) frontend.Variable {
 	var result frontend.Variable = 0
 	for _, i := range constIdx {
-		// constant * r^i: the Mul is free (constant multiplication)
 		result = api.Add(result, api.Mul(coeffs[i], rPow[i]))
 	}
 	for _, i := range varIdx {
--- a/std/algebra/native/fields_bls12377/e12_sz.go
+++ b/std/algebra/native/fields_bls12377/e12_sz.go
@@ -3,11 +3,9 @@ package fields_bls12377
 import (
 	"fmt"
 	"math/big"
-	"sync"
 
 	bls12377 "github.com/consensys/gnark-crypto/ecc/bls12-377"
 	"github.com/consensys/gnark/frontend"
-	"github.com/consensys/gnark/std/multicommit"
 )
 
 // Schwartz-Zippel based E12 multiplication for SCS.
@@ -23,7 +21,8 @@ import (
 // BLS12-377's base field is ~377 bits, so a single Fp evaluation point gives
 // ~372 bits of soundness (deg(a*b)/p = 22/p ≈ 2^{-372}), well above 128 bits.
 //
-// Gate cost: ~49 SCS gates per multiplication (vs ~264 for Karatsuba).
+// The exact SCS cost depends on commitment inputs and batching; see the
+// constraint-count tests for current numbers.
 
 // towerToMonomial12 reorders tower-basis coefficients to monomial-basis.
 //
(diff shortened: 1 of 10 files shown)


std/algebra/native/fields_bls12377/e6_sz.go
```diff
@@ -0,0 +1,354 @@
...
+	ch.checks = append(ch.checks, e6MulCheck{a: a, b: b, c: c, q: q, bSparseIdx: sparseIdx})
+}
+
+func (ch *e6SZChecker) resolve(api frontend.API) error {
+	if len(ch.checks) == 0 {
+		return nil
+	}
+
+	var toCommit []frontend.Variable
+	for i := range ch.checks {
```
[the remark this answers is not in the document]
yelhousni: The challenge r is derived from the full transcript, so r already implicitly binds to a and b

std/algebra/native/fields_bls12377/e12_sz.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+// resolve performs all accumulated checks using a single commitment challenge.
+func (ch *e12SZChecker) resolve(api frontend.API) error {
+	if len(ch.checks) == 0 {
+		return nil
+	}
+
+	var toCommit []frontend.Variable
+	for i := range ch.checks {
```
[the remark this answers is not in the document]
yelhousni: The challenge r is derived from the full transcript, so r already implicitly binds to a and b

gbotrel:
Review pass from 2026-05-07.

Findings:
- Blocking: GitHub currently reports this PR as `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, so it needs a rebase before merge.
- Test coverage gap: the negative SZ tests in `std/algebra/native/fields_bls12377/e6_sz_test.go` and `e12_sz_test.go` mostly keep the honest hint path and only replace the asserted public result with a random wrong result. That is useful, but it does not directly exercise the malicious-witness case this optimization is meant to defend against: a bad quotient/evaluation witness that tries to satisfy the final result. I did not find a concrete algebraic break, but I would add at least one test that overrides the hint output/replacement hint and proves the SZ check rejects a forged quotient independent of the final `AssertIsEqual`.

Verification run locally:
- `go test -short ./std/algebra/native/fields_bls12377/... ./std/algebra/native/sw_bls12377/...`

std/algebra/native/fields_bls12377/e6_sz.go
```diff
@@ -0,0 +1,354 @@
...
+type e6SZChecker struct {
+	checks []e6MulCheck
+}
+
+var (
+	e6szMu       sync.Mutex
+	e6szCheckers = map[frontend.Compiler]*e6SZChecker{}
+)
+
+func getE6SZChecker(api frontend.API) *e6SZChecker {
```
ivokub: I do agree with the Copilot comment here. Using the global variable here is not very efficient approach -- because we store `frontend.Compiler` as a map key, then we never let the GC to clean it. Better is to use the kvstore approach:
```go
type e6CheckerKey struct {} // defines a unique key type which prevents retrieving the same value outside of this package

func getE6Checker(api frontend.API) *e6SZChecker {
    kvstorer, ok := api.Compiler().(kvstore.Store)
    if !ok {
        // we handle the case when the compiler doesn't have local kv-storage. This doesn't happen normally unless we compile gnark circuits with a custom frontend.
        panic("compiler does not implement kvstore.Store")
    }
    checker := kvstorer.Get(e6CheckerKey{}) // gets the unique object
    if checker != nil {
        // we have the checker stored, return it
        if checkerT, ok := checker.(*e6SZChecker); !ok {
            // sanity check on the type
            panic("invalid type for stored checker")
        }
        return checker
    }
    checker = *e6SZChecker
    // schedules calling checker.check
    api.Complier().Defer(checker.resolve)
    kvstorer.Store(e6CheckerKey{}, checker)
    return checker
}
```

std/algebra/native/fields_bls12377/e6_sz.go
```diff
@@ -0,0 +1,354 @@
...
+
+func (ch *e6SZChecker) resolve(api frontend.API) error {
+	if len(ch.checks) == 0 {
+		return nil
+	}
+
+	var toCommit []frontend.Variable
+	for i := range ch.checks {
+		toCommit = append(toCommit, ch.checks[i].c[:]...)
+		toCommit = append(toCommit, ch.checks[i].q[:]...)
```
ivokub: Here I agree with Copilot again, we need to explicitly give a and b as well as inputs here -- r is computed using a hint from a and b, so there is no strong binding on a and b yet.

For example lets say the computed commitment for given `c = 0 and q=0` is `r`. Then we define
```
Let P(X) = X^6 + 5, and let r be the challenge derived from the current committed data. Since the code commits only c and q, choose:

c(X) = 0
q(X) = 0
b(X) = 1
a(X) = X - r
Then the deferred SZ check passes:

a(r) * b(r) = (r - r) * 1 = 0
q(r) * P(r) + c(r) = 0
But the actual E6 multiplication relation is false:

a(X) * b(X) = X - r != 0 = c(X)   in Fp[X]/(X^6 + 5)
In the PR’s tower layout, a(X) = X - r corresponds to:

monomial coeffs: [-r, 1, 0, 0, 0, 0]
tower coeffs:    [-r, 0, 1, 0, 0, 0]
So as an E6 value:

A.B0.A0 = -r
A.B0.A1 = 0
A.B1.A0 = 1
A.B1.A1 = 0
A.B2.A0 = 0
A.B2.A1 = 0

B = 1
C = 0
Q = 0
```

std/algebra/native/fields_bls12377/e6.go
```diff
@@ -93,12 +93,44 @@ func (e *E6) Mul(api frontend.API, e1, e2 E6) *E6 {
...
+	assignE6(e, out[:6])
+
+	var q [5]frontend.Variable
+	copy(q[:], out[6:11])
+
+	aMono := towerToMonomial6(aCoeffs)
+	bMono := towerToMonomial6(bCoeffs)
+	cMono := towerToMonomial6(e6Coeffs(e))
+
+	ch := getE6SZChecker(api)
```
ivokub: I think we could merge `getE6SZChecker` and `ch.AddCheck` into a single function for a bit of brevity.

std/algebra/native/fields_bls12377/e12_sz.go
```diff
@@ -0,0 +1,432 @@
...
+	a, b, c    [12]frontend.Variable // monomial coefficients (degree 11)
+	q          [11]frontend.Variable // quotient monomial coefficients (degree 10)
+	isSquare   bool                  // if true, b is ignored and a(r)² is checked
+	bSparseIdx []int                 // if non-nil, only these indices of b are nonzero variables
+	bConstIdx  []int                 // indices of b that are constants (e.g., 1)
+}
+
+// e12SZChecker accumulates E12 multiplication checks and resolves them
+// in a single deferred callback using a shared Fiat-Shamir challenge.
+type e12SZChecker struct {
```
ivokub: I think we could merge the e6SZChecker and e12SZChecker -- the difference seems only on the degrees of polys we evaluate, but this can be deduced by lengths of `a, b, c`.

std/algebra/native/fields_bls12377/e12_sz.go
```diff
@@ -0,0 +1,432 @@
...
+	e.C1.B2.A0, e.C1.B2.A1 = v[10], v[11]
+}
+
+// e12MulCheck stores one deferred a*b=c check in monomial form.
+type e12MulCheck struct {
+	a, b, c    [12]frontend.Variable // monomial coefficients (degree 11)
+	q          [11]frontend.Variable // quotient monomial coefficients (degree 10)
+	isSquare   bool                  // if true, b is ignored and a(r)² is checked
+	bSparseIdx []int                 // if non-nil, only these indices of b are nonzero variables
+	bConstIdx  []int                 // indices of b that are constants (e.g., 1)
```
ivokub: Do we need to keep track of `bSparseIdx` and `bConstIdx`? We can deduce what is constant by `api.Compiler().ConstantValue`? And imo the compiler should optimize out multiplications by constant (particularly with constant 0). If it doesn't, then I would say it is a bug in the compiler already :)

ivokub (changes requested):
Quick review -- soundness issue and global map are the main ones, but we can also imo simplify the deferred checker by unifying the E12 and E6 case and depend on compiler optimizations.

Didn't thoroughly check the hint definition and SZ evaluation yet, agentic review said it is fine, but will do another pass after the observed issues are fixed. Keep in mind that this adds overhead as every input to api.Commit creates additional constraint as we need to copy the value to the commitment column, so it is also important to rerun the benchmarks.

std/algebra/native/fields_bls12377/e6_sz.go
```diff
@@ -0,0 +1,217 @@
...
+	getNativeE6(&c, outputs[:6])
+
+	aMono := e6TowerToMonomialBigInt(&a)
+	bMono := e6TowerToMonomialBigInt(&b)
+
+	p := bls12377.ID.BaseField()
+	var prod [11]big.Int
+	for i := 0; i < 6; i++ {
+		for j := 0; j < 6; j++ {
+			var t big.Int
```
ivokub: Wouldn't it be more efficient to use `fp.Element` here? I also think a small comment would be useful i.e. "compute quotient such that a * b = c + q*p`

ivokub: Same goes for other hints.

std/algebra/native/fields_bls12377/e12_sz_test.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+}
+
+func TestE12MulSZConstraintCount(t *testing.T) {
+	field := ecc.BW6_761.ScalarField()
+	t.Log("=== E12 Multiplication SCS Constraint Counts (Schwartz-Zippel) ===")
+
+	for _, n := range []int{1, 2, 4, 8, 16} {
+		circuit := newE12MulNCircuit(n)
+		ccs, err := frontend.CompileGeneric[constraint.U64](field, scs.NewBuilder, circuit)
```
ivokub: Here you could just use `frontend.Compile(field, scs.NewBuilder, circuit)` it is an alias to `frontend.CompileGeneric[constraint.U64]`.

ivokub (dismissed):
Looks good to me. See the few comments but they're mostly cosmetic, imo it is good to merge as is!

ivokub (approved):
LGTM!

PR #1761, opened 2026-05-05
perf: generalize sparse solver optimizations

gbotrel:
## Summary

- Generalizes the sparse solver optimizations through `internal/generator/backend/template/representations/solver.go.tmpl` and regenerates all curve-specific constraint solvers.
- Computes SparseR1CS L/R/O vectors during solving instead of doing a second post-solve decompression pass.
- Adds direct calldata paths for built-in sparse gates, generic hints, batch inversions, and selected native hint implementations.
- Adds focused solver benchmarks for the BLS12-377 ECMul workload and `std/evmprecompiles`, with results documented in `performance_solver.md`.

## Numbers

Main BLS12-377 ECMul workload:

| Benchmark | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| `BenchmarkPlonkECMul30Solve` | `483335808 ns/op` | `119930740 ns/op` | `4.03x` |

Generated solver sanity benchmark:

| Package | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| `constraint/bn254` | `796083 ns/op` | `646986 ns/op` | `1.23x` |
| `constraint/bls12-381` | `740042 ns/op` | `725944 ns/op` | `1.02x` |
| `constraint/bw6-761` | `968236 ns/op` | `979986 ns/op` | `0.99x` |

`std/evmprecompiles` solve benchmarks:

| Circuit / curve | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| `bn254_ecmul/bn254` | `27243917 ns/op` | `10625208 ns/op` | `2.56x` |
| `bn254_ecmul/bls12_377` | `10309917 ns/op` | `10714167 ns/op` | `0.96x` |
| `ecrecover/bn254` | `51427083 ns/op` | `18528292 ns/op` | `2.78x` |
| `ecrecover/bls12_377` | `18072583 ns/op` | `17132000 ns/op` | `1.05x` |
| `bls12381_g1_msm_2/bn254` | `177734333 ns/op` | `36841959 ns/op` | `4.82x` |
| `bls12381_g1_msm_2/bls12_377` | `38587625 ns/op` | `39090875 ns/op` | `0.99x` |

BLS12-377 precompile rows are mostly flat because that generated solver had already been optimized manually before this generalization pass. The BN254 rows show the effect of moving the implementation into generated code.

## Validation

```text
GOWORK=off go generate ./internal/generator/backend
GOWORK=off go test ./constraint/solver ./frontend/cs/scs -count=1
GOWORK=off go test ./std/evmprecompiles -count=1
GOWORK=off go test ./backend/plonk/bn254 ./backend/plonk/bls12-377 ./backend/plonk/bls12-381 ./backend/plonk/bw6-761 -run 'Test.*Solve|Test.*Prove|Test.*Serialization|TestAPI' -count=1
GOWORK=off go test ./constraint/bn254 ./constraint/bls12-381 ./constraint/bw6-761 ./constraint/babybear ./constraint/koalabear ./constraint/tinyfield ./constraint/grumpkin -run '^TestSerialization/reference_small$|^TestSolve$|^Test$' -count=1
GOWORK=off go test ./constraint/bn254 ./constraint/bls12-381 ./constraint/bw6-761 ./constraint/grumpkin -run '^$' -bench 'BenchmarkSolve/scs$' -benchtime=3x -count=1
GOWORK=off go test ./std/evmprecompiles -run '^$' -bench 'BenchmarkSolveEvmprecompiles' -benchtime=1x -count=1
GOWORK=off go test ./constraint/bls12-377 -run '^$' -bench '^BenchmarkPlonkECMul30Solve$' -benchtime=50x -count=1 -benchmem
git diff --check
```

## Note

Full generated constraint-package tests for the big curves still fail in `TestSerialization/gkr_cube` on a GKR blueprint round-trip mismatch. The same class of failure was present before this work on BLS12-377 and does not exercise these solver paths.

Confirmed on `r8a.24xlarge`: 

Checked origin/master after fetch at 2d41ad0 and PR head at 978be1a. Machine was AMD EPYC 9R45 / 96 vCPU, Go 1.26.2, GOWORK=off, -cpu=96. The new benchmark files are only on the PR, so I added only those test
  files to the origin/master worktree for the “before” run.

  Key Results
  | Benchmark | Before | After | r8a speedup |
  |---|---:|---:|---:|
  | BenchmarkPlonkECMul30Solve | 1,141,107,531 ns/op | 188,974,848 ns/op | 6.04x |
  | bn254_ecmul/bn254 | 61,699,689 ns/op | 24,844,946 ns/op | 2.48x |
  | bn254_ecmul/bls12_377 | 62,583,975 ns/op | 25,066,084 ns/op | 2.50x |
  | ecrecover/bn254 | 106,005,572 ns/op | 41,691,534 ns/op | 2.54x |
  | ecrecover/bls12_377 | 106,122,650 ns/op | 41,579,889 ns/op | 2.55x |
  | bls12381_g1_msm_2/bn254 | 379,676,882 ns/op | 73,403,984 ns/op | 5.17x |
  | bls12381_g1_msm_2/bls12_377 | 394,324,855 ns/op | 72,790,309 ns/op | 5.42x |

  The generated solver sanity benchmark is effectively flat on this machine when run longer (100x): bn254 1.01x, bls12-381 0.98x, bw6-761 1.05x. The original 3x setting was too noisy.

  Finding
  The main PR claim holds, and the r8a machine shows a larger ECMul speedup than the Mac numbers (6.04x vs 4.03x). The EVM precompile results are also strongly positive, including the BLS12-377 rows that were flat
  in the original PR table.

---

> [!NOTE]
> **Medium Risk**
> Touches core constraint solving paths across multiple curves, including new fast-path arithmetic and hint execution; mistakes could cause incorrect witness assignments or unsatisfied-constraint behavior. Changes are largely performance-oriented with generic fallbacks, reducing but not eliminating correctness risk.
> 
> **Overview**
> **Speeds up SparseR1CS solving across generated curve solvers** by executing common sparse gates directly from calldata, avoiding decompression and extra passes.
> 
> The solver now **builds PLONK `L/R/O` small-domain vectors during solving** (with public-row initialization) and returns them directly, removing the post-solve `evaluateLROSmallDomain` reconstruction.
> 
> Hint handling is refactored to a calldata-based path with improved scratch-buffer reuse, plus **native in-solver implementations for selected std hints** (log-derivative count/batch-div, rangecheck decompose, bits `nBits`) with a safe big.Int fallback when not applicable or user-overridden. A new BLS12-377 ECMul solve benchmark is added to measure the impact.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5cc3578303928e8fbe99209724d225c67592261d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/generator/backend/template/representations/solver.go.tmpl
```diff
@@ -72,51 +92,116 @@ func newSolver(cs *system, witness fr.Vector, opts ...csolver.Option) (*solver,
...
+	}
+	v[0] = value
+	for n := 1; n < len(v); n *= 2 {
+		copy(v[n:], v[:n])
+	}
+}
 
 func (s *solver) set(id int, value fr.Element) {
-	if s.solved[id] {
+	if debug.Debug && s.solved[id] {
```
ivokub: `set` only rejects an already-solved wire when `debug.Debug` is enabled. The old unconditional check plus solved-count sanity check caught malformed/custom solver paths that assign the same wire twice; now release builds silently overwrite the value and `checkAllSolved` still passes. I’d keep this guard outside debug or restore an equivalent final consistency check.

internal/generator/backend/template/representations/system.go.tmpl
```diff
@@ -95,8 +95,9 @@ func (cs *system) Solve(witness witness.Witness, opts ...csolver.Option) (any, e
 	} else {
 		// sparse R1CS
 		var res SparseR1CSSolution
-		// query l, r, o in Lagrange basis, not blinded
-		res.L, res.R, res.O = evaluateLROSmallDomain(cs, solver.values)
+		res.L = solver.l
```
ivokub, line 98: Now that `Solve` returns `solver.l/r/o` directly, `evaluateLROSmallDomain` below is dead code in the template and all generated `system.go` files. This should be removed to avoid maintaining two LRO paths.

constraint/bls12-377/solver_ecmul_benchmark_test.go
```diff
@@ -0,0 +1,148 @@
...
+func BenchmarkPlonkECMul30Solve(b *testing.B) {
+	const nInstances = 30
+	spr := compileECMulSparseR1CS(b, nInstances)
+	fullWitness := buildECMulWitness(b, nInstances)
+	b.ReportMetric(float64(spr.GetNbConstraints()), "constraints")
+	b.ReportMetric(float64(len(spr.Levels)), "levels")
+	b.ReportMetric(float64(len(spr.Instructions)), "instructions")
+	b.ReportMetric(float64(len(fullWitness.Vector().(bls12377fr.Vector))), "witness_values")
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
ivokub: The new benchmarks still use `for i := 0; i < b.N; i++`. Since this PR targets modern Go, these should move to `for b.Loop() { ... }`; same pattern appears in the other new benchmark file.

ivokub (commented):
Posted the three concrete findings from the local review.

constraint/bls12-377/solver.go
```diff
@@ -29,22 +28,42 @@ type solver struct {
...
+const (
+	nativeHintNone nativeHint = iota
+	nativeHintLogderivargCount
+	nativeHintLogderivargBatchDivBySub
+	nativeHintRangecheckDecompose
+	nativeHintBitsNBits
+	nativeHintBsb22CommitmentPlaceholder
+)
+
+const (
```
ivokub: Hmm, I think this approach is not future proof when we rename the hints for example.

Couldn't we define the "native" hints as blueprints instead and then register at the corresponding packages? Imo right now having the implementations here and having then actual hints defined elsewhere is definitely to cause issues where we make changes there and they don't have effect as will be replaced in the solver.

constraint/bls12-377/solver.go
```diff
@@ -373,11 +508,59 @@ func (s *solver) Read(calldata []uint32) (constraint.U64, int) {
...
+	j := 3
+	for i := 0; i < n; i++ {
+		j += 1 + 2*int(calldata[j])
+	}
+	return j
+}
+
 // processInstruction decodes the instruction and execute blueprint-defined logic.
 // an instruction can encode a hint, a custom constraint or a generic constraint.
 func (s *solver) processInstruction(pi constraint.PackedInstruction, scratch *scratch) error {
```
ivokub, line 397: Looks good, but now becomes a bit difficult to follow, can we perhaps separate into dedicated SparseR1CS and R1CS cases and then inline swithc between the blueprint type?

constraint/bls12-377/solver.go
```diff
@@ -396,16 +579,574 @@ func (s *solver) processInstruction(pi constraint.PackedInstruction, scratch *sc
...
+		return s.solveRangecheckDecomposeHint(calldata, scratch)
+	case nativeHintBitsNBits:
+		return s.solveBitsNBitsHint(calldata)
+	case nativeHintBsb22CommitmentPlaceholder:
+		return s.solveBsb22CommitmentPlaceholderHint(hintID, calldata, scratch)
+	}
+	return s.solveGenericHintBigInt(hintID, calldata, scratch)
+}
+
+func (s *solver) solveGenericHintBigInt(hintID csolver.HintID, calldata []uint32, scratch *scratch) error {
```
ivokub: I don't understand why we now have two similar methods `solveWithHint` and `solveGenericHint` -- it seems the difference is that solveWithHint it gets as input constraint.HintMapping which has []LinearExpression as inputs and this uses callData as input. But they are related though, we can construct []LinearExpression from calldata directly no, so they are essentially the same?

constraint/bls12-377/solver.go
```diff
@@ -373,11 +508,59 @@ func (s *solver) Read(calldata []uint32) (constraint.U64, int) {
...
+		j += 1 + 2*int(calldata[j])
+	}
+	return j
+}
+
 // processInstruction decodes the instruction and execute blueprint-defined logic.
 // an instruction can encode a hint, a custom constraint or a generic constraint.
 func (s *solver) processInstruction(pi constraint.PackedInstruction, scratch *scratch) error {
 	// fetch the blueprint
 	blueprint := s.Blueprints[pi.BlueprintID]
```
ivokub, line 398: Here we obtain the blueprint from the map based on BlueprintID, but then later processSparseR1CInstruction we type switch on blueprint type.

And perhaps I missed something, but we don't call the Solve method on the blueprint itself? Why define the implementation in solveSparseR1CAdd etc methods?

constraint/bls12-377/solver.go
```diff
@@ -642,6 +1389,101 @@ func (s *solver) wrapErrWithDebugInfo(cID uint32, err error) *UnsatisfiedConstra
...
+	inputValues  []big.Int
+	inputs       []*big.Int
+	inputWords   []big.Word
+	outputValues []big.Int
+	outputs      []*big.Int
+}
+
+func (s *scratch) frElements(n int) []fr.Element {
+	if cap(s.elements) < n {
+		s.elements = make([]fr.Element, n)
```
ivokub: Imo conventional would be to use `slices.Grow` perhaps

ivokub (changes requested):
There is some leftover which is better to remove. And imo the growing implementation is becoming quite difficult to follow, we have many nested conditions and I'm not sure everything is covered.

And I also particularly dislike the idea of defining the typed methods for solving here instead of the packages where the hints are defined (as blueprints). I'm quite certain it will lead to problems in the future (when we change/fix some hint but it doesn't actually propagate to this function).

And now the solver is getting fairly complicated to follow, my first pass + agentic review said the solver is correct. In any case any bugs would come up with invalid proofs.

std/evmprecompiles/solver_benchmark_test.go
```diff
@@ -0,0 +1,180 @@
...
+	bn254fr "github.com/consensys/gnark-crypto/ecc/bn254/fr"
+	secp256k1ecdsa "github.com/consensys/gnark-crypto/ecc/secp256k1/ecdsa"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/frontend/cs/scs"
+	"github.com/consensys/gnark/std/algebra/emulated/sw_bls12381"
+	"github.com/consensys/gnark/std/algebra/emulated/sw_emulated"
+	"github.com/consensys/gnark/std/math/emulated"
+)
+
+func BenchmarkSolveEvmprecompiles(b *testing.B) {
```
ivokub, line 21: And perhaps we could move the benchmarks closer to the solver? I.e. this one and solver_ecmul_benchmark_test.go

ivokub (approved):
Thanks for addressing the comments. Looks good!

PR #1763, opened 2026-05-06
perf: use lattice scalar decomposition for tighter scalar-mul bounds

yelhousni:
# Description

This PR replaces the previous HalfGCD/Eisenstein scalar decomposition hints with lattice-backed rational reconstruction from `gnark-crypto/algebra/lattice`.

The affected paths are the existing scalar multiplication implementations:

- emulated short Weierstrass `scalarMulFakeGLV` and `scalarMulGLVAndFakeGLV`
- native BLS12-377 G1 `scalarMulGLVAndFakeGLV`
- native twisted Edwards `scalarMulFakeGLV`

The algorithms and public APIs stay the same. The PR changes the decomposition backend:

- `halfGCD` -> `rationalReconstruct`
- `halfGCDEisenstein` -> `rationalReconstructExt`

Because the lattice reconstruction has a proven LLL/Hermite bound, the GLV+fakeGLV sub-scalar width is tightened from the previous heuristic margin to `(r.BitLen()+3)/4 + 2`, saving several inner-loop iterations in the affected scalar multiplications.

Soundness guards against trivial all-zero decompositions are preserved and the related regression tests were updated to use the renamed hints.

## Type of change

- [x] Optimization / performance improvement with no intended behavior change

# How has this been tested?

Focused test run:

```bash
go test -short ./std/algebra/emulated/sw_emulated ./std/algebra/native/sw_bls12377 ./std/algebra/native/twistededwards ./internal/stats
```

Specifically covered:

- [x] emulated SW GLV/fakeGLV scalar multiplication tests
- [x] native BLS12-377 G1 GLV+fakeGLV tests
- [x] native twisted Edwards scalar multiplication tests

`internal/stats/latest_stats.csv` was regenerated. The changed snippets are the GLV/fakeGLV scalar multiplication paths; non-GLV P256 snippets are unchanged.

| Snippet | master R1CS / SCS | this PR R1CS / SCS |
|---|---:|---:|
| `scalar_mul_G1_bn254` | 115934 / 381171 | 108168 / 355353 |
| `scalar_mul_G1_bn254_incomplete` | 55441 / 200004 | 51579 / 185916 |
| `scalar_mul_secp256k1` | 117264 / 385623 | 108204 / 355502 |
| `scalar_mul_secp256k1_incomplete` | 56125 / 202518 | 51619 / 186082 |
| `scalar_mul_P256` | unchanged | unchanged |

# Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core scalar-multiplication hint/decomposition logic used across multiple curves; while intended to be behavior-preserving, any mismatch in sign/normalization or bounds could cause subtle correctness or soundness issues.
> 
> **Overview**
> **Scalar-mul decomposition backend is replaced**: the emulated SW and native implementations switch from `halfGCD`/`halfGCDEisenstein` (and `eisenstein.HalfGCD`) to lattice-based `rationalReconstruct`/`rationalReconstructExt` from `gnark-crypto/algebra/lattice`, including updated sign/absolute-value conventions.
> 
> **GLV+fakeGLV loops are tightened** by using the proven Hermite/LLL bound to reduce the bit-width used for `(u1,u2,v1,v2)` (e.g. from `BitLen/4 + 9` to `(BitLen+3)/4 + 2`), cutting several iterations in affected scalar-multiplication circuits.
> 
> **Tests and stats are updated**: regression tests that replace hints now target the new hint IDs/names, and `internal/stats/latest_stats.csv` reflects lower constraint counts for BN254/secp256k1 scalar-mul snippets (P-256 unchanged).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60084c9998ccd7d0384b957515c5fe0a38f1ba55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gbotrel:
Review pass from 2026-05-07.

Findings:
- Blocking: GitHub currently reports this PR as `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, so it needs a rebase before merge.
- `DoubleBaseScalarMul` now explicitly requires nonzero scalars and non-identity inputs in `std/algebra/native/twistededwards/curve.go`, but the public interface in `std/algebra/native/twistededwards/twistededwards.go` still advertises the generic `[s1]p1+[s2]p2` operation. There is at least one existing caller that can hit the mismatch: `std/signature/eddsa/eddsa.go` passes the public-key-derived point `_A` without guarding against the TE identity. I added a temporary regression circuit locally and confirmed the optimized path fails on BN254 for `p1=identity, s1=1, s2=2` where the mathematical result should reduce to `[2]base`. Please either preserve the previous complete behavior with a fallback for degenerate inputs, or tighten the interface and all callers with explicit constraints/tests so invalid EdDSA keys produce the intended verification result instead of an unexpected unsatisfied circuit.

Verification run locally:
- `go test -short ./std/algebra/emulated/sw_emulated/... ./std/algebra/emulated/sw_bls12381/... ./std/algebra/native/sw_bls12377/... ./std/algebra/native/twistededwards/...`
- `go test -short ./std/algebra/emulated/sw_bn254/... ./std/algebra/emulated/sw_bw6761/...`
- `go test -short ./std/signature/... ./std/evmprecompiles/... ./std/commitments/kzg/... ./std/recursion/groth16/... ./std/recursion/plonk/...`

yelhousni:
@ivokub this one is minimal; it only replaces Eisenstein by lattices in existing methods. Good on my side to review.

ivokub (approved):
This PR looks good, though there is an old soundness bug still in emulated GLV/fakeGLV/GLV+fakeGLV -- when the hint returns the subscalars, we check that they add up to the initial scalar. But the issue is that we only use subset of the bits when doing scalar mul (corresponding to the expected non-zero bits).

For example:
```
r = P-256 scalar order
n = 128
H = 2^128
s = r - H - 1

malicious decomposition:
  s1 = H + 1
  s2 = 1
  sign = 0

malicious scalar-mul output:
  R = [r - 1]P = -P

low128(s1) = 1
low128(s2) = 1
So it checks:

[1]P + [1]R = P - P = O
```

I think this shouldn't be a blocker of this PR, but we should address it in a separate PR consistently over all curves and G1/G2. I think the best approach for solving it is actually in non-native package. We currently anyway range check the outputs coming from the hint to be in range (of the non-native modulus), but could optionally make the bound smaller. This actually should reduce the constraints.

PR #1764, opened 2026-05-08
perf: G2 lattice-based GLV+FakeGLV scalar-mul

yelhousni:
# Description

This PR is stacked on #1763 and contains the G2 follow-up only. It adds lattice-based GLV+FakeGLV scalar multiplication for emulated G2 curves:

- BLS12-381 G2 switches its default `ScalarMul` path to the 4-dimensional GLV+FakeGLV decomposition.
- BN254 G2 gets a `ScalarMul` implementation using the same GLV+FakeGLV structure.
- BW6-761 G2 gets a `ScalarMul` implementation using the same GLV+FakeGLV structure.

The implementation adds curve-specific scalar multiplication hints and lattice reconstruction hints, then verifies the hinted result in-circuit through the GLV/fakeGLV relation instead of trusting the hint output. It also adds stats snippets for the new G2 scalar multiplication paths.

This PR intentionally does not include the native twisted Edwards MSM work; that is split into #1765.

## Type of change

- [x] Optimization / performance improvement

# How has this been tested?

Focused test run:

```bash
go test -short ./std/algebra/emulated/sw_bls12381 ./std/algebra/emulated/sw_bn254 ./std/algebra/emulated/sw_bw6761 ./internal/stats
```

Specifically covered:

- [x] BLS12-381 G2 scalar multiplication tests
- [x] BN254 G2 scalar multiplication tests
- [x] BW6-761 G2 scalar multiplication tests

`internal/stats/latest_stats.csv` was regenerated with new G2 scalar multiplication snippets.

| Snippet | R1CS constraints / wires | SCS constraints / wires |
|---|---:|---:|
| `scalar_mul_G2_bls12381` | 138129 / 216643 | 481714 / 463233 |
| `scalar_mul_G2_bn254` | 102279 / 158428 | 349799 / 336420 |
| `scalar_mul_G2_bw6761` | 191105 / 296444 | 659865 / 637514 |

# Checklist:

- [x] New and existing focused tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules (`gnark-crypto/algebra/lattice`)

---

> [!NOTE]
> **High Risk**
> Large changes to in-circuit elliptic-curve scalar multiplication and hint soundness; incorrect decomposition, verification, or edge-case handling could break proof soundness for pairing/G2-heavy circuits.
> 
> **Overview**
> Adds **lattice-based GLV+FakeGLV** in-circuit **G2 `ScalarMul`** for emulated **BLS12-381**, **BN254**, and **BW6-761**, replacing BLS12-381’s prior 2D Halo/GLV default and introducing the path on BN254/BW6 where it was missing.
> 
> The new flow hints **`R = [s]Q`**, decomposes **−s** into four **~r^(1/4)** sub-scalars via **`rationalReconstructExtG2`**, and verifies **`[v1]R + [v2]Φ(R) + [u1]Q + [u2]Φ(Q) = O`** with a biased accumulator check against precomputed generator constants—not trusting hints alone. **Complete** mode uses **`AddUnified`**, unified doubling, and explicit handling of **s ∈ {0, ±1}** and infinity; **`WithIncompleteArithmetic`** is now a real faster path (previously ignored on BLS12-381 G2).
> 
> Also wires **`MultiScalarMul`** through the new scalar mul on BLS12-381, registers **G2 hint** functions, adds **stats snippets** and updated **`latest_stats.csv`**, and extends **edge-case tests** per curve.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c635573954c340e45c1540d48ad18c66b7764583. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


std/algebra/emulated/sw_bn254/g2.go
```diff
@@ -290,3 +335,345 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
+	y31 := g2.fp.Eval([][]*baseEl{{&λ2.A0, &y3.A1}, {&λ2.A1, &y3.A0}, {mone, &p.P.Y.A1}}, []int{1, 1, 1})
+	y3 = &fields_bn254.E2{A0: *y30, A1: *y31}
+
+	return &G2Affine{
+		P: g2AffP{
+			X: *x3,
+			Y: *y3,
+		},
+	}
+}
```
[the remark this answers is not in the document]
yelhousni: Fixed in 342115794 by removing the unused BN254 G2 triple helper.

std/algebra/native/twistededwards/hints.go
```diff
@@ -63,40 +64,55 @@ func decomposeScalar(scalarField *big.Int, inputs []*big.Int, res []*big.Int) er
...
+	k.Sub(x, k)
+	k.Quo(k, inputs[1])
+	outputs[3].Set(k)
 
+	// s2 = −z, encoded as |s2| + signBit. signBit = 1 iff −z < 0 iff z > 0.
+	outputs[1].Abs(z)
 	outputs[2].SetUint64(0)
-	if outputs[1].Sign() == -1 {
-		outputs[1].Neg(outputs[1])
+	if z.Sign() > 0 {
```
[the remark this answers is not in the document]
yelhousni: I think this is a false positive. The hint encodes signBit=1 iff signed s2 is negative, and computes k=(x-z*s)/r. The circuit bit=1 branch asserts s1 = k*r + |s2|*s, equivalently s1 - |s2|*s = k*r, which matches x - z*s = k*r when z > 0. The bit=0 branch handles z < 0 with s1 + |s2|*s = k*r. Focused twistededwards tests pass.

yelhousni:
@ivokub this one is also ready for review (after https://github.com/Consensys/gnark/pull/1763)

std/algebra/emulated/sw_bn254/g2.go
```diff
@@ -290,3 +335,306 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
+func (g2 *G2) scalarMulGLVAndFakeGLV(Q *G2Affine, s *Scalar, opts ...algopts.AlgebraOption) *G2Affine {
+	cfg, err := algopts.NewConfig(opts...)
+	if err != nil {
+		panic(err)
+	}
+
+	// handle 0-scalar
+	var selector0 frontend.Variable
+	_s := s
+	if cfg.CompleteArithmetic {
```
ivokub: [P1] This keeps BN254 G2 scalar multiplication incomplete by default. `algopts.NewConfig()` leaves `CompleteArithmetic=false` unless the caller explicitly passes `WithCompleteArithmetic`, but `WithCompleteArithmetic` is deprecated now. This PR should follow the pattern from https://github.com/Consensys/gnark/pull/1751: arithmetic is complete by default, and callers opt into the incomplete path with `WithIncompleteArithmetic` when they can guarantee the exceptional cases are absent. As written, the default path still skips the `s=0`, `Q=(0,0)`, and `s=±1` substitutions and can hit the incomplete table additions.

std/algebra/emulated/sw_bw6761/g2.go
```diff
@@ -236,3 +271,295 @@ func (g2 *G2) AssertIsEqual(p, q *G2Affine) {
...
+func (g2 *G2) scalarMulGLVAndFakeGLV(Q *G2Affine, s *Scalar, opts ...algopts.AlgebraOption) *G2Affine {
+	cfg, err := algopts.NewConfig(opts...)
+	if err != nil {
+		panic(err)
+	}
+
+	// handle 0-scalar
+	var selector0 frontend.Variable
+	_s := s
+	if cfg.CompleteArithmetic {
```
ivokub: [P1] Same default-completeness issue here: this edge-case handling only runs when the deprecated `WithCompleteArithmetic` option is explicitly set. Per the pattern introduced in https://github.com/Consensys/gnark/pull/1751, scalar multiplication should be complete by default, with `WithIncompleteArithmetic` as the opt-out. Otherwise default BW6-761 G2 scalar multiplication still sends `s=0`, `Q=(0,0)`, and `s=±1` through the incomplete addition path.

ivokub (commented):
Two inline comments on BN254/BW6-761 default completeness.

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -607,204 +638,254 @@ func (g2 *G2) scalarMulGeneric(p *G2Affine, s *Scalar, opts ...algopts.AlgebraOp
...
+	// generator to avoid incomplete additions in the loop. At the end Acc
+	// will equal [2^(nbits-1)]G2 (the precomputed g2GenNbits).
+	Acc := g2.add(tableS[1], tablePhiS[1])
+	B1 := Acc
+	g2GenPoint := &G2Affine{P: *g2.g2Gen}
+	Acc = g2.add(Acc, g2GenPoint)
+
+	// LLL Hermite bound: u_i, v_i < γ₄·r^(1/4), fits in (BitLen+3)/4 + 2 bits.
+	nbits := (st.Modulus().BitLen()+3)/4 + 2
+	u1bits := g2.fr.ToBits(u1)
```
ivokub, line 839: [P1] The GLV/FakeGLV sub-scalar bounds are still trusted here, not constrained. The relation above constrains the full hinted emulated values `u1`, `u2`, `v1`, `v2`, but the accumulator check only consumes the low `nbits` bits returned by `ToBits`; the high bits are not constrained to zero. A malicious prover can therefore choose large hinted sub-scalars that satisfy the modular relation while making the MSM check use only a different low-bit relation. This is the same class of issue I commented on in https://github.com/Consensys/gnark/pull/1763#pullrequestreview-4267621967, and it also applies to the BN254 and BW6-761 G2 copies in this PR.

I think we can postpone fixing the bound issue in this PR, as in PR 1763, because this is a latent existing limitation in emulated arithmetic. The right fix should be a new feature in `std/math/emulated` to constrain an emulated element to an explicit bit bound, then use that here instead of trusting the hint output.

ivokub (commented):
Inline comment on the trusted emulated sub-scalar bounds issue.

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -607,204 +638,254 @@ func (g2 *G2) scalarMulGeneric(p *G2Affine, s *Scalar, opts ...algopts.AlgebraOp
 	return R0
 }
 
-// ScalarMul computes [s]Q using an efficient endomorphism and returns it. It doesn't modify Q nor s.
-// It implements an optimized version based on algorithm 1 of [Halo] (see Section 6.2 and appendix C).
+// ScalarMul computes [s]Q using GLV+FakeGLV with proven r^(1/4) sub-scalar
```
ivokub, line 650: [P2] We should document the G2 subgroup precondition for `ScalarMul` more explicitly. It may be obvious from the package context that these methods are intended for G2 subgroup points, but a minimal package-level or method-level comment would make the contract clear: callers must not pass an arbitrary point that is merely on the twist/curve. This helps avoid future misuse where someone calls scalar multiplication on a random on-curve point and then hits implementation-specific behavior across curves or scalar-mul implementations.

ivokub (commented):
Inline comment on documenting the G2 subgroup precondition.

ivokub (commented):
Inline comment on G2 scalar-mul edge-case test coverage.

std/algebra/emulated/sw_bls12381/g2_test.go
```diff
@@ -25,7 +25,7 @@ func (c *mulG2Circuit) Define(api frontend.API) error {
 	if err != nil {
 		return fmt.Errorf("new G2 struct: %w", err)
 	}
-	res1 := g2.scalarMulGLV(&c.In, &c.S)
+	res1 := g2.ScalarMul(&c.In, &c.S)
```
ivokub: [P2] Please add more edge-case coverage for the new G2 scalar-mul paths, similar to what we have/expect for G1. A single random scalar test is not enough for this change because the implementation has explicit branches for `s=0`, `s=1`, `s=-1`, and `Q=(0,0)`, plus the incomplete-arithmetic option path. The BN254 and BW6-761 G2 implementations should also have direct `ScalarMul` tests for these cases, not only pairing or seed-mul tests.

ivokub (changes requested):
In general looks good, but please see inline comments. Particularly, the implementation should now be complete by default to follow the convention as for G1.

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -607,212 +647,288 @@ func (g2 *G2) scalarMulGeneric(p *G2Affine, s *Scalar, opts ...algopts.AlgebraOp
...
+	lhs4 := g2.fr.Select(isNegu2, zero, λu2)
+	lhs := g2.fr.Add(g2.fr.Add(lhs1, lhs2), g2.fr.Add(lhs3, lhs4))
+
+	rhs1 := g2.fr.Select(isNegv1, sv1, zero)
+	rhs2 := g2.fr.Select(isNegv2, sλv2, zero)
+	rhs3 := g2.fr.Select(isNegu1, u1, zero)
+	rhs4 := g2.fr.Select(isNegu2, λu2, zero)
+	rhs := g2.fr.Add(g2.fr.Add(rhs1, rhs2), g2.fr.Add(rhs3, rhs4))
+
+	g2.fr.AssertIsEqual(lhs, rhs)
```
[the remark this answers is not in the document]
yelhousni: api.Select boolean-constrains the selector.

std/algebra/emulated/sw_bn254/g2.go
```diff
@@ -290,3 +420,305 @@ func (g2 *G2) IsEqual(p, q *G2Affine) frontend.Variable {
...
+		g2.fr.Add(rhs1, rhs2),
+		g2.fr.Add(rhs3, rhs4),
+	)
+
+	g2.fr.AssertIsEqual(lhs, rhs)
+
+	// Soundness: forbid the trivial all-zeros decomposition. With v1 + λ·v2 ≠ 0
+	// the relation forces (u1, u2) to the unique LLL lattice point and pins
+	// the hinted R = [s]Q.
+	g2.fr.AssertIsDifferent(g2.fr.Add(v1, g2.fr.Mul(g2.eigenvalue, v2)), g2.fr.Zero())
```
ivokub: [P1] This non-zero check is over the unsigned hinted limbs, but the MSM below uses the signed coefficient `+/-v1 + lambda*(+/-v2)` according to `isNegv1` and `isNegv2`. A malicious replacement hint can choose signs and absolute values so the signed coefficient of `R` is zero while this unsigned `v1 + lambda*v2` check still passes. Then the final accumulator no longer pins the hinted `R`, which can let an incorrect `[s]Q` output pass. Please assert non-zero on the signed coefficient that the MSM actually consumes.

std/algebra/emulated/sw_bls12381/g2.go
```diff
@@ -607,212 +647,288 @@ func (g2 *G2) scalarMulGeneric(p *G2Affine, s *Scalar, opts ...algopts.AlgebraOp
...
+
+	// Soundness: forbid the trivial all-zeros decomposition. With v1 + λ·v2 ≠ 0
+	// the relation forces (u1, u2) to the unique LLL lattice point and pins
+	// the hinted R = [s]Q.
+	g2.fr.AssertIsDifferent(g2.fr.Add(v1, g2.fr.Mul(g2.eigenvalue, v2)), g2.fr.Zero())
```
ivokub: [P1] Same soundness issue as in the BN254 copy: this checks the unsigned `v1 + lambda*v2`, while the group equation uses the signed coefficient selected by `isNegv1` and `isNegv2`. With adversarial hint outputs, the signed coefficient of `R` can be zero even though this unsigned check succeeds, leaving `R` unconstrained by the MSM verification. This should be checked using the signed coefficient that the MSM actually consumes.

ivokub (commented):
Inline comments on the signed R-coefficient non-zero check. I am skipping the sub-scalar range-bound finding here since it is tracked separately in #1767.

std/algebra/emulated/sw_bw6761/g2.go
```diff
@@ -236,3 +350,301 @@ func (g2 *G2) AssertIsEqual(p, q *G2Affine) {
...
+		g2.fr.Add(rhs1, rhs2),
+		g2.fr.Add(rhs3, rhs4),
+	)
+
+	g2.fr.AssertIsEqual(lhs, rhs)
+
+	// Soundness: forbid the trivial all-zeros decomposition. With v1 + λ·v2 ≠ 0
+	// the relation forces (u1, u2) to the unique LLL lattice point and pins
+	// the hinted R = [s]Q.
+	g2.fr.AssertIsDifferent(g2.fr.Add(v1, g2.fr.Mul(g2.eigenvalue, v2)), g2.fr.Zero())
```
ivokub: [P1] Same signed-coefficient issue here: the non-zero assertion ignores `isNegv1`/`isNegv2`, but the MSM uses `+/-v1 + lambda*(+/-v2)`. A malicious replacement hint can make the signed coefficient of the hinted `R` equal zero while this unsigned expression remains non-zero, so the accumulator check does not force `R = [s]Q`. Please assert non-zero on the signed coefficient used by the MSM.

ivokub (changes requested):
The previous issues are addressed. Now there is only one small issue still left.

ivokub (approved):
Thanks! All perfect now!

PR #1765, opened 2026-05-08
perf: twisted Edwards lattice-based double-base MSM

yelhousni:
# Description

This PR is stacked on #1763 and contains the native twisted Edwards MSM follow-up only. It adds lattice-based double-base scalar multiplication variants for native twisted Edwards curves:

- `DoubleBaseScalarMulNonZero` as the optimized non-degenerate path.
- MSM(3, 2n/3) for curves without a GLV endomorphism.
- MSM(6, n/3) for curves with a GLV endomorphism, including Bandersnatch endomorphism parameters.
- lattice reconstruction hints for the multi-scalar decomposition.
- in-circuit scalar decomposition verification helpers.

`DoubleBaseScalarMul` remains complete for the public API and handles zero scalars / identity edge cases before dispatching to the optimized non-zero path. The branch also includes the review fixes that remove redundant hint outputs, wire scalar decomposition verification into the twisted Edwards scalar multiplication path, and bind the non-GLV MSM hint to a single result with the sound relation `[x1]P1 + [x2]P2 - [z]R = O`.

This PR intentionally does not include the emulated G2 GLV+FakeGLV work; that is split into #1764.

## Type of change

- [x] Optimization / performance improvement

# How has this been tested?

Focused test run:

```bash
go test -short ./std/algebra/native/twistededwards ./internal/stats
```

Specifically covered:

- [x] native twisted Edwards scalar multiplication tests
- [x] native twisted Edwards double-base scalar multiplication tests
- [x] regression coverage for malicious trivial scalar decompositions
- [x] regression coverage for forged double-base MSM hint outputs

`internal/stats/latest_stats.csv` was regenerated with new MSM(2) snippets.

| Snippet | R1CS constraints / wires | SCS constraints / wires |
|---|---:|---:|
| `msm_G1_bn254_2` | 208925 / 312617 | 688811 / 658743 |
| `msm_P256_2` | 185846 / 288056 | 635297 / 608874 |
| `msm_secp256k1_2` | 208997 / 312737 | 689104 / 659028 |
| `msm_babyjubjub_2` | 5269 / 5683 | 12389 / 11848 |
| `msm_jubjub_2` | 5276 / 5754 | 12332 / 11855 |
| `msm_bandersnatch_2` | 5532 / 6301 | 13470 / 12918 |

# Checklist:

- [x] New and existing focused tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules (`gnark-crypto/algebra/lattice`)

---

> [!NOTE]
> **High Risk**
> Changes in-circuit elliptic-curve arithmetic and hint verification (subgroup binding, torsion resistance); incorrect constraints would be a proof soundness bug, though coverage targets known attacks.
> 
> **Overview**
> Adds **fast lattice-based double-base scalar multiplication** for native twisted Edwards curves while keeping **`DoubleBaseScalarMul` complete** by default; **`algopts.WithIncompleteArithmetic`** routes to **`DoubleBaseScalarMulNonZero`** (3-MSM + LogUp on non-GLV curves, 6-MSM + GLV φ on Bandersnatch).
> 
> **Soundness hardening** for hinted MSM/GLV paths: **`assertInSubgroup`** (`[cofactor]S == R` plus **on-curve** preimage `S`), extended **`scalarMulHint`** / new **`doubleBaseScalarMulHint`** and multi-rational reconstruction hints, and **emulated-field** decomposition checks in **`scalar_decomp.go`** (replacing redundant native checks and dropping the unused `k` output from **`rationalReconstruct`**).
> 
> Regression tests cover **trivial decompositions**, **forged partial MSM hints**, **2-torsion forgeries** (scalar mul and double-base, BN254 and Bandersnatch), and **off-curve subgroup preimages**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6908b38526a2934b5ed852cc3ee61a051f671af0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


std/algebra/native/twistededwards/point.go
```diff
@@ -181,3 +176,370 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+	}
+
+	// Verify accumulator equals identity (0, 1)
+	api.AssertIsEqual(res.X, 0)
+	api.AssertIsEqual(res.Y, 1)
+
+	// Return Q1 + Q2
+	p.add(api, &Q1, &Q2, curve)
+
+	return p
```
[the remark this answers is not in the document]
yelhousni: Fixed in 7b70e6700 by changing the non-GLV path to verify one hinted result R with the shared-denominator relation [x1]P1 + [x2]P2 - [z]R = O. Also added regression coverage for forged double-base MSM hint outputs and regenerated stats.

std/algebra/native/twistededwards/point.go
```diff
@@ -181,3 +176,329 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+
+	// Scalar bit length: ~r^(1/3) ≈ 85 bits for 254-bit order
+	n := (curve.Order.BitLen() + 2) / 3
+
+	bX1 := api.ToBinary(absX1, n)
+	bY1 := api.ToBinary(absY1, n)
+	bX2 := api.ToBinary(absX2, n)
+	bY2 := api.ToBinary(absY2, n)
+	bZ := api.ToBinary(absZ, n)
+	bT := api.ToBinary(absT, n)
```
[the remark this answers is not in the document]
yelhousni: Good catch. Fixed in d230b53ef by returning the six ToBinary bit slices from verifyScalarDecomposition6D and reusing them in doubleBaseScalarMul6MSMLogUp instead of decomposing the same witnesses twice. Regenerated stats now show msm_bandersnatch_2 dropping to 5016/5791 on groth16 and 12450/11904 on plonk. Verified with go test -short ./std/algebra/native/twistededwards ./internal/stats, goimports, and git diff --check.

std/algebra/native/twistededwards/point.go
```diff
@@ -131,26 +134,18 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
-	rhs := api.Select(bit, api.Add(_k, _s2), _k)
-	api.AssertIsEqual(lhs, rhs)
-	// A malicious hint can provide s1=s2=0, which makes the relation vacuous.
-	api.AssertIsEqual(api.IsZero(s2), 0)
-
-	n := (curve.Order.BitLen() + 1) / 2
-	b1 := api.ToBinary(s1, n)
-	b2 := api.ToBinary(s2, n)
+	b1, b2 := verifyScalarDecomposition(api, s1, s2, bit, checkedScalar, curve)
+	n := len(b1)
```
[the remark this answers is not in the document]
yelhousni: Valid cleanup. Fixed in 82edcb39f by changing rationalReconstruct to return only (s1, |s2|, signBit), updating scalarMulFakeGLV to request 3 outputs, and adjusting the forged-hint regression helper. Stats regeneration produced no latest_stats.csv diff. Verified with go test -short ./std/algebra/native/twistededwards ./internal/stats, goimports, and git diff --check.

yelhousni:
@ivokub this one is also ready for review (after https://github.com/Consensys/gnark/pull/1763)

ivokub (commented):
Audit notes from the twisted Edwards double-base scalar multiplication review. The main concern is that the optimized `DoubleBaseScalarMulNonZero` relation appears to characterize the result only modulo the prime-order subgroup, while the interface currently accepts generic on-curve twisted Edwards points.

std/algebra/native/twistededwards/curve.go
```diff
@@ -44,8 +45,29 @@ func (c *curve) ScalarMul(p1 Point, scalar frontend.Variable) Point {
...
 	return p
 }
+
+// DoubleBaseScalarMulNonZero computes s1*p1 + s2*p2 using the most efficient
+// lattice-based MSM variant available for the curve:
+//   - GLV-equipped curves (Bandersnatch): 6-MSM with r^(1/3)-bounded sub-scalars.
+//   - non-GLV curves (Jubjub, BabyJubjub, edBLS12-377, edBW6-761): 3-MSM with
+//     r^(2/3)-bounded sub-scalars and LogUp lookups.
+//
+// The scalars s1, s2 must be nonzero and p1, p2 must not be the TE identity
```
ivokub: I think the precondition is too weak for the optimized path. The verification in `DoubleBaseScalarMulNonZero` proves a scaled relation modulo `curve.Order`, but the twisted Edwards curves here have cofactors. Requiring only nonzero scalars and non-identity points still permits torsion components.

For example, if the decomposition uses an even nonzero denominator such as `z = 2`, a prover can return `R + T` with the rational 2-torsion point `T = (0, -1)`: `[2](R + T) = [2]R`, so the MSM check passes while the returned affine point is not the requested result. This should either require/enforce prime-subgroup inputs and result uniqueness, or avoid using the scaled relation as the sole binding check over the full curve group.

std/algebra/native/twistededwards/point.go
```diff
@@ -183,3 +176,321 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+	h, err := api.NewHint(multiRationalReconstructHint, 6, s1, s2, curve.Order)
+	if err != nil {
+		panic(err)
+	}
+	absX1, absX2, absZ := h[0], h[1], h[2]
+	signX1, signX2, signZ := h[3], h[4], h[5]
+
+	// Verify the decomposition using emulated arithmetic to avoid native field
+	// overflow. Also range-checks x1, x2, z and ensures z is non-zero.
+	bX1, bX2, bZ := verifyScalarDecomposition3D(api, s1, s2, absX1, absX2, absZ, signX1, signX2, signZ, curve)
```
ivokub, line 274: This decomposition check ensures `z != 0` modulo the subgroup order and that `x1`, `x2`, `z` match the scalars, but it does not make multiplication by `z` injective on the full Edwards group. Since `R` comes from hints and is only bound through `[z]R = [x1]P1 + [x2]P2`, any torsion component killed by `z` can be added to `R` without changing this check.

This is a circuit soundness issue unless all involved points are constrained to the prime-order subgroup or the torsion component is otherwise ruled out.

std/algebra/native/twistededwards/point.go
```diff
@@ -183,3 +176,321 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+	// Also range-checks sub-scalars and ensures the shared denominator is non-zero.
+	bX1, bY1, bX2, bY2, bZ, bT := verifyScalarDecomposition6D(api, s1, s2,
+		absX1, absY1, absX2, absY2, absZ, absT,
+		signX1, signY1, signX2, signY2, signZ, signT,
+		curve, endo,
+	)
+
+	// Compute φ(P1), φ(P2), φ(R)
+	var phiP1, phiP2, phiR Point
+	phiP1.phi(api, p1, curve, endo)
```
ivokub, line 383: The Bandersnatch GLV path applies `phi` with unchecked divisions to `P1`, `P2`, and the hinted `R`. The comment says the endomorphism acts as scalar multiplication by `Lambda` on the prime-order subgroup, but the public precondition does not require or enforce subgroup membership.

There is also an exceptional case when the final result is identity, for example `P2 = -P1` and equal nonzero scalars. Then `phi(R)` has denominator `x*y = 0`, so the optimized method is not complete for all nonzero scalar / non-identity point inputs currently described by the interface.

std/algebra/native/twistededwards/curve_test.go
```diff
@@ -296,6 +312,21 @@ func TestDoubleBaseScalarMul(t *testing.T) {
 	}
 }
 
+func TestDoubleBaseScalarMulNonZero(t *testing.T) {
```
ivokub, line 317: The current tests cover random nonzero inputs and one forged partial-hint case, but they do not cover the edge cases that matter for this optimization: torsion-shifted hinted results, non-prime-subgroup inputs, the `(0, -1)` torsion point, result-identity cases in the Bandersnatch GLV path, or forged decomposition hints.

I would add negative tests for those cases before treating the optimized path as sound. In particular, a test that tries to return `correctR + (0, -1)` with an even denominator decomposition would exercise the cofactor ambiguity directly.

std/algebra/native/twistededwards/curve.go
```diff
@@ -44,8 +45,29 @@ func (c *curve) ScalarMul(p1 Point, scalar frontend.Variable) Point {
...
+// (0, 1). Use DoubleBaseScalarMul for complete edge-case handling.
+func (c *curve) DoubleBaseScalarMulNonZero(p1, p2 Point, s1, s2 frontend.Variable) Point {
```
ivokub, line 87: And perhaps instead of defining new method `DoubleBaseScalarMulNonZero` we could call directly from `DoubleBaseScalarMul` when the `algopts.WithIncompleteArithmetic` option is set? We currently don't have it in that method signature, but can add as a variadic argument, it does not break API.

ivokub (changes requested):

yelhousni:
@ivokub Addressed the review + a related bug found while fixing:

- Soundness (cofactor torsion): the scaled relation [z]R = [x1]P1+[x2]P2 only pinned R mod the z-torsion, so an even z let a prover return R+(0,-1). Fixed by binding the hinted result into the prime-order subgroup: hint a preimage S and assert R = [cofactor]·S (power-of-two cofactor ⇒ [cofactor]·E = ⟨G⟩). ~12–17 constraints.
- GLV path: binding runs before φ, so φ(R)=[λ]R holds; documented the φ(identity) exception (result must be non-identity).
- API: DoubleBaseScalarMul now takes ...algopts.AlgebraOption and routes to the fast path under WithIncompleteArithmetic; documented subgroup/non-identity/nonzero preconditions.
- Tests: added torsion-rejection tests for the 3-MSM (even-z) and Bandersnatch 6-MSM paths; updated the forged-hint regression. Ablation-verified they catch the bug.
- Same bug in scalarMulFakeGLV: confirmed ScalarMul accepted [s]P+(0,-1) for even s2; applied the same [cofactor]·S binding + regression test. Note: this tightens ScalarMul's contract to require prime-order-subgroup input.
- Stats regenerated for the affected snippets; twistededwards/eddsa tests green.

std/algebra/native/twistededwards/point.go
```diff
@@ -183,3 +186,358 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+// twisted Edwards curve is a power of two, so multiplication by it annihilates
+// the entire 2-Sylow and [cofactor]·E equals the prime-order subgroup. Hence a
+// satisfying preimage S exists iff R is already in the subgroup; a
+// torsion-shifted R (e.g. R + (0,-1)) has no preimage and is rejected.
+//
+// This is the binding that makes the scaled MSM relation
+// [z]R = [x1]P1 + [x2]P2 sound over the full (cofactored) group: on its own
+// that relation only pins R modulo the z-torsion, so an even denominator z
+// would leave the 2-torsion component of R unconstrained.
+func assertInSubgroup(api frontend.API, R, S *Point, curve *CurveParams) {
```
ivokub, line 225: It is better but still not complete. The check `[cofactor]S == R` assumes that `S` and `R` are on curve, but we never check that (we just get them hinted). See comment below:
>`assertInSubgroup` should also constrain the hinted preimage `S` to be on-curve before using it.
>
>The argument in the comment is group-theoretic: `[cofactor]S == R` proves `R` is in the prime-order subgroup only if `S` is known to be a curve point. In the circuit, `S` comes from a hint, so without an on-curve constraint it is just an arbitrary field pair passed through the Edwards doubling formulas.
>
>There are exceptional off-curve inputs to the doubling formula that can map into small torsion. For example, in the BN254 cofactor-8 case, an off-curve `S` can satisfy `[8]S = (0,-1)` under the circuit’s iterated doubling equations. Then in an identity-result edge case such as `P2 = -P1`, `s1 = s2 = 1`, an even denominator decomposition can make the scaled MSM accept `R = (0,-1)` because `[2]R = O`; the subgroup check is supposed to reject this, but may not if `S` is unconstrained.
>
>Suggested fix: add `S.assertIsOnCurve(api, curve)` inside `assertInSubgroup` before the doubling loop, and add a regression test with a malicious hint returning a small-torsion `R` plus an off-curve preimage.

ivokub (changes requested):
It seems the snippets also disappeared during merge commit:
> It looks like the stats snippets added for this PR were dropped during the merge from master. In `55799fa49`, `internal/stats/snippet.go` still registered `msm_babyjubjub_2`, `msm_jubjub_2`, and `msm_bandersnatch_2`, and `internal/stats/latest_stats.csv` had the corresponding rows. At current HEAD these registrations and CSV rows are gone. The merge commit `4842f113d` modified both stats files, likely resolving to the master-side map-to-curve stats changes and accidentally losing the MSM snippets. Could you restore the twisted Edwards MSM snippets and regenerate/update `latest_stats.csv`?

And it also seems that the hinted result checking is not complete as assumes hinted result being on-curve.

std/algebra/native/twistededwards/point.go
```diff
@@ -183,3 +186,358 @@ func (p *Point) scalarMulFakeGLV(api frontend.API, p1 *Point, scalar frontend.Va
...
+		panic("twistededwards: cofactor must be a power of two")
+	}
+	var cS Point
+	cS.X, cS.Y = S.X, S.Y
+	for c := cofactor; c > 1; c /= 2 {
+		cS.double(api, &cS, curve)
+	}
+	api.AssertIsEqual(cS.X, R.X)
+	api.AssertIsEqual(cS.Y, R.Y)
+}
```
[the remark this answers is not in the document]
yelhousni: same as Ivo's comment

ivokub (approved):
Thanks! Looks good now!

PR #1770, opened 2026-05-20
Fix/gkr/poseidon serialization

### LINKED ISSUE #1771: bug: ccs.Solve panics after CBOR round-trip of a GKR constraint system
Deserializing a constraint system that contains a GKR blueprint and calling Solve on it panics:

`panic: level N: unknown proving level type *constraint.GkrSkipLevel`
Root cause. `GkrProvingSchedule = []GkrProvingLevel` is serialized via `fxamacker/cbor/v2` with a tagged-type set. When a tagged interface slot is decoded, the library always allocates via `reflect.New(registeredType)`, which produces a pointer regardless of whether a value type was registered. The construction and prover paths build schedule levels as value types and assert against value types `(constraint.GkrSkipLevel)`, so the pointer coming out of decode never matches and the type switch falls through to the panic branch.

Reproduction. Compile any GKR-backed circuit (e.g. GKR-Poseidon2), write the constraint system to disk with `ccs.WriteTo`, read it back in a fresh process with `ccs.ReadFrom`, and call `ccs.Solve`.

Affects. Any code path that serializes a compiled GKR circuit and solves against it in a separate process or after a restart — i.e., all production solver/prover deployments using GKR blueprints.

Tabaie:
When deserializing, CBOR returns pointers to structs. The GKR backend currently expects struct types only for its proof schedule metadata, causing all production flows to fail.
This PR adds an integration test replicating the failure reported by @gusiri, and fixes it by having the backend work exclusively with pointers rather than with structs directly.

---

> [!NOTE]
> **Medium Risk**
> Touches GKR proving/verification and schedule serialization/deserialization to switch level types from values to pointers; errors here can break proof generation/verification or backward compatibility for serialized circuits.
> 
> **Overview**
> Fixes GKR schedule CBOR round-tripping by making all `GkrProvingLevel` implementations operate as *pointer types* (and registering pointer types in `constraint/marshal.go`), matching how CBOR decodes tagged structs.
> 
> Updates the GKR core, generated backends, and schedule builder/serializer to consistently type-assert `*constraint.GkrSkipLevel`, `*constraint.GkrSumcheckLevel`, and `*constraint.GkrSingleSourceZeroCheckLevel`, and adjusts helper APIs like `CollectOutgoingEvalPoints` accordingly.
> 
> Replaces the old `std/gkrapi` serialization test with a new end-to-end integration test that compiles a small Poseidon2 GKR circuit, writes `.scs` + witness files, then loads them in a package that does not import `gkrapi` and calls `Solve` to ensure deserialization works in production-like flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bda87be940bb33f71b35775eb0f377920f362818. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gusiri (approved):

AlexandreBelling (dismissed):

internal/gkr/gkrcore/serialize.go
```diff
@@ -180,11 +180,11 @@ func SerializeSchedule(w io.Writer, s constraint.GkrProvingSchedule) error {
...
-		case constraint.GkrSumcheckLevel:
+		case *constraint.GkrSumcheckLevel:
 			levelType = 0
-		case constraint.GkrSkipLevel:
+		case *constraint.GkrSkipLevel:
 			levelType = 1
-		case constraint.GkrSingleSourceZeroCheckLevel:
+		case *constraint.GkrSingleSourceZeroCheckLevel:
 			levelType = 2
 		}
```
ivokub, line 191: should panic in case unknown. otherwise levelType would be silently 0

std/permutation/poseidon2/gkr-poseidon2/gkrposeidon2testing/circuit.go
```diff
@@ -0,0 +1,43 @@
+// Package gkrposeidon2testing exposes a validator circuit that asserts the
```
ivokub: Usually package documentation should be:
```
// Package gkrposeidon2testing exposes common GKR Poseidon2 testing utilities.
//
// It exposes a validator circuit that asserts the
// GKR-backed Poseidon2 compressor agrees with the direct in-circuit Poseidon2
// compressor. It is intended for test and test-vector generation code; importing
// it transitively brings in gkrapi, so consumers should be limited to the builder
// side of the pipeline (compilation, witness generation).
```

ivokub: Though imo it is not a good convention to leave the testing packages public. It only confuses when using externally if it is some "testing" for inputs etc. or only a utility package.

We already have `internal/gkr/gkrtesting` and `internal/gkr/test_vectors` package. Imo we shouldn't sprinkle small GKR packages all around the repo, particularly when they just define a single type.

ivokub: For documentation convention, see https://go.dev/doc/comment. And as an example how it looks when shown in godoc: https://pkg.go.dev/github.com/consensys/gnark#section-directories

internal/regression_tests/gkr_solve/gkr_solve_test.go
```diff
@@ -0,0 +1,47 @@
+// Package gkr_solve hosts a regression test that exercises the full GKR
```
ivokub: Same comment about the package comment (i.e. first line a short intro for showing in a directory list and then the next section for full description).

However for this package, please look at the other packages under `regression_test` - they refer to concrete reported issues through issue number, imo we should follow the same pattern for this package as well.

internal/regression_tests/gkr_solve/gkr_solve_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/consensys/gnark-crypto/ecc"
+	_ "github.com/consensys/gnark-crypto/ecc/bls12-377/fr/mimc" // register native MIMC, used by GKR Fiat-Shamir during Solve
+	"github.com/consensys/gnark/backend/plonk"
+	"github.com/consensys/gnark/backend/witness"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGkrIntegration(t *testing.T) {
+	ccs := plonk.NewCS(ecc.BLS12_377)
+	scsFile, err := os.Open("testdata/gkr_poseidon2.scs")
```
ivokub: How is the test instance generated?

What I'd like to avoid is to have a odd drift in the future when we use the static testdata and when we update compiled CS definitions then we would get some sudden CI failures with difficult-to-debug solutions.

I see that the testdata is generated in `internal/gkr/test_vectors/generate.go` though, but it is quite far away from this package, so it is not obvious at all how to fix if this test starts failing. My recommendation -- this test imo suits better in `internal/gkr/test_vectors` as before. Plus I think we should have in the test output a recommendation a guide how to fix if this test fails. A la
```go
_, err := ccs.ReadFrom(scsFile)
require.NoError(t, err, "Plonk CS deserialization failed with GKR instructions, have you run `go generate`?")
```

ivokub (changes requested):
Several things:
* we have unhandled default case which may leave the levelType to silently be set to 0
* some comments
* imo the new testing packages don't make complete sense -- testdata generation is away from the actual test itself, plus we add new public testing packages which at least confuses me

One more thing I would do to have strong type safety and to avoid issues in the future. Right now we have GkrProvingLevel interface and now we have implementing types GkrSkipLevel, GkrSumCHeckLevel, GkrSingleSourceZeroCheckLevel. Right now we have methods:
```go
func (l GkrSumcheckLevel) NbOutgoingEvalPoints() int { return 1 }
func (l GkrSumcheckLevel) NbClaims() int {
	n := 0
	for _, g := range l {
		n += len(g.Wires) * len(g.ClaimSources)
	}
	return n
}
func (l GkrSumcheckLevel) ClaimGroups() []GkrClaimGroup         { return l }
func (l GkrSumcheckLevel) FinalEvalProofIndex(wireI, _ int) int { return wireI }
```

this means that both `GkrClaimGroup{}` and `&GkrClaimGroup{}` implement GkrProvingLevel. If we would miss by accident somewhere to take pointer of it then we get the serialization issue again. However, if we would define:
```go
func (l *GkrSumcheckLevel) NbOutgoingEvalPoints() int { return 1 }
func (l *GkrSumcheckLevel) NbClaims() int {
	n := 0
	for _, g := range l {
		n += len(g.Wires) * len(g.ClaimSources)
	}
	return n
}
func (l *GkrSumcheckLevel) ClaimGroups() []GkrClaimGroup         { return l }
func (l *GkrSumcheckLevel) FinalEvalProofIndex(wireI, _ int) int { return wireI }
```
then we are type safe as `GkrSumcheckLevel{}` would not implement the GkrProvingLevel interface, only `&GkrSumcheckLevel{}` would.

I wrote a small example what I meant, see https://go.dev/play/p/VcnVo3tEKA0

ivokub:
And btw -- the tests were actually failing before, it is just it didn't surface in the CI. If you run
```bash
$ go test -v -run ^TestSerialization/gkr_cube$ ./constraint/bn254/
=== RUN   TestSerialization
=== RUN   TestSerialization/gkr_cube
    r1cs_test.go:52: gkr_cube
    r1cs_test.go:87: round trip mismatch (-want +got):
          &cs.system{
                System: constraint.System{
                        ... // 2 identical fields
                        Type:         1,
                        Instructions: {{BlueprintID: 3, WireOffset: 7}, {BlueprintID: 1, WireOffset: 9, StartCallData: 5}, {BlueprintID: 1, ConstraintOffset: 1, WireOffset: 9, StartCallData: 15}, {BlueprintID: 3, ConstraintOffset: 2, WireOffset: 9, StartCallData: 25}, ...},
                        Blueprints: []constraint.Blueprint{
                                ... // 2 identical elements
                                &constraint.BlueprintBatchInverse[github.com/consensys/gnark/constraint.U64]{},
                                &gkr.BlueprintSolve{Circuit: {...}, NbInstances: 2},
        -                       &gkr.BlueprintProve{
        -                               SolveBlueprintID: 3,
        -                               SolveBlueprint: &gkr.BlueprintSolve{
        -                                       Circuit:     gkrcore.Circuit[github.com/consensys/gnark/internal/gkr/gkrcore.GateBytecode]{{}, {...}, {...}},
        -                                       NbInstances: 2,
        -                                       outputWires: []int{1, 2},
        -                               },
        -                               Schedule: constraint.GkrProvingSchedule{
        -                                       constraint.GkrSumcheckLevel{{...}}, constraint.GkrSumcheckLevel{{...}},
        -                                       constraint.GkrSingleSourceZeroCheckLevel{Wires: []int{...}, ClaimSources: []constraint.GkrClaimSource{...}},
        -                               },
        -                               HashName: "MIMC",
        -                       },
        +                       &gkr.BlueprintProve{
        +                               SolveBlueprintID: 3,
        +                               Schedule: constraint.GkrProvingSchedule{
        +                                       &constraint.GkrSumcheckLevel{{...}}, &constraint.GkrSumcheckLevel{{...}},
        +                                       &constraint.GkrSingleSourceZeroCheckLevel{Wires: []int{...}, ClaimSources: []constraint.GkrClaimSource{...}},
        +                               },
        +                               HashName: "MIMC",
        +                       },
                                &gkr.BlueprintGetAssignment{SolveBlueprintID: 3},
                        },
                        CallData:      {5, 0, 1, 1, ...},
                        NbConstraints: 3687,
                        ... // 5 ignored and 9 identical fields
                },
                CoeffTable: {Coefficients: {{0, 0, 0, 0}, {12436184717236109307, 3962172157175319849, 7381016538464732718, 1011752739694698287}, {6425625360762666998, 7924344314350639699, 14762033076929465436, 2023505479389396574}, {10902020042510041094, 17381486299841078119, 5900175412809962030, 2475245527108272378}, ...}},
                ... // 1 ignored field
          }
--- FAIL: TestSerialization (0.33s)
    --- FAIL: TestSerialization/gkr_cube (0.32s)
FAIL
FAIL    github.com/consensys/gnark/constraint/bn254     0.336s
FAIL
```

The same test with `-short` passes though (I guess it doesn't enforce serialization test).

std/permutation/poseidon2/gkr-poseidon2/gkrposeidon2testing/circuit.go
Tabaie: Yes perhaps this package adds complexity rather than reducing it.

internal/regression_tests/gkr_solve/gkr_solve_test.go
Tabaie: Makes sense. We can just define it as an integration test rather than a regression test.

Tabaie:
@ivokub Think I implemented all your suggestions.

ivokub:
Thanks, I rechecked the latest branch and this addresses my review comments.

The GKR schedule serialization path now returns an error for unknown level types instead of silently encoding them as sumcheck levels, and the `GkrProvingLevel` methods now use pointer receivers so value-typed schedule levels no longer satisfy the interface accidentally.

The test-structure comments also look addressed: the extra public `gkrposeidon2testing` package is gone, and the serialized solve check now lives under `internal/gkr/test_vectors/integration_test` next to the generated data.

I also reran the focused checks:

```bash
go test ./internal/gkr/test_vectors/integration_test ./internal/gkr/gkrcore ./internal/gkr
go test ./std/permutation/poseidon2/gkr-poseidon2 ./std/hash/poseidon2/gkr-poseidon2 ./constraint
go test -v -run "^TestSerialization/gkr_cube$" ./constraint/bn254/
```

All pass. Looks good from my side.

ivokub (approved):

PR #1772, opened 2026-05-22
fix: bound GLV hint subscalar bit widths

### LINKED ISSUE #1767: fix: bound the subscalar widths in GLV methods
There is an old soundness bug still in emulated GLV/fakeGLV/GLV+fakeGLV -- when the hint returns the subscalars, we check that they add up to the initial scalar. But the issue is that we only use subset of the bits when doing scalar mul (corresponding to the expected non-zero bits).

For example:
```
r = P-256 scalar order
n = 128
H = 2^128
s = r - H - 1

malicious decomposition:
  s1 = H + 1
  s2 = 1
  sign = 0

malicious scalar-mul output:
  R = [r - 1]P = -P

low128(s1) = 1
low128(s2) = 1
So it checks:

[1]P + [1]R = P - P = O
```

We should address it consistently over all curves and G1/G2. I think the best approach for solving it is actually in non-native package. We currently anyway range check the outputs coming from the hint to be in range (of the non-native modulus), but could optionally make the bound smaller. This actually should reduce the constraints.

_Originally posted by @ivokub in https://github.com/Consensys/gnark/pull/1763#pullrequestreview-4267621967_

1sgtpepper:
# Description

Fixes #1767.

This constrains hinted subscalars to the bit windows consumed by the emulated scalar-mul accumulator loops. The fix covers:
- generic GLV
- joint generic GLV
- fake GLV
- GLV + fake GLV
- BLS12-381 G2 GLV

Without these bounds, a malicious hint can satisfy the decomposition relation while placing data above the consumed bit window, so the accumulator checks only the low bits.

The regressions replace decomposition hints with forbidden high-bit outputs and matching forged results. They are rejected after the unused high bits are constrained.

## Type of change

# How has this been tested?

- [x] `go test -run 'TestScalarMulG2RejectsWideSubscalars|TestScalarMulG2TestSolve' ./std/algebra/emulated/sw_bls12381`
- [x] `go test -run 'TestScalarMulFakeGLVRejectsWideSubscalars|TestScalarMulGLVAndFakeGLVRejectsWideSubscalars|TestScalarMulGLVAndFakeGLV_TrivialDecompositionRegression' ./std/algebra/emulated/sw_emulated`
- [x] `go test -short ./std/algebra/emulated/sw_bls12381`
- [x] `go test -short ./std/algebra/emulated/sw_emulated`
- [x] `golangci-lint run -v --timeout=5m ./std/algebra/emulated/sw_bls12381 ./std/algebra/emulated/sw_emulated`
- [x] `git diff --check`

Not run; this adds soundness constraints to enforce existing decomposition bounds rather than changing the scalar-mul algorithm.

# Checklist:

---

> [!NOTE]
> **High Risk**
> Adds new circuit constraints in multiple scalar-mul implementations to prevent malicious hints from hiding data in unused high bits, which is a security/soundness-critical area and could affect proof validity and constraint counts.
> 
> **Overview**
> Fixes a soundness hole in emulated scalar multiplication by **constraining hinted GLV/fake-GLV subscalars to the bit windows actually consumed by the accumulator loops** (generic `scalarMulGLV`, joint GLV, `scalarMulFakeGLV`, and `scalarMulGLVAndFakeGLV`). This is implemented via a shared `assertBitLength` helper in `sw_emulated/point.go` and an explicit `130`-bit check in `sw_bls12381/g2.go`.
> 
> Adds regression tests that replace decomposition hints with *wide/high-bit* outputs (and forged scalar-mul hint outputs) and assert the circuits now reject them, including a new `TestScalarMulG2RejectsWideSubscalars` and similar tests for fake-GLV and GLV+fake-GLV paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d71e2be2dbaa74382acdbccef3f452cb54a428b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/std/algebra/emulated/sw_bls12381/g2.go
+++ b/std/algebra/emulated/sw_bls12381/g2.go
@@ -674,6 +674,12 @@ func (g2 *G2) scalarMulGLV(Q *G2Affine, s *Scalar, opts ...algopts.AlgebraOption
 	s1bits := g2.fr.ToBits(s1)
 	s2bits := g2.fr.ToBits(s2)
 
+	nbits := 130
+	for i := nbits; i < len(s1bits); i++ {
+		g2.api.AssertIsEqual(s1bits[i], 0)
+		g2.api.AssertIsEqual(s2bits[i], 0)
+	}
+
 	// precompute -Q, -Φ(Q), Φ(Q)
 	var tableQ, tablePhiQ [3]*G2Affine
 	negQY := g2.Ext2.Neg(&Q.P.Y)
@@ -757,7 +763,6 @@ func (g2 *G2) scalarMulGLV(Q *G2Affine, s *Scalar, opts ...algopts.AlgebraOption
 	// note that half the points are negatives of the other half,
 	// hence have the same X coordinates.
 
-	nbits := 130
 	for i := nbits - 2; i > 0; i -= 2 {
 		// selectorY takes values in [0,15]
 		selectorY := g2.api.Add(
--- a/std/algebra/emulated/sw_bls12381/g2_test.go
+++ b/std/algebra/emulated/sw_bls12381/g2_test.go
@@ -9,6 +9,7 @@ import (
 	bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381"
 	"github.com/consensys/gnark-crypto/ecc/bls12-381/fp"
 	fr_bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381/fr"
+	"github.com/consensys/gnark/constraint/solver"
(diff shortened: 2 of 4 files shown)


ivokub:
Thanks for the PR. However we have another idea of solving it directly by redefining the non-native hint calls. Your current approach requires additional binary decomposition and assertions which is wasteful in adding excessive constraints.

In our proposed solution we instead reduce the range check amount in non-native hint calling, allowing to reduce the constraints instead. But this is very intricate design and we'd better address it by the team.

PR #1778, opened 2026-06-08
fix: GLV subscalar range checks

### LINKED ISSUE #1767: fix: bound the subscalar widths in GLV methods

ivokub:
# Description

Fixes #1767 

Also implements in generic hint calling in field emulation to define the range check amounts. I think this is also independently relevant elsewhere where we expect smaller outputs from the hint.

## Type of change

# How has this been tested?

Existing tests for scalarmul pass

Added tests for the reduced rangecheck edge cases in field emulation.

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes in-circuit soundness assumptions for GLV/fake-GLV hints—incorrect bit bounds could admit malicious decompositions—though bounds match documented lattice limits and are covered by new tests.
> 
> **Overview**
> Adds **`WithHintOutputRangeCheckBits`** to emulated generic hints so hint outputs can be range-checked and packed at a **custom bit width** (via `packLimbsWithWidth` / refactored `enforceWidth`), instead of always using full modulus-width checks.
> 
> **GLV / fake-GLV scalar multiplication** (emulated `sw_emulated`, native G2 on BLS12-381/BN254/BW6-761, Grumpkin) now passes **LLL/Hermite-sized bounds** on decomposed sub-scalars (`u1,u2,v1,v2`, `s1,s2`, etc.) when calling `NewHintGeneric`, aligning constraints with proven decomposition sizes and **reducing scalar-mul cost** (reflected in `latest_stats.csv`).
> 
> Removes **unused** `decomposeScalarG1` / `decomposeScalarG2` hint registrations where decomposition moved to shared emulated paths; Grumpkin’s `callDecomposeScalar` drops the `simple` flag and uses the generic hint API with **127-bit** output checks.
> 
> New **`field_hint_test`** coverage exercises custom range-check indexing, multi-limb widths, and validation errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 852a650137c22a9821fec8b358b648b822ecf8f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


std/math/emulated/field_hint.go
```diff
@@ -334,13 +355,49 @@ func unwrapGenericHintOutputs[T1, T2 FieldParams](field *big.Int, fp1 *Field[T1]
...
+		// we only return the element with custom number of limbs so the caller
+		// can never access the potentially non-zero limbs beyond nbCustomLimbs,
+		// and thus there is no soundness issue.
+		outputs[idx] = fp.packLimbsWithWidth(limbs[:nbCustomLimbs], bits)
+	} else {
+		// bits <= 0: pack with default width but without range check.
+		// however, return it as non-internal so that when it is used as input to another operation,
+		// it will be properly range checked.
+		outputs[idx] = &Element[T]{Limbs: limbs, overflow: 0, internal: false}
+		// outputs[idx] = fp.newInternalElement(limbs, 0)
```
yelhousni: dead comment?

std/algebra/emulated/sw_emulated/point.go
```diff
@@ -1372,7 +1380,9 @@ func (c *Curve[B, S]) scalarMulFakeGLV(Q *AffinePoint[B], s *emulated.Element[S]
 
 	// First we find the sub-salars s1, s2 s.t. s1 + s2*s = 0 mod r and s1, s2 < sqrt(r).
```
yelhousni: comment should be updated to |s1|,|s2| < γ₂·√r ≈ 1.15·√r

yelhousni (dismissed):
LGTM. The bound tightening is mathematically sound — (BitLen+3)/4 + 2 covers γ₄·r^(1/4) and (BitLen+1)/2 covers γ₂·√r with comfortable slack. The WithHintOutputRangeCheckBits API is a clean generalization and the enforceWidth refactor preserves existing semantics. Test coverage (multi-limb, zero-bits, bad indices) is good.

std/math/emulated/field_hint.go
ivokub: Yup, removed

std/algebra/emulated/sw_emulated/point.go
ivokub: Updated comment.

PR #1420, opened 2025-02-13
Perf: optimize selector.Mux with recursive BinaryMux for various sizes

weijiguo:
# Description

(edited)

This PR is an optimized complement to the `selector.Mux` function. Instead of dot product, we recursively do `selector.BinaryMux` even when the length of inputs is not 2's power, each time with a different size depending on how the length is decomposed into binary.

A `bound` parameter is required to construct `BoundedComparator`. Users are responsible for ensuring `|sel - len(inputs)| <= absDiffUpp` as required by `cmp.NewBoundedComparator`.

# How has this been tested?

- [x] All existing tests

# How has this been benchmarked?

- [x] Added new tests showing the saved constraints. We moved the replaced implementation to test scope for benchmarking purpose, and count the constraint counts from 2 to 899. The constraint counts have been improved when the `inputs` array is large enough (`> 16`), but this is subject to the bound passed to the test case (`TestBenchMux1`), and users should have the option to choose based on benchmarking data.

- [x] Additionally we compare `BinaryMux` against `dotProduct` when n is 2's power (`TestBenchMux2`). `BinaryMux` is strictly better.

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I did not modify files generated from templates
- [x] `golangci-lint` does not output errors locally
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1809, opened 2026-08-18
feat: allow modifying the upper bound in bits.ToBinary

### LINKED ISSUE #1434: feat: allow modifying the upper bound in `bits.ToBinary`
Currently, when calling `bits.ToBinary`, we constrain that the bits are bits, the bits add up to the initial value and that the bits do not overflow (by checking that the bits are less than scalar field modulus bits). But for other uses it would be great if we could have other constant bound we check against, see for example discussions in https://github.com/Consensys/gnark/pull/1420

zexoverz:
Closes #1434.

`ToBinary` compares the decomposition against the native modulus minus one, which is the weakest bound that still keeps the decomposition unique. Callers who already know their value lives in a smaller range had no way to say so, which came up in the discussion on #1420.

`WithUpperBound` sets that constant. Two guards come with it. A bound at or above the modulus is rejected, because it readmits the `a` / `a+r` ambiguity the check exists to prevent. And it is refused alongside `OmitModulusCheck`, since one asks for a bound check and the other removes it.

The part worth reviewing closely is that an explicit bound always enforces the comparison, including when `WithNbDigits` is below the field bitlength and the check would otherwise be skipped. The digit count implies its own bound, not the caller's, so without this `WithNbDigits(8)` together with `WithUpperBound(100)` would have silently constrained nothing. There is a test for exactly that case.

Constant inputs are checked at compile time, matching how `WithNbDigits` already reports an out-of-range constant.

With the option unset every path is unchanged. Tests run across the curves and backends via `CheckCircuit`; `std/rangecheck`, `std/math/cmp` and `std/selector` still pass.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit d8cd6fad51c63f8d67d625329857ddf47cbb5d61. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/std/math/bits/conversion.go
+++ b/std/math/bits/conversion.go
@@ -2,6 +2,7 @@ package bits
 
 import (
 	"errors"
+	"math/big"
 
 	"github.com/consensys/gnark/frontend"
 )
@@ -53,6 +54,7 @@ type baseConversionConfig struct {
 	UnconstrainedInputs  bool
 
 	omitModulusCheck bool
+	upperBound       *big.Int
 }
 
 // BaseConversionOption configures the behaviour of scalar decomposition.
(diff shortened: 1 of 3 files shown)


zexoverz:
@ivokub when you have a moment, this one is ready for a look (closes #1434).

PR #1828, opened 2026-09-21
feat(GKR): Scheduler sum-check batching

Tabaie:
This PR optimizes the scheduler to group contiguous zero-checks of the same degree into the same sum-check level.
It also fixes a bug regarding the hashing of source claims for input wires with multiple outs (that would be subject to a zero-check, not just verifier-side evaluation.)

---

> [!NOTE]
> **High Risk**
> Changes GKR proving schedules and Fiat–Shamir transcript contents for final evaluation claims, both of which are soundness-critical for the proof system.
> 
> **Overview**
> **DefaultProvingSchedule** now merges contiguous, same-degree wires that each need a sumcheck into **one** `GkrSumcheckLevel` with several claim groups (one sumcheck invocation), instead of emitting a separate level per wire batch. A shared `batchForWire` helper drives skip vs single-source vs sumcheck choice and the new multi-batch sumcheck loop.
> 
> **Fiat–Shamir binding** for `finalEvalProof` is tightened: `BindGkrFinalEvalProof` takes per-wire levels via new `GkrProvingSchedule.WireLevels`, and only **input wires assigned to a `GkrSkipLevel`** are left out of the transcript. Multi-source inputs proved through sumcheck or single-source zero-check are bound again, fixing incorrect omission when those wires were not skip-only.
> 
> Prove/verify paths (template, gadget, and field backends) precompute `wireLevels` and pass them into binding. **gkrtesting** adds a MiMC-like topology (`mimcLastGate`, updated `MiMCCircuit`) plus schedule shape tests at depth 2 and 3.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3c2b83773cb9ba345c15be3889b696be99c9a0e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ivokub (approved):
Thanks! Looks good, it fixes the reported issue.

I don't think we need to care about bugbot comment about breaking API. I think the `constraints` package is internal enough that we can break APIs there.

